PR #10, opened 2026-04-05
fix: allow switching accounts when no current account is logged in

treerobin06:
## Summary

- When adding a new account fails or the user logs out, `_perform_switch` raises `SwitchError("No current account to switch from")`, leaving the user stuck with no way to recover
- Now when no current account is detected, the backup step is skipped and target account credentials are written directly
- Transaction/rollback logic is conditional — only created when there is a current account to back up

## Reproduce

1. `claude-swap --add` → add a new account but let it fail (e.g. cancel OAuth mid-flow)
2. `claude-swap 1` → try to switch back to the original account
3. **Before**: `SwitchError: No current account to switch from`
4. **After**: switches successfully

## Test plan

- [x] All 61 existing tests pass (`uv run pytest tests/test_switcher.py`)
- [x] No changes to public API

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -966,40 +966,47 @@ class ClaudeAccountSwitcher:
             target_email = data["accounts"][target_account]["email"]
             current_identity = self._get_current_account()
 
-            if current_identity is None:
-                raise SwitchError("No current account to switch from")
-            current_email, _ = current_identity
+            has_current = current_identity is not None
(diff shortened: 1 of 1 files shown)


realiti4:
@treerobin06 thanks a lot for identifying this bug and for the fix. PR #22 covers the same scenario, plus fixes for the switch() rotate path and the missing-~/.claude.json edge case, as part of a broader import/export feature. Closing this in favor of that. Thanks again for the contribution!

PR #11, opened 2026-04-05
feat: add --slot option to --add-account

treerobin06:
## Summary

- Add `--slot NUM` option to `--add-account` for specifying which slot number to store the account in
- When the target slot is already occupied by a different account, prompt for confirmation before overwriting
- When the account already exists in a different slot, safely migrate it (old slot cleanup happens only after all confirmations pass)
- Without `--slot`, existing behavior is fully preserved (auto-assign next number, or refresh credentials in place)

## Motivation

Currently `--add-account` always auto-assigns `max(existing_slots) + 1`, which leads to gaps (e.g., after removing accounts 2-10, the next add goes to slot 14). The `--slot` option gives users control over organization.

## Usage

```bash
# Add to a specific slot
claude-swap --add-account --slot 2

# If slot 2 is occupied, prompts:
# ⚠ Slot 2 already occupied: user@example.com [Org Name]
# Overwrite slot 2? [y/N]
```

## Test plan

- [x] All 61 existing `test_switcher.py` tests pass
- [ ] Manual test: `--add-account --slot N` with empty slot
- [ ] Manual test: `--add-account --slot N` with occupied slot (confirm + cancel)
- [ ] Manual test: `--add-account --slot N` migrating from existing slot
- [ ] Manual test: `--add-account` without `--slot` (unchanged behavior)
- [ ] Manual test: `--slot` with non-add commands rejected

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,16 @@ cswap --add-account
 
 This will update the stored credentials without creating a duplicate.
 
+### Add account to a specific slot
+
+By default, `--add-account` assigns the next available slot number. Use `--slot` to place the account in a specific slot:
+
+```bash
+cswap --add-account --slot 3
+```
+
+If the slot is already occupied by a different account, you'll be prompted to confirm before replacing it.
+
 ### Other commands
 
 ```bash
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -46,6 +46,12 @@ Examples:
         action="store_true",
         help="Show OAuth token expiry state (use with --list)",
     )
+    parser.add_argument(
+        "--slot",
+        type=int,
+        metavar="NUM",
+        help="Specify slot number when adding account (use with --add-account)",
+    )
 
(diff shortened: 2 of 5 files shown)


f-liva:
This PR might also resolve the issue reported in #16 (https://github.com/realiti4/claude-swap/issues/16).

The problem described there is that `--add-account` always updates Account 1 instead of adding a new account when the user is logged in with a different account. With the `--slot` option introduced in this PR, users could explicitly specify which slot to use (e.g., `cswap --add-account --slot 3`), bypassing the auto-assignment logic that seems to be causing the bug.

Having explicit slot control would give users a workaround for the issue while also preventing gaps in account numbering after removing accounts. Just wanted to connect the two for the maintainers' awareness.

realiti4:
Hi, thanks for the feature. I think it is going to be useful, it looks good.

Could you also add this to readme under 'Other commands' section?

realiti4:
Merged it, thank you for the pr.

PR #13, opened 2026-04-05
fix: use UTF-8 for JSON config I/O on Windows

PatrickSys:
## Summary

On Windows, `Path.read_text()` / `write_text()` default to the process encoding (typically cp1252). Claude Code stores JSON as UTF-8, so configs that contain real Unicode (smart quotes, display names, etc.) can raise `UnicodeDecodeError` before `json.loads` runs.

This is especially noticeable after https://github.com/realiti4/claude-swap/commit/92094c96afc477998ce98e15428d44cb9d7fe1b5, which reads both possible config locations inside `_get_claude_config_path`.

## Change

- Pass `encoding="utf-8"` for JSON/config-related reads and writes in `switcher.py`, `models.py`, and `cache.py`.
- Use `credentials.encode("utf-8")` when writing the credentials file.
- Catch `UnicodeDecodeError` in `_read_json` and `_get_claude_config_path` where JSON parse errors are already swallowed.
- Add a regression test: UTF-8 `~/.claude.json` with smart quotes in `oauthAccount`.

Closes #12

## Tests

- `uv run pytest tests/test_switcher.py::TestGetClaudeConfigPathUtf8 tests/test_cache.py -q`

(Full `pytest` on this machine: 126 passed, 2 skipped; 1 unrelated failure in `test_cli.py` on Windows/Python 3.14 due to `os.geteuid` not existing.)

### DIFF
--- a/src/claude_swap/cache.py
+++ b/src/claude_swap/cache.py
@@ -19,10 +19,16 @@ def read_cache(path: Path, ttl: float, default=MISSING):
     callers can distinguish "no cache" from a cached ``None`` value.
     """
     try:
-        raw = json.loads(path.read_text())
+        raw = json.loads(path.read_text(encoding="utf-8"))
         if time.time() - raw["timestamp"] < ttl:
             return raw["data"]
-    except (OSError, json.JSONDecodeError, KeyError, TypeError):
+    except (
+        OSError,
+        json.JSONDecodeError,
+        UnicodeDecodeError,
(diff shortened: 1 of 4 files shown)


realiti4:
Thank you for the fix, I merged it. Also releasing this fix with [v0.7.2](https://github.com/realiti4/claude-swap/releases/tag/v0.7.2)

PR #20, opened 2026-04-17
feat: support managed api keys

ae94:
# Support `/login` managed API keys on macOS

### Problem

`cswap` only recognized credentials stored in the macOS Keychain under the `"Claude Code-credentials"` service — which is where OAuth (browser-based) logins land. Users who authenticate via `/login` with a managed API key are not covered: Claude Code stores those credentials in a separate Keychain entry under the service name `"Claude Code"` with a raw `sk-ant-...` value. As a result, `cswap --add-account` would fail to read their active credentials and account switching would not work.

### What changed

**`switcher.py`**

- Added two named constants for the two macOS Keychain services Claude Code uses:
  - `_KEYCHAIN_OAUTH_SERVICE = "Claude Code-credentials"` — browser OAuth tokens
  - `_KEYCHAIN_MANAGED_SERVICE = "Claude Code"` — `/login` managed API keys

- Added `_CREDS_MANAGED_SOURCE_PREFIX = "SOURCE:managed:"` — a prefix written into cswap's own backup keyring to distinguish managed-key accounts from OAuth accounts when restoring.

- Added `_read_macos_keychain(service)` — a small helper that reads any generic-password Keychain entry and returns `""` on not-found vs. `None` on error, removing duplicated subprocess boilerplate.

- `_read_credentials()` now tries `"Claude Code-credentials"` first, then falls back to `"Claude Code"`, and records the source in `self._active_creds_source` (`"oauth"` or `"managed"`).

- `_write_credentials()` writes to whichever Keychain service matches `_active_creds_source`, then **deletes the other service entry**. This prevents Claude Code from picking up stale credentials from a previous account after a switch.

- `_write_account_credentials()` prefixes the backup keyring value with `_CREDS_MANAGED_SOURCE_PREFIX` for managed-key accounts.

- `_read_account_credentials()` strips the prefix on read and restores `_active_creds_source`, so the downstream `_write_credentials()` call knows where to put the credentials back.

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -54,6 +54,16 @@ from claude_swap.process_detection import get_running_instances
 KEYRING_SERVICE = "claude-code"
 KEYRING_ACTIVE_USERNAME = "active-credentials"
 
+# macOS Keychain service names used by Claude Code.
+# OAuth tokens (browser /login) are stored under "Claude Code-credentials".
+# Managed API keys (/login with a key) are stored under "Claude Code".
+_KEYCHAIN_OAUTH_SERVICE = "Claude Code-credentials"
+_KEYCHAIN_MANAGED_SERVICE = "Claude Code"
+
+# Prefix stored in cswap's own backup keyring to mark managed-key accounts.
+# Backups without this prefix are assumed to be OAuth accounts.
+_CREDS_MANAGED_SOURCE_PREFIX = "SOURCE:managed:"
+
 # Usage cache
 _USAGE_CACHE_TTL = 15  # seconds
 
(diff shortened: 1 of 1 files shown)


ae94:
@realiti4 can we merge this one when you get a chance please?

kundralaci:
Facing the same issue, this would be superuseful! 🙏

realiti4:
Sorry for the delay. I was planning a cross-platform solution for this, but I haven’t had a chance to look into it yet. I’ll take a look shortly.

realiti4:
@ae94 thanks a lot for the PR and for pointing this out, and sorry for the long delay.

I've opened #72 with cross-platform API-key support, so I'll close this in favor of that. 

Also releasing as a pre-release [v0.15.0b1](https://github.com/realiti4/claude-swap/releases/tag/v0.15.0b1) 

Please let me know if everything looks good. Pre-releases can be installed like this:

`uv tool upgrade claude-swap --prerelease=allow`

We can use it like this, I also updated the readme:

`cswap --add-token sk-ant-api03-...`

PR #21, opened 2026-04-23
fix: fall back to file credentials on macOS when Keychain is unavailable

houko:
## Problem

On some macOS setups, Claude Code stores credentials in `~/.claude/.credentials.json` instead of the Keychain. This causes `cswap` to fail with:

```
Error: No credentials found for current account
```

and on `--add-account`:

```
keyring.errors.PasswordSetError: Can't store password on keychain: (-25308, 'Unknown Error')
```

Error `-25308` is `errSecInteractionNotAllowed` — Keychain interaction is blocked in certain macOS environments.

## Fix

- **`_read_credentials`**: Try Keychain first; fall back to `~/.claude/.credentials.json` if Keychain returns empty or fails
- **`_write_credentials`**: Try Keychain first; fall back to file write if Keychain fails
- **`_read/_write/_delete_account_credentials`**: Use file-based storage on macOS (same as Linux/WSL), avoiding Keychain entirely for cswap's own account storage

This is a non-breaking change — users with working Keychain are unaffected (Keychain is still tried first for reading/writing active credentials).

## Tested

Verified on macOS with credentials stored in file only — `cswap --add-account`, `--list`, and `--switch` all work after this fix.


realiti4:
Hi thank you for pointing this out and for the fixes.

I think we should keep -A fix cb4f750f6fefca2bf8c9a1dc6bf067561a3e7499, but not sure about the -a credentials one bc9db7688034f3743af0a2a5fc5d6ba710e92261

Based on Claude Code's source, it also uses `acct=$USER`

From `macOsKeychainHelpers.ts`:

```
export function getUsername(): string {
  try {
    return process.env.USER || userInfo().username
  } catch {
    return 'claude-code-user'
  }
}
```

And `macOsKeychainStorage.ts`:

```
const username = getUsername()
execSyncWithDefaults_DEPRECATED(
  `security find-generic-password -a "${username}" -w -s "${storageServiceName}"`
)
```

Hardcoding -a credentials would break existing keychain reads for users whose entries were created by Claude Code (which is everyone using the keychain path). Can we revert that change? Based on Claude Code's source your Mac setup also shouldn't need it, but can you confirm?

Also, I'm not sure about moving claude-swap storage to file-based on Mac. I think we should keep a similar "keychain first, fall back to file" pattern for claude-swap data as well.

If for both Claude Code creds and claude-swap creds we follow the same pattern, we should be safe:

Read: try primary (keychain) first; if it returns nothing, fall back to secondary (file).
Write: try keychain first. If it fails, write to file and delete the keychain entry if one exists. If for some reason keychain become unavailable we might left with stale data on next reads.

Claude Code's source uses this same pattern, so I think following this should be fine.

Could you update the PR based on these? Thanks again.

realiti4:
Hi, I don't have access to an environment that keychain is unavailable. Would you like to continue working on this, or should we close it?

realiti4:
Hi, I am closing this for now. If you'd like us to reopen, please let me know. Thanks.

PR #25, opened 2026-04-27
fix: skip backup write when switching to the same account (cross-container reload)

Ewal11:
## Problem

In multi-container setups where `~/.claude-swap-backup` is a shared volume, the following workflow silently fails:

1. **Container A**: `cswap --add-account` → refreshes account 1's tokens in the shared backup ✓
2. **Container B**: `cswap --switch-to 1` → Claude Code reports **not authenticated** ✗

### Root cause

`_perform_switch` always executes **Step 1: backup current account** before reading the target. The backup path is keyed only by `(account_num, email)`. When Container B's active account is the same slot/email as the target (common in multi-container setups), Step 1 overwrites the just-refreshed shared backup with Container B's stale local credentials. Step 2 then reads that overwritten data back, leaving Container B with an expired token.

\`\`\`
Container A: write fresh tokens → .creds-1-user@x.enc  ✓
Container B: --switch-to 1
  Step 1: write STALE local tokens → .creds-1-user@x.enc  ← clobbers Container A's work
  Step 2: read .creds-1-user@x.enc → gets stale tokens back
  Step 3: write stale tokens locally → not authenticated
\`\`\`

## Fix

When `current_account == target_account` and `current_email == target_email`, skip the backup write and reload directly from the shared backup. This mirrors the existing `current_identity is None` fresh-machine path already present in this function.

The user sees `Reloaded Account-N` instead of `Switched to Account-N` to signal the different code path.

## Testing

Shared-volume multi-container setup:
- `docker-compose.yml` mounts `~/.claude-swap-backup` as a shared volume across containers
- Container A: `cswap --add-account` (refreshes tokens)
- Container B: `cswap --switch-to 1` → now correctly loads Container A's fresh tokens

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -1127,6 +1127,48 @@ class ClaudeAccountSwitcher:
 
             current_email, _ = current_identity
 
+            # Same-account reload path: current and target are the same account
+            # (identified by slot number and email). This happens in multi-container
+            # setups where one container refreshes the shared backup via
+            # --add-account and another container wants to pull the update via
+            # --switch-to. The normal flow would overwrite the just-refreshed backup
+            # with the local container's stale credentials before reading it back,
+            # losing the update. Skip the backup write and reload directly.
+            if current_account == target_account and current_email == target_email:
+                target_creds = self._read_account_credentials(
+                    target_account, target_email
+                )
+                target_config = self._read_account_config(target_account, target_email)
+
+                if not target_creds or not target_config:
+                    raise SwitchError(
+                        f"Missing backup data for Account-{target_account}"
+                    )
+
+                self._write_credentials(target_creds)
+                try:
+                    target_config_data = json.loads(target_config)
+                except json.JSONDecodeError as exc:
+                    raise SwitchError(f"Invalid backup config: {exc}")
+                oauth_section = target_config_data.get("oauthAccount")
+                if not oauth_section:
+                    raise SwitchError("Invalid oauthAccount in backup")
+                current_config_data = self._read_json(config_path) or {}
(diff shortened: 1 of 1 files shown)


realiti4:
Hi, thanks for the fix and PR.

One concern is that this also affects normal setups where we do `switch-to (active account)` when claude code refreshed live tokens, but we'd assume the account is same and restore stale tokens from backup.

So how about adding this behaviour with an optional flag like `--restore-from-backup` or something? On your setup it would look like this:

`cswap --switch-to 1 --restore-from-backup`

Thanks again for the pr.

src/claude_swap/switcher.py
```diff
@@ -1127,6 +1127,48 @@ def _perform_switch(self, target_account: str) -> None:
...
             current_email, _ = current_identity
 
+            # Same-account reload path: current and target are the same account
+            # (identified by slot number and email). This happens in multi-container
+            # setups where one container refreshes the shared backup via
+            # --add-account and another container wants to pull the update via
+            # --switch-to. The normal flow would overwrite the just-refreshed backup
+            # with the local container's stale credentials before reading it back,
+            # losing the update. Skip the backup write and reload directly.
+            if current_account == target_account and current_email == target_email:
```
realiti4, line 1137: With the flag, this condition becomes just `if restore_from_backup: `

Ewal11:
Yeah I think --restore-from-backup sounds good.
Again thanks for the repo, very usable day to day.

realiti4:
Feel free to push new flag changes whenever you have time. We can merge with that.

realiti4:
Is this still needed and do you plan to keep working on it? If so please let me know so we can reopen.

Issue #26, opened 2026-04-28
Switch to XDG Base Directory spec for storage location on Linux


babs:
## Summary

`claude-swap` currently stores all data under `~/.claude-swap-backup/` on every platform. On Linux this doesn't follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), which most modern CLI tools adhere to. Proposing to move the default storage to an XDG-aware location on Linux, with automatic migration for existing users.

I'd like to implement this myself — happy to open a PR once the approach is agreed.

## Current behavior

All platforms use:

```
~/.claude-swap-backup/
├── sequence.json
├── configs/
├── credentials/      # Linux only (base64-encoded creds)
├── cache/
└── claude-swap.log
```

## Proposed behavior

Move everything under a single XDG-aware root on Linux:

```
${XDG_DATA_HOME:-~/.local/share}/claude-swap/
├── sequence.json
├── configs/
├── credentials/
├── cache/
└── claude-swap.log
```

Note: this is **data**, not user-edited config — so `$XDG_DATA_HOME` (`~/.local/share/`) is the correct target, *not* `$XDG_CONFIG_HOME` (`~/.config/`).

Rationale for keeping it under one root rather than splitting across `$XDG_CACHE_HOME` / `$XDG_STATE_HOME`: the tool is small, the cache is negligible, and a single directory keeps `--purge`, backup, and migration trivially simple. The main XDG win — getting out of `$HOME` root and respecting user overrides — is preserved.

## Scope

- **Linux / WSL only.** macOS and Windows have their own conventions (`~/Library/Application Support/`, `%APPDATA%\`) and the current flat-home layout mirrors upstream Claude Code (`~/.claude/`, `~/.claude.json`), so changing those is out of scope here.

## Migration

To avoid breaking existing installs:

1. On startup, if the new XDG path doesn't exist but `~/.claude-swap-backup/` does → move the data once.
2. If **both** paths exist, abort migration and surface a clear error — never merge or overwrite.
3. Preserve file modes across the move (`0600` on credentials, `0700` on directories).
4. Print a one-line notice on the first run after migration so users know where data went.
5. Update `cswap --purge` to clean both the new path and any leftover `~/.claude-swap-backup/`.

## Documentation

- Update the "Data locations" table in `README.md`.
- Mention the `XDG_DATA_HOME` env var override.

## Out of scope

- Changing macOS/Windows defaults.
[...]

PR #27, opened 2026-04-28
feat: store backups under $XDG_DATA_HOME on Linux/WSL

babs:
- Comply with the XDG Base Directory Spec on Linux/WSL; macOS/Windows keep the legacy ~/.claude-swap-backup/ layout because their file-vs- keyring split is already tied to it.
- Existing installs migrate automatically on first run, guarded by a <target>.migrating flag so an interrupted run resumes from a known state rather than re-copying from a possibly-partial source.
- When both legacy and new paths exist without a flag, refuse to act — better to surface a clear error than risk silently overwriting data the tool didn't create.
- Surface filesystem failures via a typed MigrationError so the CLI exits cleanly instead of dumping a traceback.
- cswap --purge also sweeps any stale legacy directory left behind.

Address #26

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,9 @@ cswap --purge                   # Remove all claude-swap data
 |----------|-------------|----------------|
 | Windows | Windows Credential Manager | `~/.claude-swap-backup/` |
 | macOS | macOS Keychain | `~/.claude-swap-backup/` |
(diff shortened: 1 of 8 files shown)


realiti4:
Thanks for the PR, looking great. Merged it.

PR #28, opened 2026-04-29
fix: use file-based credential backup on Windows to avoid WCM size limit

figu6378-max:
## Problem

Two bugs break `cswap` on Windows.

### Bug 1: `--add-account` fails with WinError 1783

```
win32ctypes.pywin32.pywintypes.error: (1783, 'CredWrite', 'The stub received bad data.')
```

`_write_account_credentials` routes Windows through the keyring backend (Windows Credential Manager), which has a **~2560-byte hard limit** on credential blobs. Claude Code on Windows stores credentials in `~/.claude/.credentials.json`, which regularly exceeds this limit.

### Bug 2: `--list` crashes with UnicodeEncodeError

```
UnicodeEncodeError: 'charmap' codec can't encode character '├' in position 5: character maps to <undefined>
```

Windows PowerShell defaults to cp1252 encoding. Box-drawing characters used in `--list` output (`├` `└`) cannot be encoded by cp1252.

## Fix

**Commit 1** (`switcher.py`): Route Windows through the same file-based backup path already used by Linux/WSL. macOS continues to use the system keyring (Keychain). Also skip `os.chmod` on Windows (no-op).

**Commit 2** (`cli.py`): Reconfigure stdout/stderr to UTF-8 at CLI startup on Windows before any output is written.

## Testing

Verified on Windows 10 with Python 3.14:
- `cswap --add-account` succeeds (previously crashed on CredWrite)
- `cswap --list` renders correctly with box-drawing chars

### DIFF
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -14,6 +14,9 @@ from claude_swap.switcher import ClaudeAccountSwitcher
 
 def main() -> None:
     """Main entry point for the CLI."""
+    if sys.platform == "win32":
+        sys.stdout.reconfigure(encoding="utf-8", errors="replace")
+        sys.stderr.reconfigure(encoding="utf-8", errors="replace")
     parser = argparse.ArgumentParser(
         description="Multi-Account Switcher for Claude Code",
         formatter_class=argparse.RawDescriptionHelpFormatter,
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -258,10 +258,11 @@ class ClaudeAccountSwitcher:
     def _read_account_credentials(self, account_num: str, email: str) -> str:
         """Read account credentials from backup.
 
-        On Linux/WSL: Uses file-based storage to avoid keyring backend issues.
-        On macOS/Windows: Uses system keyring.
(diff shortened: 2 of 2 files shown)


realiti4:
Hi, thanks for pointing this out.

Could you split 9b1b00e0a905c39dc3f9aee9a1741f631c84a4eb into a separate PR? That UTF-8 stdout/stderr fix is small and safe, and I'm happy to merge it independently.

For the Windows credential-storage change, I'd like to reproduce the issue first and understand the actual limits/conditions. In my testing, credentials are around 1500–2000 bytes and Windows Credential Manager works fine. I'd like to check if we can do something to optmize the data first, then moving to file-based storage.

Moving Windows backups from keyring to file-based storage is a larger behavior change. To merge that safely, we'd need a migration/fallback plan for existing Windows users, plus tests and stale-data cleanup. Without that, users who already have accounts stored in Windows Credential Manager could lose access to their managed accounts after updating.

realiti4:
Thank you for the PR. I am closing this in favor of #46 as I mentioned we need a migration plan as well for older versions when we move to file based storage. Thanks again for pointing this out.

PR #29, opened 2026-04-29
feat: add --add-token for registering raw OAuth setup-tokens

### LINKED ISSUE #24: Feature request: add account from a raw OAuth setup-token (no prior Claude Code login)
Thanks for landing #22 — `--export`/`--import` already covers most of the multi-machine workflow.

One scenario it doesn't cover: adding an account from a **raw long-lived OAuth token** produced by `claude setup-token`. These tokens have no refresh_token, don't expire, and are the full credential by themselves.

### Why this matters

Today `cswap --add-account` requires the account to already be logged in on the current machine, because it reads from the platform keychain / `~/.claude.json` via `_read_credentials()`. That's a problem when:

- Setting up cswap on a **headless server** where running the full OAuth browser flow is awkward.
- Receiving a setup-token from another machine and wanting to register it as a cswap account **without first running `claude` to populate the local keychain**.
- Managing multiple long-lived tokens that I only ever hold as the raw token string.

`--import backup.cswap` doesn't help here — it only consumes cswap's own export schema, not a bare token.

### Proposed interface

```bash
cswap --add-token <SETUP_TOKEN>                  # auto-assign slot
cswap --add-token <SETUP_TOKEN> --slot 5         # explicit slot
cswap --add-token - --slot 5                     # read token from stdin (keeps it out of shell history)
```

### Implementation sketch

The new path would reuse `_write_account_credentials` / `_write_account_config` and the slot/sequence machinery — the only new piece is constructing the credential blob from the raw token (mirroring what Claude Code persists after `claude setup-token`) instead of reading it from `_read_credentials()`. Account email/label can be required as a CLI flag (`--email`) so the implementation stays free of any assumed Anthropic API call.

Happy to send a small PR if you're open to this.

treerobin06:
## Summary

Closes #24. Adds `cswap --add-token` for registering an account from a raw long-lived OAuth setup-token (e.g. produced by `claude setup-token`) without requiring a prior Claude Code login on the current machine.

```bash
cswap --add-token sk-ant-oat01-... --email user@example.com
cswap --add-token sk-ant-oat01-... --email user@example.com --slot 3
cswap --add-token - --email user@example.com    # token from stdin
cswap --add-token --email user@example.com      # prompt securely (getpass, no echo)
```

Per the issue discussion, `--email` is required so cswap's metadata stays aligned with the rest of the accounts. No Anthropic API calls are made.

## Why

Today `--add-account` reads from the platform keychain / `~/.claude.json`, which means the account must already be logged in on the current machine. That's a problem on headless servers, or when receiving a setup-token from another machine. `--import` doesn't help because it only consumes cswap's own export schema.

## Implementation

- New `add_account_from_token(token, email, slot)` in `switcher.py`. Constructs the credential blob (`{"claudeAiOauth": {"accessToken": ...}}`) and a minimal config blob (`{"oauthAccount": {"emailAddress": ...}}`) instead of reading from `_read_credentials()`.
- Reuses `_write_account_credentials` / `_write_account_config` and the slot/sequence machinery — `--slot` works the same as with `--add-account`, and re-adding the same email without `--slot` refreshes credentials in place.
- Token sources: inline value, `-` for stdin, or empty for `getpass.getpass()` prompt.

## Test plan

- [x] 10 unit tests for `add_account_from_token` (basic add, blob format, slot handling, in-place refresh, validation, stdin, sequence ordering)
- [x] 4 CLI tests for `--add-token` (requires `--email`, dispatches to switcher, forwards `--slot`, appears in `--help`)
- [x] All 104 existing tests still pass

## Notes

- Token-added accounts are always treated as **personal** (`organizationUuid=""`). Org accounts would need an Anthropic API call to discover the UUID, which the issue explicitly ruled out.
- `accountUuid` is stored as `""` because we have no way to derive it from a raw token without an API call. This is consistent with what the rest of the code tolerates (e.g., `_get_display_tag` and migration code).

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -698,6 +698,166 @@ class ClaudeAccountSwitcher:
         self._logger.info(f"Added account {account_num}: {current_email} (org: {organization_uuid or 'personal'})")
         print(f"{accent('Added')} Account {account_num}: {current_email} {muted(f'[{tag}]')}")
 
+    def add_account_from_token(
+        self, token: str, email: str, slot: int | None = None
+    ) -> None:
+        """Register a raw OAuth setup-token as a new account.
+
+        Useful for headless servers or when the token is received from another
+        machine, without needing a prior Claude Code login on this machine.
+        No Anthropic API calls are made; ``email`` is taken as-is from the flag.
+
+        Args:
+            token: Raw OAuth access token, or ``"-"`` to read one line from
+                   stdin, or ``""`` to prompt securely via getpass.
+            email: Email address to associate with the account (required).
+            slot:  Slot number to use; auto-assigned when ``None``.
+        """
+        import getpass
+
+        if token == "-":
+            token = sys.stdin.readline().rstrip("\n")
+        elif not token:
+            token = getpass.getpass("Setup token: ")
+
+        token = token.strip()
+        if not token:
+            raise ValidationError("Token cannot be empty")
+
+        if not self._validate_email(email):
+            raise ValidationError(f"Invalid email format: {email}")
+
+        self._setup_directories()
+        self._init_sequence_file()
+        self._migrate_org_fields()
+
+        # If the account already exists (same email, personal), refresh in place.
+        if slot is None and self._account_exists(email, ""):
+            seq = self._get_sequence_data()
+            account_num = next(
+                (num for num, acc in seq.get("accounts", {}).items()
+                 if acc.get("email") == email
+                 and acc.get("organizationUuid", "") == ""),
+                None,
+            )
+            credentials = json.dumps({"claudeAiOauth": {"accessToken": token}})
+            config = json.dumps({
+                "oauthAccount": {
+                    "emailAddress": email,
+                    "accountUuid": "",
+                    "organizationUuid": None,
+                    "organizationName": None,
+                }
+            })
+            self._write_account_credentials(account_num, email, credentials)
+            self._write_account_config(account_num, email, config)
+            seq["lastUpdated"] = get_timestamp()
+            self._write_json(self.sequence_file, seq)
+            self._logger.info(f"Updated token for account {account_num}: {email}")
+            print(
+                f"{accent('Updated token')} for Account {account_num} "
+                f"({email} {muted('[personal]')})."
+            )
+            return
+
+        displace_slot = None
+        migrate_from = None
+
+        if slot is not None:
+            if slot < 1:
+                raise ConfigError("Slot number must be >= 1")
+            account_num = str(slot)
+            data = self._get_sequence_data()
+
+            if self._account_exists(email, ""):
+                old_num = next(
+                    (num for num, acc in data.get("accounts", {}).items()
+                     if acc.get("email") == email
+                     and acc.get("organizationUuid", "") == ""),
+                    None,
+                )
+                if old_num and old_num != account_num:
+                    migrate_from = old_num
+
+            if account_num in data.get("accounts", {}):
+                existing = data["accounts"][account_num]
+                existing_email = existing.get("email", "unknown")
+                is_same = (
+                    existing_email == email
+                    and existing.get("organizationUuid", "") == ""
+                )
+                if not is_same:
+                    existing_tag = self._get_display_tag(
+                        existing_email,
+                        existing.get("organizationName", ""),
+                        existing.get("organizationUuid", ""),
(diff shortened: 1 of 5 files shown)


src/claude_swap/switcher.py
```diff
@@ -698,6 +698,166 @@ def add_account(self, slot: int | None = None) -> None:
...
+        # If the account already exists (same email, personal), refresh in place.
+        if slot is None and self._account_exists(email, ""):
+            seq = self._get_sequence_data()
+            account_num = next(
+                (num for num, acc in seq.get("accounts", {}).items()
+                 if acc.get("email") == email
+                 and acc.get("organizationUuid", "") == ""),
+                None,
+            )
+            credentials = json.dumps({"claudeAiOauth": {"accessToken": token}})
```
realiti4: On my own testing, claude didn't recognize the token. After checking claude code source code I think it also expects scope as well. Could we update these two credentials blob with a helper like this?

```
@staticmethod
    def _setup_token_credentials(token: str) -> str:
        """Return the Claude Code OAuth storage shape for setup-token accounts."""
        return json.dumps({
            "claudeAiOauth": {
                "accessToken": token,
                "refreshToken": None,
                "expiresAt": None,
                "scopes": ["user:inference"],
                "subscriptionType": None,
                "rateLimitTier": None,
            }
        })
```

So this line would be:

`credentials = self._setup_token_credentials(token)`

src/claude_swap/switcher.py
```diff
@@ -698,6 +698,166 @@ def add_account(self, slot: int | None = None) -> None:
...
+                        print(f"\n{dimmed('Cancelled')}")
+                        return
+                    if answer not in ("y", "yes"):
+                        print(dimmed("Cancelled"))
+                        return
+                    displace_slot = (account_num, existing_email)
+        else:
+            account_num = str(self._get_next_account_number())
+
+        credentials = json.dumps({"claudeAiOauth": {"accessToken": token}})
```
realiti4: This would use the helper as well:

`credentials = self._setup_token_credentials(token)`

realiti4:
Hi, thanks a lot for the PR. I left a couple of comments. Could you check those please?

You may also need to update the tests based on the new credential blob including scopes.

treerobin06:
Hi @realiti4 — thanks for the review!

I addressed the scopes feedback in 928fd6f: the credential blob now seeds the default Claude Code scopes (`user:profile`, `user:inference`, `user:sessions:claude_code`) on both the basic-add and in-place-refresh paths, with corresponding test assertions. Full suite passes locally (264 passed / 2 skipped).

For the other inline comments you mentioned — I don't see any inline review comments on my end, so they may still be in a pending review on your side. Could you click "Submit review" so I can see them and address them too?

Happy to rebase onto main once the rest is covered.

realiti4:
Yeah, sorry you're right, I forgot to submit the review. I pushed a merge commit and a commit narrowing the scope to just ("user:inference",), because based on the Claude Code source I was worried wider scopes might trigger 403s on profile-scoped endpoints. Sorry, I pushed the commit myself didn't realize you didn't see the review content

realiti4:
Looking really good, thanks again for the PR. I think we can merge it in this state, if it's ok with you as well.

treerobin06:
@realiti4 thanks for merging! One quick note in case it's useful:

While digging through the Claude Code binary to verify this works end-to-end, I found why your narrow-scope shape is actually sufficient (without needing the 4 extra fields from your earlier helper proposal). Setup-token users only ever obtain the access-token string itself:

- `refreshToken` / `expiresAt` — by design absent for long-lived tokens
- `subscriptionType` / `rateLimitTier` — only returned by the full OAuth callback, not setup-token issuance
- `scopes` — server-fixed to `["user:inference"]`

The binary's `readAsync()` only validates `accessToken`, refresh gracefully short-circuits on missing `refreshToken`, and `subscriptionType` / `rateLimitTier` use `?? null` fallbacks. So a 6-field blob with 4 explicit `null`s and the current 2-field shape produce identical client behavior. The *"didn't recognize the token"* you ran into earlier was almost certainly from the original `f86ce5f` (no `scopes` field at all) — your narrow-scope commit fixed it.

One small follow-up idea: `--email` could default to something like `setup-token-{slot}@local` since users can't meaningfully provide an email for a token-only registration. Happy to send a small PR if you think it's worth doing.

realiti4:
Yeah I think it is a good idea. Please do when you have time.

PR #30, opened 2026-04-29
feat: add --tui interactive arrow-key menu

treerobin06:
## Summary

Adds an interactive single-level TUI activated via `cswap --tui` so users don't have to memorize flags. Uses stdlib `curses` only — **zero new dependencies**.

```
claude-swap
Active: user@example.com [personal]  ·  3 managed

> Switch account
  Add account
  Remove account
  Refresh credentials (current login, in-place)
  List accounts (with usage)
  Status
  Quit

[↑/↓] move  [Enter] select  [Esc/q] cancel
```

The TUI is a pure navigation layer — every action shells out to the existing `ClaudeAccountSwitcher` methods. No reimplementation of account logic, no new behavior. Existing CLI flags and behavior are unchanged.

## Why

Several users (myself included) have hit "I can't remember which flag combination does what" — especially around `--add-account` vs the many ways to refresh credentials, plus the slot/email options. A discoverable arrow-key menu lowers the bar for first-time users without changing anything for power users.

## Sub-flows

- **Switch account** → list of managed accounts → pick → calls `switch_to(N)`
- **Add account** → submenu (current login / setup-token) → calls `add_account()` or `add_account_from_token()`
- **Remove account** → list → pick → typed `y`/`n` confirm → calls `remove_account(N)`
- **Refresh credentials** → calls `add_account()` (which already has refresh-in-place semantics for the active login)
- **List accounts** / **Status** → shell out to existing methods so users see the rich colored output

After each action the user returns to the main menu — the TUI doesn't auto-exit.

## Add-account / setup-token branch

The submenu offers a second source ("From a setup-token") only when `add_account_from_token` exists on the switcher. Today that's [#29](https://github.com/realiti4/claude-swap/pull/29). On upstream/main as-is, the option is hidden — so this PR is **independent** of #29 and can be reviewed on its own.

## Implementation

- New module `src/claude_swap/tui.py` (~330 lines)
- `_select_from` and `_prompt_text` are thin curses wrappers (mockable for tests)
- `_shell_out` saves/restores curses state via `def_prog_mode` / `reset_prog_mode` so `list_accounts` / `status` can render with their existing ANSI colors, then returns to the menu
- `--tui` is added to the existing mutually-exclusive argument group
- The `tui` import in `cli.py` is deferred so platforms missing curses (Windows without `windows-curses`) only see an error if they actually invoke `--tui`
- Min terminal size check (12 rows × 60 cols) — bails to a friendly stderr message if too small

## Test plan

- [x] 20 unit tests for TUI: menu navigation, key bindings (↑/↓/k/j/Enter/Esc/q), wrap-around, cancel sentinels, all three sub-flows (switch / add / remove), dispatch correctness
- [x] 1 CLI integration test verifying `--tui` shows in `--help` and dispatches
- [x] All 110 tests pass on Python 3.14 / macOS (`pytest tests/`)
- [ ] **Manual smoke test required**: rendering on real terminals — TUI itself can't be auto-tested without a PTY harness. Tested locally on macOS Terminal; would appreciate Linux + Windows confirmation.

## Notes

- Pure monochrome (bold + dim attrs only) for portability
- No window-resize handling for v1 — if the user resizes mid-flow, the next redraw catches up
- ESC backs out of any sub-page; `q` only quits from the main menu (so accidental `q` in a text input doesn't lose the form)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,17 @@ cswap --add-account
 
 This will update the stored credentials without creating a duplicate.
 
+### Interactive TUI
+
+If you don't want to remember flags, launch the arrow-key menu:
+
+```bash
+cswap --tui
+```
+
+You get a single-level menu (Switch / Add / Remove / Refresh / List / Status / Quit).
+Pure stdlib `curses`, no extra dependencies. On Windows, install `windows-curses`.
+
 ### Other commands
 
 ```bash
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -29,6 +29,7 @@ Examples:
   %(prog)s --purge
   %(prog)s --export backup.cswap
   %(prog)s --import backup.cswap
+  %(prog)s --tui                              # interactive arrow-key menu
         """,
     )
 
@@ -117,6 +118,11 @@ Examples:
         metavar="PATH",
         help="Import accounts from file (use '-' for stdin)",
     )
+    group.add_argument(
+        "--tui",
+        action="store_true",
+        help="Launch interactive arrow-key menu (single-level)",
+    )
 
     args = parser.parse_args()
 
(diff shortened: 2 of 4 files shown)


realiti4:
Thanks, I think it is a great idea! I merged it.

PR #34, opened 2026-05-01
feat: make --email optional for --add-token

treerobin06:
Follow-up to #29 per https://github.com/realiti4/claude-swap/pull/29#issuecomment-4358859871.

Setup-token registration via `cswap --add-token` carries no real email metadata — users obtain only the access-token string from `claude setup-token`. Requiring them to invent an email was noise.

## Changes
- `--email` is now optional. When omitted, the switcher synthesizes `setup-token-{slot}@token.local` (matches the existing `oauth-N@token.local` placeholder convention).
- Explicit `--email` still wins.
- Inverse guard: `--email` may now only be used with `--add-token`; using it on other commands surfaces a clear error.

## Examples
```
cswap --add-token sk-ant-oat01-...                  # → setup-token-1@token.local
cswap --add-token sk-ant-oat01-... --slot 3         # → setup-token-3@token.local
cswap --add-token sk-ant-oat01-... --email me@x.com # → me@x.com (explicit wins)
```

## Tests
- `test_add_token_without_email_dispatches_with_none`
- `test_email_without_add_token_errors`
- 5 new switcher tests for default-email derivation, blob propagation, slot uniqueness, and explicit override
- Full suite: 335 passed / 2 skipped

### DIFF
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -20,9 +20,10 @@ def main() -> None:
         epilog="""
 Examples:
   %(prog)s --add-account
-  %(prog)s --add-token --email user@example.com
-  %(prog)s --add-token sk-ant-oat01-... --email user@example.com --slot 3
-  %(prog)s --add-token - --email user@example.com
+  %(prog)s --add-token sk-ant-oat01-...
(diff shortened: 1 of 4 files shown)


realiti4:
Thank you, I've merged it.

PR #35, opened 2026-05-02
fix: tolerate throwaway artifacts in legacy→XDG migration

babs:
## Summary

Two-part fix for a race where the legacy → XDG backup-dir migration aborted with `Both legacy and new ... Refusing to merge` after a prior no-op run laid down `cache/` + `claude-swap.log` in the XDG path and a legacy directory appeared between runs (e.g. synced or restored from another machine).

- **`paths.migrate_legacy_backup_dir`** — collision check now distinguishes real data from throwaway artifacts (`cache/`, `claude-swap.log*`). When the target only contains the latter, wipe and migrate normally. Self-heals already-affected installs.
- **`logging_config`** — new `_LazyDirRotatingFileHandler` defers parent-dir creation until the first log emit. Read-only flows (e.g. `cswap --status` on a fresh box with no managed accounts) no longer materialize the XDG path, preventing the trap from re-arming for users without legacy data.

## Test plan

- [x] `pytest tests/test_paths.py` — 27 pass, including two new tests:
  - `test_target_with_only_throwaway_artifacts_is_wiped` — regression for the user-reported scenario
  - `test_target_with_real_data_alongside_artifacts_still_collides` — strict collision check still fires when real data is present alongside artifacts
- [x] `pytest tests/test_logging_config.py` — 3 new tests covering lazy dir creation
- [x] Full suite: 340 passed, 2 skipped

### DIFF
--- a/src/claude_swap/logging_config.py
+++ b/src/claude_swap/logging_config.py
@@ -5,8 +5,26 @@ from logging.handlers import RotatingFileHandler
 from pathlib import Path
 
 
+class _LazyDirRotatingFileHandler(RotatingFileHandler):
+    """RotatingFileHandler that creates its parent dir on first emit.
+
+    Keeps the backup root from being materialized just because the switcher
+    was instantiated. Necessary so a no-op run (e.g. ``cswap --status`` with
+    no managed accounts) doesn't lay down ``cache/`` or log files inside the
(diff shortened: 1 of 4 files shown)


realiti4:
Hi, thanks for the PR. It's looking good, I've merged it.

Issue #41, opened 2026-05-18
cswap --switch and --export fail with "Missing backup data" when an account has no credentials


stonelazy:
Description

  cswap --switch and cswap --export both fail when one of the managed accounts has no stored credentials ("no credentials" shown in
  cswap --list). The commands abort entirely rather than skipping the account or prompting the user to re-add credentials.

  Steps to Reproduce

  1. Have two managed accounts where one account has lost its stored credentials (shows "no credentials" in cswap --list)
  2. Run cswap --switch or cswap --export backup.cswap

  Observed Behavior

  cswap --switch:
  INFO: Backed up account 2
  ERROR: Switch failed: Missing backup data for Account-1, attempting rollback
  Error: Missing backup data for Account-1

  cswap --export backup.cswap:
  Error: no backup credentials found for account 1 (<redacted>)

  cswap --switch-to 1:
  INFO: Backed up account 2
  ERROR: Switch failed: Missing backup data for Account-1, attempting rollback
  Error: Missing backup data for Account-1

  The active account (Account-2) is backed up successfully, but the operation aborts when it encounters Account-1 which has no
  credentials.

  Expected Behavior

  - --switch should either skip accounts with missing credentials (with a warning) or prompt the user to re-add the account, rather
  than failing the entire operation.
  - --export should either skip accounts with missing credentials (with a warning) or respect --account to export only valid
  accounts.

  Environment

  - cswap version: 0.10.1
  - Installed via: uv tools
  - Python: 3.13.11
  - OS: macOS (Darwin 25.4.0 arm64)

  Notes

  It's unclear how the account ended up in this "no credentials" state. Regardless, having a managed account with missing
  credentials should not be a fatal state that blocks all switch and export operations.

realiti4:
Thank you for pointing this out. This should be fixed with 34d3108c86f3a29a314627334a8b1fb07317c3c8

`switch` and `export ` commands should skip now for missing config data with a console warning message.

realiti4:
Released with [v0.10.2](https://github.com/realiti4/claude-swap/releases/tag/v0.10.2). Closing the issue, please let me know if it doesn't resolve the problem.

stonelazy:
Thanks for the fix in v0.10.2 — the skip behavior now works correctly and no longer crashes.

[...]

PR #42, opened 2026-05-24
fix: fall back to file-based credentials when keyring uses null backend

stonelazy:
## Summary

- When `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring` is set (common for tools like DVC, poetry, etc.), the Python `keyring` library silently discards all writes and returns `None` on reads. This causes `cswap --add-account` to appear to succeed while credentials are never actually stored, making every subsequent `--switch` fail with "no stored credentials/config."
- This PR detects the null keyring backend at init time and automatically falls back to the same file-based credential storage already used on Linux/WSL.
- Adds 5 tests covering null backend detection, credential roundtrip, and deletion.
- Updates existing keyring contract tests to be compatible with the new fallback.

## Root cause

The null backend is typically activated via the `PYTHON_KEYRING_BACKEND` env var. Users may set this for unrelated Python tools (e.g. DVC, pip) without realizing it affects all `keyring` consumers, including `cswap`. The null backend has priority -1 and is specifically designed to be a no-op — it never raises errors, so the failure is completely silent.

## Related

Follow-up to #41 — the v0.10.2 fix made `--switch` skip broken slots gracefully, but the underlying credential loss was still happening for users with this env var set.

## Test plan

- [x] 5 new tests in `TestNullKeyringFallback` — all pass
- [x] Full test suite: zero new regressions (37 pre-existing failures unchanged)
- [x] Verified locally on macOS with `PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring` set

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -89,6 +89,36 @@ class ClaudeAccountSwitcher:
         self.credentials_dir = self.backup_dir / "credentials"
         self.lock_file = self.backup_dir / ".lock"
         self._logger = setup_logging(self.backup_dir, debug=debug)
+        self._use_file_credentials = self._should_use_file_credentials()
+
+    def _should_use_file_credentials(self) -> bool:
+        """Whether to use file-based credential storage instead of keyring.
+
+        Returns True on Linux/WSL (where file storage is always used) or
+        when the Python keyring resolves to a null/dummy backend that
+        silently discards writes (e.g. PYTHON_KEYRING_BACKEND is set to
+        keyring.backends.null.Keyring).
+        """
+        if self.platform in (Platform.LINUX, Platform.WSL):
+            return True
+        try:
+            from keyring.backends.null import Keyring as NullKeyring
+            backend = keyring.get_keyring()
+            if isinstance(backend, NullKeyring):
+                self._logger.warning(
+                    "Keyring resolved to null backend (PYTHON_KEYRING_BACKEND=%s); "
+                    "using file-based credential storage instead",
+                    os.environ.get("PYTHON_KEYRING_BACKEND", "<unset>"),
+                )
+                return True
+        except Exception:
+            self._logger.warning(
+                "Failed to detect keyring backend; "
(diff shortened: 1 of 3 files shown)


realiti4:
Hi, sorry for the late reply!

We've actually moved to file-based storage on Windows as well, since the Credentials Manager had some size limits that were causing issues.

With that in mind, I've been thinking about just dropping the keyring dependency entirely and doing our own Keychain operations instead. Wouldn't that fix this at the root? The null backend is a keyring thing, so talking to the Keychain directly would side-step it completely.

I'm a little hesitant about conditional storage detection in general, I feel like we'd end up needing to migrate data from one store to the other, and could leave orphaned data behind otherwise.

If dropping keyring does solve this, I'm already planning another PR for it that would address this automatically. Let me know what you think!

stonelazy:
Agreed - dropping keyring is the right root-cause fix, and your side-step reasoning is correct. This bug actually demonstrates it: in my environment the live-credential path (_read_credentials/_write_credentials, already using the security CLI) worked fine throughout - only the keyring-based backup path silently lost data. The security binary doesn't read PYTHON_KEYRING_BACKEND, so direct Keychain ops bypass the whole problem class by construction.

On the migration concern - I think it's lighter than it looks on macOS. keyring.backends.macOS and the security CLI are both generic-password items in the same login keychain, so it's an API swap over the same store, not a move between stores. I verified this locally: an entry written by keyring.set_password("claude-code", "account-N-email", …) is read back verbatim by security find-generic-password -s claude-code -a account-N-email, and security add-generic-password -U updates it in place (no duplicate/orphan). So as long as the service name (claude-code) and account format (account-N-email) stay the same, existing users' entries are picked up with no migration step.

Happy to close this in favor of your planned PR, or repurpose it to do the keyring removal if that saves you time - whichever you prefer.

realiti4:
With #51 now we are using native `security` command instead of keyring. Thanks again for pointing this out and the PR. Closing this in favor of #51

PR #50, opened 2026-06-07
Add auto-switch at usage limit (Beta) to the TUI

7GMA:
## What

Adds an opt-in **Auto-switch at limit (Beta)** entry to `cswap --tui`. When the active account's 5h/7d usage reaches a configurable threshold (default **95%**), a foreground monitor automatically rotates to the next managed account — using the existing `--switch` rotation logic, so no Claude Code restart is required.

The existing menu and behaviour are unchanged; this is purely an additional option.

## How it works

From `cswap --tui` → **Auto-switch at limit (Beta)**:

- **Enable / Disable** — persisted in `sequence.json` (`autoSwitch`), so the choice survives across runs.
- **Set threshold** — usage % that triggers a switch (1–100, default 95).
- **Start monitor now** — polls the active account's usage every 60s. On reaching the threshold it calls the same `switch()` rotation, then keeps watching the new active account. `s` checks immediately, `q`/`Esc` stops.

Because switching doesn't need a restart (Linux/Windows pick up the new credentials on the next message; macOS once the Keychain cache expires), the monitor works on a live session. It runs only while its TUI screen is open — there is **no background daemon**.

The status line shows `auto-switch ON (95%)` when enabled.

## Why Beta

It's a new foreground watcher and is labelled **(Beta)** throughout the UI and docs. Usage percentages come from the same API used by `cswap --list`.

## Implementation notes

- `switcher.py`
  - `get_auto_switch_config()` / `set_auto_switch_config(enabled=, threshold=)` — read/write settings in `sequence.json` (validates threshold 1–100; never clobbers accounts).
  - `get_active_usage_pct()` — highest 5h/7d utilization for the active account; reuses the existing short-lived usage cache; active account is never refreshed (`is_active=True`). The dollar-based `spend` entry is intentionally ignored.
  - `_max_usage_pct()` helper.
- `tui.py` — new settings/monitor sub-flow (`_do_auto_switch`, `_run_auto_monitor`, `_auto_perform_switch`, `_draw_monitor`), pure decision helper `_should_auto_switch`, and a status-line indicator.
- No new runtime dependencies (stdlib `curses`/`time` only).

## Tests

26 new tests in `tests/test_auto_switch.py` covering config persistence, usage-pct resolution, the decision helper, the settings sub-flow, and the monitor loop (switch / no-switch paths). Full suite: **407 passed, 2 skipped**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,33 @@ cswap --switch-to user@example.com
 
 **Note:** You usually don't need to restart — on Linux/Windows the new account is picked up automatically, and on macOS after the Keychain cache expires. To apply it instantly, restart Claude Code or reopen the VS Code extension tab. See [Tips](#tips) for the per-platform details.
 
+### Auto-switch at usage limit (Beta)
+
+Launch the interactive menu and open **Auto-switch at limit (Beta)**:
+
+```bash
+cswap --tui
+```
+
+From there you can:
+
+- **Enable/Disable** automatic switching (the setting persists across runs).
+- **Set threshold** — the usage percentage that triggers a switch (default `95%`).
+- **Start monitor now** — runs a foreground watcher that polls the active
+  account's 5h/7d usage every 60 seconds. When usage reaches the threshold, it
+  rotates to the next managed account (same rotation as `cswap --switch`), then
+  keeps watching the new account. Press `s` to check immediately, or `q`/`Esc`
+  to stop.
+
+Because switching doesn't require a Claude Code restart (see the note above),
+the new account takes effect on your next message — on macOS once the Keychain
+cache expires. The monitor runs only while its screen is open in the TUI; there
+is no background daemon.
+
+> **Beta:** this feature is new and runs as a foreground watcher. The usage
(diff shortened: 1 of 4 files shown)


realiti4:
Hi, thanks for the PR! I’ve also been considering a TUI overhaul and including auto-switching monitor there. I think we can use your implementation as the base for an initial beta.

Before we merge could you adjust couple of things though:

Could you disable this on macOS? Claude Code reads credentials from Keychain and may cache them for around 30 seconds, so we cannot guarantee that a running process will use the newly selected account immediately. I'll look into this about what we can do without being too hacky.

Let's use the new `switch(strategy="next-available")` strategy so the monitor skips accounts that are already at their usage limit.

Can you also rebase onto the latest main and resolve the conflicts?

realiti4:
Hi, thank you for the PR! I'd been working on this version of auto switcher myself #81 which also takes claude code .claude.lock into account for safety.

Since that is merged, I'll close this in favor of that. Thanks again.

Issue #19, opened 2026-04-16
Support for different claude code account for different vscode window?


MaxwellFang:
Hi this tool is great, I wonder if there is potentially ways to support different cc account for different vscode (instances?)

realiti4:
Hi, thanks. I am not sure if we can do that for vscode windows or tabs, without being too hacky, but for claude cli it has been on my plan and currently a work in progress. I'm thinking of making it possible to start sessions with a different account like this:

`cswap run 1 (account number)`

I will let you know when there is progress.

hhourani27:
Thank you ! this feature would be extremely helpful for us!

realiti4:
> Hi, thanks. I am not sure if we can do that for vscode windows or tabs, without being too hacky, but for claude cli it has been on my plan and currently a work in progress. I'm thinking of making it possible to start sessions with a different account like this:
> 
> `cswap run 1 (account number)`
> 
> I will let you know when there is progress.

Sorry for the wait! The new session mode feature should support this with #52. As I mentioned, per-tab sessions in VS Code are probably not possible, but you can now do this with the Claude CLI:

`cswap run 3` using the account id or email address. So VS Code stays on the default active account, and for other parallel tasks, CLI sessions can be started.

More details will be in the README as well.

realiti4:
This is also released now in [v0.13.0](https://github.com/realiti4/claude-swap/releases/tag/v0.13.0).

Since it has landed, I'm closing this, but please let me know if you run into any issues.

PR #54, opened 2026-06-13
Fix Windows `cswap run` always failing session validation

hhourani27:
Hello, I tried running `cswap run` on my windows PC and it threw an error. After debugging (with the help of Claude), I am submitting this PR.

## Problem
On Windows, `cswap run <account>` always fails with "failed validation", no matter the account. This makes the session feature (#52) unusable on Windows. Relates to #19.

## Cause
`_is_session_valid` checks the profile by running `subprocess.run(["claude", "auth", "status", "--json"])`. On Windows, `claude` is a `.cmd` shim, and a bare `"claude"` doesn't resolve there (subprocess doesn't apply `PATHEXT`), so it raises `FileNotFoundError`. That error is caught and treated as "not valid", so cswap deletes the new profile and reports failure.

This is why `--switch` works (it never runs `claude`) and why `run` works on macOS/Linux (where `claude` resolves without an extension). `run()`'s own launch already used `shutil.which`; only this validation check used the bare name.

## Fix
Resolve `claude` with `shutil.which` (which applies `PATHEXT`) before running it.

## Testing
- Added a regression test checking the probe uses the resolved path.
- Verified on Windows: `cswap run` now passes validation and launches the correct account in session mode.

### DIFF
--- a/src/claude_swap/session.py
+++ b/src/claude_swap/session.py
@@ -410,9 +410,14 @@ class SessionManager:
         """
         if not session_dir.is_dir():
             return False
+        # On Windows `claude` is a `.cmd` shim, and a bare "claude" passed to
+        # subprocess won't resolve it (PATHEXT isn't applied) — it raises
+        # FileNotFoundError, which the handler below turns into a false
+        # "failed validation". shutil.which finds the shim.
+        claude_bin = shutil.which("claude") or "claude"
         try:
             result = subprocess.run(
-                ["claude", "auth", "status", "--json"],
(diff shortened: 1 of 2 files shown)


realiti4:
Hi, thanks for the fix! In my Windows test, it looks like it resolved to claude.exe, so I didn’t experience the issue. However, as you found out, claude.cmd can cause issues on Windows.

I've merged it. Also releasing as [v0.13.1](https://github.com/realiti4/claude-swap/releases/tag/v0.13.1)

PR #55, opened 2026-06-14
Fix Windows hang: Platform.detect() blocks on WMI query

SoetZombie:
## Problem

On Windows, every `cswap` command hangs indefinitely (no output, no error). It never gets past startup.

## Root cause

`Platform.detect()` in `src/claude_swap/models.py` calls `platform.system()`. On Windows that resolves through `platform.uname()` → `win32_ver()` → `platform._wmi_query()`. On machines where the WMI service is slow or unresponsive, that query blocks forever.

`Platform.detect()` runs at import time (`cache.py` imports `get_backup_root()`, which calls it), so the hang affects **every** command, not just account switching.

Reproduced on Windows 11 with the uv-managed CPython 3.14 runtime. A bare `python -c "import platform; platform.system()"` never returns, while `sys.platform` returns instantly.

## Fix

Use `sys.platform` for detection instead of `platform.system()`. It never touches WMI, and it's already what the rest of the codebase uses for Windows checks (`locking.py`, `switcher.py`, etc.), so this also makes detection consistent.

- `darwin` → macOS
- `win32` → Windows
- `linux*` → Linux (or WSL when `WSL_DISTRO_NAME` is set)

The platform-detection tests are updated to patch `sys.platform` accordingly.

## Testing

- `cswap --status` (and other commands) now return immediately instead of hanging.
- `TestPlatformDetection` passes (5/5).

### DIFF
--- a/src/claude_swap/models.py
+++ b/src/claude_swap/models.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import json
 import os
-import platform as platform_module
 import sys
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


realiti4:
Hi, thanks for the fix. Looking good, I've merged it.

PR #57, opened 2026-06-15
Add event-driven multi-account load balancer + cmux integration (supersedes #50)

GoatInAHat:
## Supersedes #50 — auto-switch + an event-driven load balancer + cmux

This builds on **#50 (Add auto-switch at usage limit, by @7GMA)** and supersedes it: it includes that work (with the README conflict resolved) and grows it into a full multi-account **load balancer**. Credit to @7GMA for the original auto-switch feature — its threshold/usage model is carried forward.

### What's new

**Event-driven load balancer (Beta).** Run multiple Claude Code sessions across multiple accounts. When a session's account nears its usage limit, cswap migrates that session to a **higher-priority account with headroom** (concentrating load on top accounts, spilling to the next tier only when one fills), while **minimizing migrations** (each re-bills the context window). Single-subscription users instead get **pause + auto-resume** via `claude --resume` when the window resets, so in-flight workflows survive the limit.

- **Fully event-driven** — the in-session statusline reports usage on each message and triggers a churn-minimizing rebalance. No polling loop, no background daemon (the prior 60s auto-switch monitor is removed).
- **Credential-safe** — the statusline only writes registry *intents*; each session's supervisor owns its own profile and re-points only its own credentials (verbatim seed, no refresh side-effects); managed profiles live under a separate root from `cswap run`.
- New: `cswap launch`, `cswap --install` (also runs on upgrade via a migration), `cswap --status` embed-health, `cswap --balance` dashboard, per-account `cswap --set-priority`. QoL defaults for managed sessions (latest model, `--dangerously-skip-permissions`, high effort).

**cmux integration (Beta, macOS).** `cswap cmux setup` adds a "Balanced Claude (cswap)" command to manaflow-ai/cmux; `cswap cmux <N>` fans out N panes, each landing on a different account. `cswap launch` is cmux-wrapper-aware (preserves `CLAUDE_CONFIG_DIR`). No fork — CLI + config only.

### Testing
- **586 unit tests** (pure balancer is table-driven; registry/statusline/supervisor/embed/cmux covered).
- **Live E2E** on real accounts: managed launch authenticates on the right subscription account, statusline populates the registry, migration intent + per-account re-point verified, vanilla isolation held (`~/.claude/settings.json` byte-identical).
- **Two adversarial review rounds** (a local multi-agent pass + a cloud ultrareview): every confirmed finding fixed, each with a regression test — including the auto-resume session-id, the cross-process account-spread reservation, atomic credential writes, and the pause/resume livelock.

### Notes
- Beta. The cmux fanout's live multi-pane spawn needs cmux socket control enabled (`Settings → Automation`); the config surface + the balancer spread work regardless.
- "ultracode" effort is session-only, so managed sessions use the persistable `xhigh` level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,32 +82,87 @@ cswap run 2 --no-share          # don't share your ~/.claude customizations
 
 Your `~/.claude` customizations (settings, keybindings, CLAUDE.md, skills, commands, agents) are shared into the session by default — use `--no-share` for a bare profile. Conversation history stays per-account.
 
-### Auto-switch at usage limit (Beta)
+### Load balancer (Beta)
 
-Launch the interactive menu and open **Auto-switch at limit (Beta)**:
+Run several Claude Code sessions across all your accounts and let cswap keep them
+fed. When a session's account nears its usage limit, cswap migrates that session
+to a higher-priority account that still has headroom — concentrating load on your
+top accounts and only spilling to the next tier when one fills up. Migrations are
+minimized (each one re-bills the context window), so a session only moves when its
+account is actually exhausted, never for a marginal gain.
+
+Only one subscription? There's nowhere to migrate, so cswap **pauses** the session
+instead and **auto-resumes** it (via `claude --resume`) the moment the limit
+window resets — your in-flight, dynamic workflow survives the limit instead of
+being thrown away.
+
+It's fully **event-driven**: the in-session statusline reports usage on each
+message, which is what triggers a rebalance. There's no polling loop and no
+background daemon. It's also credential-safe — each session's supervisor owns its
+own profile and only ever re-points its own credentials.
+
+#### Setup
 
 ```bash
-cswap --tui
+cswap --install                 # embed cswap into Claude Code (one-time, idempotent)
+cswap --status                  # shows load-balancer state + embed health
 ```
 
-From there you can:
+`--install` installs the statusline and QoL layer via a non-shared
+`settings.local.json` inside each managed profile, so plain `claude` and
+`cswap run` stay completely vanilla. It also runs automatically on upgrade.
 
-- **Enable/Disable** automatic switching (the setting persists across runs).
-- **Set threshold** — the usage percentage that triggers a switch (default `95%`).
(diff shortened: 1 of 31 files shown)


GoatInAHat:
Pushed `831495b` adding two fixes for managed/balanced sessions (both surfaced by a real incident where two sessions stranded ~1h40m and lost history):

1. **Backup usage-headroom probe + intelligent re-probe cadence** — when the usage API 429-backs-off, the account was treated as zero-headroom for the whole window, so a stranded session couldn't resume onto a healthy account. Adds a `/v1/messages` capability probe (separate rate-limit bucket): 2xx⇒headroom, 429⇒capped, 401/other⇒unknown. Gated to subscription tiers (never bills PAYG/API), never probes the active/live login, claimed under the lock to throttle the cross-process herd, ~120s cadence. Opt-in via `build_world(probe_unavailable=)` on the supervisor's resume/strand/auth-recovery paths only. Enters the pure balancer as `signal="probe"` with a conservative `PROBE_CONFIRMED_PCT` (admits one stranded session; the online reservation blocks a second from co-exhausting it).

2. **Force session-transcript persistence under cmux** — cmux sets `CLAUDE_CODE_CHILD_SESSION=1`, which makes Claude Code skip writing the main `<sessionId>.jsonl` for an interactive non-tmux child session, so managed sessions couldn't be `--resume`d and chat history was silently lost. Sets `CLAUDE_CODE_FORCE_SESSION_PERSISTENCE=1` in the supervisor child env (verified end-to-end on cmux against Claude Code 2.1.178; cmux wrapper passes the key through).

794 passed, 3 skipped. Adversarially reviewed (6 blocker/major findings fixed).

EdenGibson:
+1 this is cool.

realiti4:
Hi, thanks a lot for the PR! I'd been wanting to build this one myself, and it's now merged as #81 which also takes claude code `.claude.lock` into account for safety.

Since that is merged, I'll close this in favor of that. But I'd happily accept new PRs for new ideas on top of the new AutoSwitchEngine. Thanks again!

PR #59, opened 2026-06-18
feat: usage-aware account switching (--switch --strategy)

quicksandznzn:
## Summary

  Adds two opt-in modifiers to `cswap --switch` that pick the target account by **remaining rate-limit quota** instead of rotating blindly:

  - **`--switch --best`** — switch to the account with the most remaining 5h/7d headroom. Never moves you onto an account worse than your current one: if you already hold the most quota (or every other
  account is at its limit), it stays put and says so.
  - **`--switch --skip-exhausted`** — rotate to the next account, skipping any at their 5h/7d limit; if all rotation targets are exhausted, it stays on the current account rather than switching onto a
  maxed-out one.

  Headroom is derived from the 5-hour and 7-day utilization windows already fetched for `--list` (pay-as-you-go spend is ignored). Decisions use the live login identity and fall back to plain rotation when
  usage data can't be fetched, so the feature **never blocks a switch**. Both modifiers apply to the normal path with an active login; the fresh-machine path (no login yet, e.g. right after `--import`) is
  unchanged.

  ## Examples

  ```bash
  # Jump to the account with the most remaining 5h/7d quota
  cswap --switch --best

  # Rotate to the next account, skipping any at their 5h/7d limit
  cswap --switch --skip-exhausted
  ```

  ### `--best`

  Switches when another account has strictly more headroom:

  ```text
  $ cswap --switch --best
  Switched to Account-3 (dev@example.com)
  Accounts:
    1: a@example.com   ├ 5h: 92%  └ 7d: 71%
    3: dev@example.com ├ 5h: 12%  └ 7d: 40%  (active)
  ```

  Stays put when you already hold the most quota (does **not** move you onto a worse account):

  ```text
  $ cswap --switch --best
  Already on the account with the most remaining quota (Account-2).
  ```

  When every account is maxed out, switching wouldn't help — it stays put:

  ```text
  $ cswap --switch --best
  All accounts are at their 5h/7d limit — staying on Account-2.
  ```

  If usage can't be fetched (e.g. offline), it falls back to plain rotation so the switch is never blocked:

  ```text
  $ cswap --switch --best
  Usage data unavailable — rotating to the next account.
  Switched to Account-2 (b@example.com)
  ```

  ### `--skip-exhausted`

  Skips rate-limited accounts during rotation:

  ```text
  $ cswap --switch --skip-exhausted
  Skipping Account-1 (at 5h/7d limit)
  Switched to Account-3 (dev@example.com)
  ```

  If every other account is exhausted, it stays on the current one instead of switching onto a maxed-out account:

  ```text
  $ cswap --switch --skip-exhausted
  All other accounts are at their 5h/7d limit — staying on Account-2.
  ```

  ### Validation

  The modifiers only make sense with `--switch`:

  ```text
  $ cswap --best --list
  cswap: error: --best and --skip-exhausted can only be used with --switch
  ```

  ## Implementation

  - **`oauth.account_headroom()`** — pure helper returning the binding window's remaining percentage (`100 - max(5h%, 7d%)`), or `None` when usage is unavailable / has no window data. `spend` is
  deliberately ignored.
  - **`switcher`** — extract `_build_accounts_info()` / `_collect_usage()` / `_usage_by_account()` out of `list_accounts` so usage fetching has a single cache-first path (reuses the existing `usage.json`
  15s TTL cache). Add `_select_best_switchable()` (compares against the current account's headroom; recommends a switch only when it lands on strictly more) and wire `best` / `skip_exhausted` into
  `switch()`.
  - **`cli`** — add `--best` / `--skip-exhausted` flags, validation requiring `--switch`, and help/epilog examples.
  - **`README`** — new "Switch by remaining usage" section.

  ## Behavior guarantees (all covered by tests)

  - Never switches you onto an account with less headroom than your current one.
  - Usage unavailable for all accounts → falls back to plain rotation (never blocks).
  - Every account at its limit → stays put with a clear message.
  - `--skip-exhausted` does **not** skip accounts whose usage is unknown (gives them a chance).
  - `skip_exhausted=False` path is byte-for-byte the existing rotation behavior.

  ## Testing

  - `oauth.account_headroom` unit tests (binding window, single window, at-limit, spend-ignored, malformed/None).
  - `switch()` selection + stay-put fallbacks for both `--best` and `--skip-exhausted`.
  - CLI flag validation and forwarding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,8 @@ cswap --switch-to user@example.com
 Let claude-swap pick the account for you based on how much quota is left, instead of rotating blindly:
 
 ```bash
-cswap --switch --best              # jump to the account with the most 5h/7d quota left
-cswap --switch --skip-exhausted    # rotate to the next account, skipping any at their limit
+cswap --switch --strategy best             # jump to the account with the most 5h/7d quota left
+cswap --switch --strategy next-available   # rotate to the next account, skipping any at their limit
 ```
 
 These use the same 5-hour and 7-day usage windows shown by `--list` (pay-as-you-go spend is ignored). They never move you onto an account that's worse than the one you're on: if you already hold the most quota, or every other account is at its limit, claude-swap stays put and tells you. If usage can't be fetched at all, it falls back to plain rotation. They apply when you have an active Claude login; right after a fresh `--import` (no login yet) they're ignored.
@@ -108,8 +108,8 @@ This will update the stored credentials without creating a duplicate.
 ```bash
 cswap run 2                     # Run an account in this terminal only (session mode)
 cswap --list                    # Show all accounts with 5h/7d usage and reset times
-cswap --switch --best           # Switch to the account with the most quota left
-cswap --switch --skip-exhausted # Rotate to the next account, skipping rate-limited ones
+cswap --switch --strategy best           # Switch to the account with the most quota left
+cswap --switch --strategy next-available # Rotate to the next account, skipping rate-limited ones
 cswap --status                  # Show current account
 cswap --add-account --slot 3    # Add account to a specific slot (prompts before overwrite)
 cswap --remove-account 2        # Remove an account
--- a/README.md
+++ b/README.md
@@ -67,6 +67,17 @@ cswap --switch-to 2
 cswap --switch-to user@example.com
 ```
 
+### Switch by remaining usage
+
+Let claude-swap pick the account for you based on how much quota is left, instead of rotating blindly:
+
+```bash
+cswap --switch --best              # jump to the account with the most 5h/7d quota left
+cswap --switch --skip-exhausted    # rotate to the next account, skipping any at their limit
+```
+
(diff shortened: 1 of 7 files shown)


realiti4:
Hi, thanks for the PR great idea! One thought on the interface: could we
use a single `--strategy` argument instead of two flags? Like:

```
cswap --switch
cswap --switch --strategy best
cswap --switch --strategy next-available
```

I think it fits a bit better, and it also avoids the case where passing both
`--best` and `--skip-exhausted` lets `best` silently win. Leaves room to add
more strategies later too.

quicksandznzn:
> Hi, thanks for the PR great idea! One thought on the interface: could we use a single `--strategy` argument instead of two flags? Like:
> 
> ```
> cswap --switch
> cswap --switch --strategy best
> cswap --switch --strategy next-available
> ```
> 
> I think it fits a bit better, and it also avoids the case where passing both `--best` and `--skip-exhausted` lets `best` silently win. Leaves room to add more strategies later too.

Thanks for the thoughtful review! Great point — I've reworked the interface accordingly.

The two boolean flags are now consolidated into a single `--strategy` argument:

```bash
cswap --switch                            # plain rotation (unchanged)
cswap --switch --strategy best            # jump to the account with the most 5h/7d headroom
cswap --switch --strategy next-available  # rotate to the next account, skipping any at its limit
```

This makes the strategies **mutually exclusive by construction** (argparse `choices`), so the ambiguous `--best --skip-exhausted` case where `best` silently won is gone entirely. It also leaves a clean extension point for adding more strategies down the line.

realiti4:
Thanks, looking good I've merged it.

Issue #60, opened 2026-06-19
macOS: support file-based credentials (headless / no-GUI sessions can't use Keychain)


kayinchang:
## Summary

On macOS, `cswap` reads and writes Claude Code credentials exclusively via the login Keychain (`switcher.py` `_read_credentials` / `_write_credentials`, gated on `self.platform == Platform.MACOS`). But Claude Code on macOS falls back to a plaintext file `~/.claude/.credentials.json` whenever the Keychain is unavailable — which is the normal state in **headless / non-GUI sessions**: remote VS Code Server, SSH, CI, or a worker account whose GUI/Aqua session belongs to a different user. In that mode every `cswap` command fails.

## Symptom

```
$ cswap --add-account
Error: No credentials found for current account
```

`_read_credentials()` returns `""` because the Keychain has no `Claude Code-credentials` item — the credentials actually live in `~/.claude/.credentials.json`.

## Root cause

The account has no GUI/Aqua session, so the login Keychain is unreachable. `security` confirms it — even reading the keychain's own settings is denied:

```
$ security show-keychain-info ~/Library/Keychains/login.keychain-db
security: ... User interaction is not allowed.

$ security add-generic-password -s probe -a "$USER" -w x
security: ... User interaction is not allowed.
```

So Claude Code stores its credentials in the plaintext file (chmod 600). cswap's file backend — already used on Linux/WSL/Windows — reads and writes that exact file correctly; the macOS path just never routes to it.

## Requested fix

Add an opt-in way to use the file backend on macOS, mirroring the existing `CLAUDE_CONFIG_DIR` env handling. Either:

- Honour an env var, e.g. `CSWAP_FILE_CREDENTIALS=1`, or
- Auto-detect: on macOS, if the Keychain returns `errSecInteractionNotAllowed` (or `~/.claude/.credentials.json` exists and there is no Keychain item), use the file backend.

All credential routing keys off the single `self.platform` value, so the change is small. As a local workaround I patched `__init__` to set `self.platform = Platform.LINUX` on macOS when `get_credentials_path().exists()`, and `--add-account` / `--status` then work correctly against the file.

## Environment

- macOS (Apple Silicon)
- Claude Code via VS Code remote server
- `cswap 0.13.2`
- Account logged in over a remote tty; the GUI console session is owned by a different user.

realiti4:
Thanks for pointing this out. I’ll look into a solution. I think we can auto-detect this and fall back to the file backend on macOS as well.

[...]

PR #64, opened 2026-06-22
feat(macos): opt into file-based credential storage via CSWAP_FILE_STORAGE

kathar0s:
## Problem

On a headless or remote **macOS** host (SSH/mosh, CI, a detached daemon), the login Keychain is often locked or unreachable. The `security` CLI then prompts for an unlock that never gets answered — or fails outright — so cswap can report `no stored credentials/config` even when backups exist, and switches don't take effect.

Claude Code itself already falls back to `~/.claude/.credentials.json` in that situation, so cswap and Claude Code can end up reading from different stores on the same machine.

## Change

Add a `_prefer_file_storage()` gate. Setting `CSWAP_FILE_STORAGE=1` makes macOS use **file storage** — active credentials (`~/.claude/.credentials.json`) and `.enc` per-account backups — exactly like Linux/WSL/Windows, keeping cswap in lockstep with Claude Code.

Routed through the gate:
- `_read_credentials` / `_write_credentials` (active credential)
- `_uses_file_backup_backend` (per-account backup backend)
- `_print_switch_followup` (the file-storage "no restart needed" note now applies on macOS too)

**Opt-in only** — the Keychain remains the macOS default, so the carefully engineered silent-`security` path is untouched for everyone who doesn't set the variable. Linux/WSL/Windows always use files and ignore the variable.

## Notes / caveats

- File-stored and Keychain-stored credentials live in different places, so the variable should be set from the start (or accounts re-added afterwards). Documented in the README.
- No default behavior changes.

## Tests

Added `TestMacosFileStorageOptIn` covering the gate (default off, opt-in on, non-macOS always-on), file round-trips for active + backup credentials, Keychain bypass, and the followup message. Full suite: **512 passed, 3 skipped**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,12 +123,30 @@ cswap --purge                   # Remove all claude-swap data
 | Platform | Credentials | Config backups |
 |----------|-------------|----------------|
 | Windows | File-based (inside the backup directory, under `credentials/`) | `~/.claude-swap-backup/` |
-| macOS | macOS Keychain | `~/.claude-swap-backup/` |
+| macOS | macOS Keychain (or file-based with `CSWAP_FILE_STORAGE`) | `~/.claude-swap-backup/` |
 | Linux / WSL | File-based (inside the backup directory, under `credentials/`) | `${XDG_DATA_HOME:-~/.local/share}/claude-swap/` |
 
 Session-mode profiles (`cswap run`) live under the backup directory in `sessions/`.
 
 On Linux/WSL, set `XDG_DATA_HOME` to override the default location. Data from older installs under `~/.claude-swap-backup/` is migrated automatically on first run.
+
+### macOS file-based storage (`CSWAP_FILE_STORAGE`)
+
+macOS stores credentials in the login Keychain by default. On a headless or remote
+macOS host (SSH/mosh, CI, a detached daemon) the Keychain is often locked or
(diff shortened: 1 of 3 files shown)


realiti4:
Hi, thank you for the PR! I was also working on this in #66, trying to address issue #60.

I think it's better to mirror what Claude Code does here. Claude Code stores its credentials either in the Keychain or in a file, so this proposed env variable will either happen to match that or not, exposing the choice to the user makes it more fragile, I believe. From what I checked, Claude Code itself doesn't provide a way to override this behaviour either.

And by the same logic, we can decide where to store our own backup data automatically too, so there's no need to separate Claude Code's storage from our backend behind an option.

realiti4:
> Hi, thank you for the PR! I was also working on this in #66, trying to address issue #60.
> 
> I think it's better to mirror what Claude Code does here. Claude Code stores its credentials either in the Keychain or in a file, so this proposed env variable will either happen to match that or not, exposing the choice to the user makes it more fragile, I believe. From what I checked, Claude Code itself doesn't provide a way to override this behaviour either.
> 
> And by the same logic, we can decide where to store our own backup data automatically too, so there's no need to separate Claude Code's storage from our backend behind an option.

Thanks again for the PR, I am closing this in favor of #66 since that is merged.

PR #65, opened 2026-06-22
feat: macOS menu bar app with usage stats, account switching, and auto-switch

deathemperor:
## Summary

Adds an optional **macOS menu bar app** to claude-swap (`cswap --menubar`) that surfaces each managed account's usage and lets you switch with a click, plus an opt-in **auto-switcher** that moves you to a fresh account when the active one runs low. It's a thin GUI shell over the existing `ClaudeAccountSwitcher` — no account/usage/switch logic is re-implemented — and ships behind an optional extra so non-macOS users and the base install are unaffected.

## Menu bar (`cswap --menubar`)

- **Live usage** per account (5h / 7d / spend), mirroring `--list`, refreshed on a configurable timer over the existing 15s usage cache.
- **Switching parity** with the CLI: click any account to switch to it, plus Rotate to next / Switch to best / Next available.
- **TUI action parity**: add account (from current login or a setup-token), remove account (with confirmation), refresh current credentials.
- **Menu-bar title** shows the active account's email local-part and tightest quota %, e.g. `⇄ loc · 42%`.
- **Settings submenu**: toggle title content, refresh interval, and launch-at-login (LaunchAgent plist).

## Auto-switch (opt-in)

When enabled, the app switches automatically once the active account crosses a usage threshold:

- Trigger when the active account hits the threshold on its 5h or 7d window; target the account with the most headroom, skipping any that are themselves at the threshold.
- **Cooldown** between switches (manual switches feed the same cooldown); a single rate-limited "no fresh account" notification when nothing has headroom.
- Configurable from Settings: **Threshold** (80/90/95%), **Cooldown** (5m/10m/30m), **Check** (with display refresh, or an independent 1m/3m/5m cadence).
- Off by default; defaults are 95% / 10m, computed in-process from the same usage data the menu shows.

## Install

```bash
uv tool install 'claude-swap[menubar]'   # or: pipx install 'claude-swap[menubar]'
cswap --menubar
```

## Implementation notes

- New module `src/claude_swap/menubar.py`; one new flag (`--menubar`) wired into the CLI alongside `--tui`, with a clean error on non-macOS and a helpful hint when `rumps` is absent.
- `rumps` is an **optional extra** (`menubar = ["rumps>=0.4.0"]`), imported lazily inside the launcher — the module imports fine without it, so CI and other platforms are untouched.
- Pure helpers (settings, formatters, the auto-switch decision/planning logic, plist rendering) are unit-tested; the AppKit glue is verified manually since it can't run headless.
- Thread-safety: usage is fetched on a worker thread into an in-memory snapshot; all AppKit mutation and switching happen on the main thread.

## Testing

- `pytest`: **552 passed, 3 skipped**. New tests cover the settings/state stores and the full auto-switch decision matrix (thresholds, best-candidate selection, tie-breaks, unknown-usage exclusion, cooldown, notification rate-limiting).
- The menu bar and auto-switch were exercised manually on macOS (rendering, click-to-switch, notifications, settings persistence, launch-at-login, threshold-triggered auto-switch).


realiti4:
Hi @deathemperor, thanks for the PR, and sorry for the slow response. I really like the idea, and it would be really nice to have.

I was planning to work on the auto-switch feature myself, so since you opened this, we landed the auto-switch engine in `#81` with `cswap auto`, plus a per-account usage store in `#88`. Those changes also respect Claude’s credentials lock file for safety and use adaptive usage polling to avoid the rate limiter.

That means two parts of this PR now overlap with `main`:

The auto-switch engine inside `menubar.py` duplicates `cswap auto`.

The `switcher.py` / `oauth.py` usage-fetch changes, including `429` backoff, last-known-good usage, and partial fetch handling, are covered by the usage store in `#88`.

What do you think about dropping the auto-switch engine from this PR and making it menu-bar-only, wired through the new engine from `#81`? So the path would be: rebase on `main`, then make the menu bar’s auto-switch a toggle that drives the merged engine instead, sharing `autoswitch_state.json` and the `autoswitch.*` keys from the `cswap` config. The menu bar would keep only its display settings.

That makes it the “thin GUI shell” your docstring describes, and keeps the door open for other platforms later.

Also, your consume-first strategy is a great idea, but I don’t think it should be buried in the GUI. Could we add it to the core engine’s `--strategy` flag so it can be used from the CLI as well? You can add it to this PR or open a new one; either way is appreciated.

Overall really nice work, thanks again!

deathemperor:
Thanks for the detailed direction — agreed on all of it. The engine in #81 and the usage store in #88 are the right homes for this, and the overlap should go.

Plan for this PR:

- Rebase on `main` and drop the duplicated auto-switch engine and the `switcher.py`/`oauth.py` usage-fetch changes (429 backoff / last-known-good / partial fetch). Usage display will read from `UsageStore`.
- Rewire the menu bar as a thin GUI shell: an auto-switch **enable toggle** plus threshold/strategy that write `autoswitch.*` settings and let `AutoSwitchEngine` drive decisions off the shared `autoswitch_state.json`. The menu bar keeps only its display settings (account name, title %, per-account usage logging).

I'll bring `consume-first` to the core engine's `--strategy` in a **separate PR** so this one stays a clean menu-bar-only change.

Starting the rebase now.

realiti4:
Thanks for the adjustments, looking good.

One thing I noticed: it looks like the agent `.md` files were pushed too, adding around ~3,700 lines. Can we drop the `docs` folder entirely before the merge?

deathemperor:
@realiti4 My apology about the docs folder, I removed it.

realiti4:
Merged it, thanks for the adjustments!

PR #69, opened 2026-06-23
feat: macOS auto-switch — reactive + proactive consume-first strategies

ynaamane:
## Summary

Adds an opt-in, **macOS-only auto-switch** feature to `cswap`: a background monitor (a launchd daemon, or a foreground `cswap watch`) that switches the default Claude account on its own based on remaining usage, so you don't hit a hard wall mid-work.

Two strategies decide **when** to switch (`cswap auto on --strategy <name>`):

- **`consume-first`** (default for new installs) — *proactive*. Keeps you on the account to consume **first**: the available account whose **weekly window resets soonest** (use‑it‑or‑lose‑it, so quota isn't wasted at reset). It switches whenever that account changes — a reset re‑orders the queue, the active account exhausts, or an account that hit its 5‑hour limit clears and reclaims the soonest‑reset slot. If the chosen account hits its 5‑hour session limit while its weekly still has room, it moves to another account temporarily and returns **promptly** once the 5‑hour window resets.
- **`reactive`** — *threshold‑only* (the original behaviour). Stays put until the active account crosses a limit (5h ≥ 98% or 7d ≥ 99%), then jumps to the soonest‑reset account that still has headroom.

## What's included

- `cswap auto on | off | status` (installs/loads a launchd LaunchAgent) and `cswap watch` (foreground), both accepting `--strategy {reactive,consume-first}`, `--session-threshold`, `--weekly-threshold`, `--no-notify`.
- **Connection‑loss fallback** — never switches on stale/unknown data; one‑shot "offline" / "back online" macOS notifications; `cswap auto status` surfaces the monitoring state and a per‑account last‑known‑usage view.
- macOS desktop notifications on a switch and when all accounts are exhausted; a single‑flight lock so the daemon and a manual `cswap watch` never both drive switches; the switch itself reuses the existing transactional `_perform_switch` (with rollback). Only the **default** login is ever changed — `cswap run` session profiles are untouched.

## Switch before the hard wall, prompt re‑rank, no thrash

- **Critical band (switch before 100%).** As the active account's **5‑hour** window nears the limit the cadence tightens to a brief `critical_interval` (default **15s**) so the daemon observes the threshold crossing and flips the default login **before** the hard 100% wall — which would otherwise abort the in‑flight request and force a restart. The switch threshold stays 98%; an in‑flight conversation continues on your **next message** (it isn't migrated mid‑stream), so the tight poll is what makes that next message land on a fresh account.
- **Reset‑aware scheduler.** Wakes shortly after the soonest relevant reset — every account's **7d** reset (re‑rank the queue) **and** a 5h‑blocked account's **5h** reset (it becomes available again) — so a freed‑up account is reclaimed within seconds, not after a full poll interval.
- **Wake from sleep.** The monitor sleeps in ≤30s chunks; when a chunk's wall‑clock overshoots by >60s the process was suspended (the Mac slept), so it re‑polls immediately on wake rather than waiting out a stale interval (a scheduled re‑rank can't fire while suspended).
- **Hysteresis** on the 5‑hour axis prevents available↔blocked thrash around the threshold.

## Resilient under rate‑limit contention

The **active** account is the one Claude Code is actively using *and* polling for its statusline, so the daemon's own usage call on it contends with those and is the one that 429s under load.

- **Reuse the statusline cache.** If a CodexBar‑compatible statusline already caches usage (default `usage_cache_file = /tmp/claude/statusline-usage-cache.json`), the daemon **reuses it for the active account** instead of a redundant call — but only when fresh (≤90s) **and** its 7‑day reset matches the active account's last‑known reset (an account‑identity guard, with a 60s tolerance for the API's sub‑second `resets_at` jitter). Otherwise it **falls back to a direct fetch**, so a setup with no statusline works unchanged. Set `usage_cache_file` to `""` to disable.
- **Fast offline recovery.** With a usage cache configured the daemon retries offline at the floor cadence (the cache is a cheap, API‑free recovery path) instead of an exponential backoff, so a transient 429 doesn't leave it "offline" for minutes. With no cache the original backoff is kept.
- **Inactive accounts** show their **last‑known** usage: the daemon never refreshes an expired inactive token (an OAuth refresh rotates a one‑time token, and a rotation it can't persist — a locked Keychain while asleep — would brick the account). A foreground `cswap --list` / switch refreshes them. The fast retry of an unreadable peer is **bounded** so a persistently unreadable peer can't pin the floor cadence.

## Design notes

- The decision logic is a set of **pure, side‑effect‑free functions** (`decide_switch`, `decide_consume_first`) with a shared ranking key, so the two strategies can't drift.
- The weekly window is treated as a **fixed reset** (utilisation drops cleanly back to ~0 at its reset time), which is what makes "consume the soonest‑resetting account first" worthwhile.
- macOS‑only behaviour; the pure decision engine is cross‑platform and unit‑tested everywhere. **No new dependencies.**

## Testing

- **845** unit/integration tests pass (pure decision tables, the hysteresis FSM, the offline state machine, the adaptive + critical‑band cadence, the reset‑aware scheduler, the chunked wake‑detection sleep, the statusline‑cache reuse + identity guard, launchd plist generation, the CLI surface). ~89–93% line coverage on the new modules.
- The launchd install/run/uninstall lifecycle and a **real consume‑first switch** were validated live on macOS, and the statusline‑cache reuse was live‑verified against the real cache (the active account's usage was served from it with the 7‑day reset matched).
- A 13‑scenario validation (each backed by an executed repro against the decision engine) plus a 3‑agent adversarial review covered the 5‑hour stories and the wake/retry logic.


realiti4:
Hi, thanks a lot for the PR! I'd been wanting to build this one myself, and it's now merged as #81 which also takes claude code .claude.lock into account for safety and it is also cross platform.

Since that is merged, I'll close this in favor of that. But I'd happily accept new PRs for new ideas on top of the new AutoSwitchEngine. Thanks again!

PR #71, opened 2026-06-24
feat: map accounts to directories for automatic per-repo selection

erik-sutton95:
## What & why

Adds **per-directory account mappings** so the right Claude account is selected automatically per repo/project — e.g. a personal account for one repo and a work account for another, without switching by hand each time.

`cswap map <NUM|EMAIL> [PATH]` binds a directory (default: current) to a stored account. `cswap run` with **no account argument** then auto-launches the account mapped to the current directory in session mode (per-terminal isolation, so two repos can run different accounts at the same time). The nearest mapped ancestor directory wins, so subfolders inherit. In an unmapped directory, `cswap run` falls back to launching plain `claude` with the current default login.

```bash
cswap map 2 ~/work/client-app     # map a directory to account 2
cswap map user@example.com        # map the current directory
cswap map                         # list mappings
cswap unmap ~/work/client-app     # remove a mapping

cd ~/work/client-app/src
cswap run                         # -> launches account 2 [session mode]
cd ~/scratch
cswap run                         # -> plain claude, current default login (unmapped)
```

## Design notes

- New `MappingStore` (`mappings.py`) persists `<backup_dir>/mappings.json`, keyed by a normalized absolute path and valued by the stable `(email, organizationUuid)` identity rather than the slot number, so mappings survive remove/re-add renumbering. Atomic `0600` writes; nearest-ancestor resolution uses `Path` semantics (so `/foo/bar` never matches `/foo/barbaz`).
- `map` / `unmap` are pre-dispatched in `cli.py` exactly like the existing `run` subcommand.
- `SessionManager.exec_default` handles the unmapped fallback — it passes the **unmodified** environment so it behaves exactly like typing `claude` (no auth-override scrubbing, unlike an explicit `cswap run N`).
- Mappings are intentionally local-only: excluded from `--export` / `--import` (paths are machine-specific) and removed by `--purge` (they live under the backup dir). `remove-account` prunes any mappings pointing at the removed account.
- `mappings.py` has no dependency on `switcher` (mirrors the existing `session.py` decoupling).

## Testing

- 35 new tests (`test_mappings.py`, plus additions to `test_cli.py`, `test_session.py`, `test_switcher.py`) covering nearest-ancestor / sibling-prefix resolution, path normalization (incl. Windows `normcase`), corrupt/missing file handling, account pruning, the `run` no-account resolution + fallback paths, and the root guards.
- Full suite: **601 passed, 3 skipped**.

## Notes

- No new dependencies (standard library only); no changes to existing behavior or the package version.
- Developed with the help of Claude Code (Claude Opus); I reviewed the full diff before opening this PR. Happy to adjust the command surface (e.g. `map`/`unmap` naming), behavior, or scope to fit the project's direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,27 @@ Run `cswap` on its own (or `cswap tui`) for the full-screen dashboard: live usag
 <img src="assets/tui-watch.png" width="760" alt="cswap watch — live 5h/7d usage bars for every account, with reset times and the active account marked">
 
 
+### Use a specific account per directory (auto-select)
+
+Map an account to a repo/project directory, then launch with no account argument
+and claude-swap picks the mapped account automatically — in session mode, so
+different repos can run different accounts at the same time.
+
+~~~bash
+cswap map 2 ~/work/client-app        # map a path to account 2 (work)
+cswap map user@example.com           # map the CURRENT directory to an account
+cswap map                            # list all mappings
+cswap unmap ~/work/client-app        # remove a mapping (defaults to cwd)
+
+cd ~/work/client-app/src
+cswap run                            # no account → launches account 2 [session mode]
+~~~
+
+The nearest mapped ancestor directory wins, so subfolders inherit their parent's
+mapping. In an unmapped directory, `cswap run` (with no account) just launches
+Claude Code with your current default login. Mappings are stored locally per
+machine and are not included in `--export` / `--import`.
+
 ### Refresh expired tokens
 
 If an account's token expires, log back into Claude Code with that account and re-run:
@@ -144,6 +165,10 @@ This will update the stored credentials without creating a duplicate.
 
 ```bash
 cswap run 2                     # Run an account in this terminal only (session mode)
+cswap run                       # Run the current dir's mapped account (or default login if unmapped)
+cswap map 2 ~/work/app          # Map a directory to an account
+cswap map                       # List directory→account mappings
+cswap unmap ~/work/app          # Remove a directory mapping
 cswap auto                      # Auto-switch when nearing rate limits (see above)
 cswap config                    # Show or edit settings (see Configuration below)
 cswap list                      # Show all accounts with 5h/7d usage and reset times
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -10,7 +10,14 @@ import sys
 from claude_swap import __version__
 from claude_swap.exceptions import ClaudeSwitchError
 from claude_swap.json_output import error_envelope
-from claude_swap.printer import dimmed, error, force_utf8_output, muted
+from claude_swap.printer import (
+    accent,
+    dimmed,
+    error,
+    force_utf8_output,
+    muted,
+    warning,
+)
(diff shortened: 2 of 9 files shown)


erik-sutton95:
@realiti4 Any thoughts? Worth merging?

realiti4:
Hi, thanks for the PR! The feature is useful and the implementation looks good overall. Could we adjust couple of things before merging?

The branch conflicts with main now after #80 added `--share-history` to cswap run after you branched. When rebasing, please also add the flag to your run parser and pass `share_history` through the mapped-account path, otherwise `cswap run --share-history` with no account would silently drop the flag.

Also please don't call the switcher's private methods (`_get_sequence_data_migrated`, `_find_account_slot`, `_get_display_tag`) from cli.py. There's a "public accessors for session mode" section in switcher.py for exactly this. A small public method like `slot_for_directory(cwd)` on the switcher would fix that and keep the resolution logic out of cli.py.

Can we also prune mappings when an account is displaced by slot overwrite/import/migration, and not only `--remove-account`. Otherwise stale mappings can remain.

Couple of things that would nice to have: the root guard is copy-pasted into map/unmap now, worth extracting a helper. And `cswap map 2 /typo/path` succeeds silently since the path isn't checked, a warning when it doesn't exist would be nice.

Otherwise good design and looking good separate mappings.json keyed by (email, org) and excluded from export makes sense. Thanks.

realiti4:
Hi, if you'd prefer, I can also push the adjustments myself. Just let me know.

erik-sutton95:
Thanks for the review! Rebased onto main and pushed, all points addressed:

- **`--share-history`**: the `run` parser now carries the flag from #80, and it's forwarded through the mapped-account path too. Added a regression test that `cswap run --share-history` in a mapped directory passes it through.
- **Private methods**: gone from cli.py. Added `slot_for_directory()` and `list_mappings()` to the "public accessors for session mode" section, so the resolution and listing logic lives in switcher.py now.
- **Pruning on displacement**: extracted a `_prune_mappings` helper, called from `remove-account` and both slot-overwrite paths (`add --slot`, `add-token --slot`). I looked at import and migration too and left those alone on purpose: mappings are keyed by `(email, organizationUuid)` and both preserve that identity `--import --force` only overwrites the slot whose identity already matches (a colliding exported slot number allocates a fresh slot instead of displacing), and slot migration moves the same identity to a new slot, so the mapping still resolves. Added a test that a mapping survives a slot migration and resolves to the new slot. If you'd rather prune there anyway for belt-and-suspenders, happy to add it.
- **Root guard**: extracted into a `_guard_root()` helper, shared by `run`/`map`/`unmap`.
- **Path check**: `cswap map 2 /typo/path` now warns "is not an existing directory (mapping it anyway)", kept it a warning rather than an error so you can still map a directory before cloning into it.

Full suite: **1202 passed, 3 skipped**. No need to push anything yourself, but thanks for the offer, let me know if anything else needs adjusting.

realiti4:
Thank you for the adjustments! Looking good, I've merged it.

PR #73, opened 2026-06-25
Add memorable subcommands (cswap list/switch/status) and a clean program name in help

questbibek:
## What & why

`claude-swap` currently exposes everything as flags (`--switch-to`, `--list`, …).
This PR adds **word-style subcommands** as a thin, fully backward-compatible
layer on top, so the common commands are easier to remember and type:

| Subcommand | Existing flag |
|---|---|
| `cswap list` (`ls`) | `--list` |
| `cswap status` (`st`) | `--status` |
| `cswap switch` | `--switch` (rotate) |
| `cswap switch <num\|email>` | `--switch-to` |
| `cswap add` | `--add-account` |
| `cswap add-token [TOKEN\|-]` | `--add-token` |
| `cswap remove` (`rm`) `<num\|email>` | `--remove-account` |
| `cswap export` / `import` / `tui` / `purge` / `upgrade` (`update`) | matching flags |
| `cswap help` | `--help` |

`cswap run …` is unchanged (it was already a subcommand).

## Design — additive and non-breaking

- A leading verb is rewritten to the equivalent flag **before** argparse runs
  (`_translate_subcommand`). The translation only fires when the first token is a
  recognized verb (which never starts with `-`), so the entire existing `--flag`
  interface — and every existing test that drives it — is untouched.
- Tokens after the verb pass through verbatim, so flags still combine:
  `cswap switch --strategy best`, `cswap list --json`, `cswap add --slot 3`.
- Bare `cswap` (no args) still errors via the existing required group.

## Also: clean program name in help

argparse defaulted `prog` to `os.path.basename(sys.argv[0])`, which for an
installed entry-point shim renders as an ugly absolute path (e.g.
`python.exe C:\Users\me\.local\bin\cswap`) on every usage/example line.
Added `_prog_name()` to show the bare command (`cswap` / `claude-swap`).

## Tests

Adds `TestSubcommandAliases` covering the translation table, the `switch`
bare-vs-target special case, alias dispatch, and that `run`/`help` still work.
Existing tests are unchanged and pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,13 +22,13 @@ pipx install claude-swap
 git clone https://github.com/realiti4/claude-swap.git
 cd claude-swap
 uv sync
-uv run cswap --help
+uv run cswap help
 ```
 
 ### Updating
 
 ```bash
-cswap --upgrade        # uv/pipx installs on macOS/Linux: auto-detects and upgrades
+cswap upgrade          # uv/pipx installs on macOS/Linux: auto-detects and upgrades
 # or run your installer directly:
 uv tool upgrade claude-swap
 pipx upgrade claude-swap
@@ -41,7 +41,7 @@ pipx upgrade claude-swap
 Log into Claude Code with your first account, then:
 
 ```bash
-cswap --add-account
+cswap add
 ```
 
 ### Add more accounts
@@ -49,7 +49,7 @@ cswap --add-account
 Log in with another account, then:
 
 ```bash
-cswap --add-account
+cswap add
 ```
 
 ### Switch accounts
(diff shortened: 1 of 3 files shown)


questbibek:
@realiti4 Let me know if you want any changes.

realiti4:
Hi, thanks a lot for the PR! I think we really need this. Bare subcommands for actions and `--flags` for modifiers is the more modern shape, and I also like that it folds `--switch` and `--switch-to` into one command.

- Can you rebase against the current `main` please? `cswap auto` landed since you branched, the new help text should include it too.

- In the README, let's replace every action command with the new form. I think there's no need for a separate section; we can adopt these from now on. One legacy line like "The original flag spellings (`cswap --switch`, `cswap --list`, …) keep working." at the end of the `Usage` section is enough.

- We could follow the same idea in `cswap help`: the `epilog` can drop the per-command `(--flag)` parentheticals and carry that same single legacy line instead.

- Let's drop the `st` alias. `ls`, `rm`, and `update` are conventional, but `st` doesn't feel that widely shared.

questbibek:
Sure

questbibek:
Thanks for the review! All four points are in — rebased and force-pushed:

- **Rebased onto current main** — `cswap auto` is now included in the new help (command list plus an `auto --once` cron example in the flags section).
- **README** — every action command now uses the subcommand form throughout; the separate subcommands section/table is gone, replaced by a single line at the end of Usage: "The original flag spellings (`cswap --switch`, `cswap --list`, ...) keep working."
- **Help epilog** — dropped the per-command `(--flag)` parentheticals and carries that same single legacy line instead.
- **Dropped `st`** — aliases are now just `ls=list rm=remove update=upgrade`.

The rebase also picked up main's newer test conventions (`create=True` on the `os.geteuid` patches for Windows, and the new `force=`/`share_history=` kwargs). Full suite passes locally: 687 passed, 28 skipped.

realiti4:
Thanks a lot for the adjustments, looking good!

Just a heads-up I'll hold the merge until I'm ready to cut the 0.17.0 release, so the README on main doesn't show commands the released version doesn't support yet.

questbibek:
Sounds good, thanks! Holding for 0.17.0 makes sense — happy to rebase again if anything else lands in the meantime.

PR #75, opened 2026-06-25
Render TUI menu actions in-place and add a live Watch dashboard

Steamvis:
## What this does

The interactive TUI (`cswap --tui`) had one rough edge: every action dropped out of the curses menu, printed to stdout, and waited for Enter. After a couple of actions the screen was a wall of scrollback and it was hard to follow what had happened.

This PR keeps everything on screen.

### In-place rendering
**List**, **Status**, **Switch** and **Remove** now render inside a scrollable panel *within* the menu instead of dumping to stdout. The action's output is captured, the existing printer colors are mapped onto curses attributes (so it looks the same), and the result is shown in a small pager — arrows / PgUp / PgDn to scroll, `q` to go back. **Add** and **Refresh** still shell out, since they can need the real terminal for login.

### Watch dashboard
A new **Watch** menu item opens a live, read-only dashboard of the active account and per-account usage that auto-refreshes on a timer. `+`/`-` change the interval, `r` refreshes now, `q` exits. Usage is cache-gated at 15s, so very short intervals re-render cached numbers rather than hammering the API.

### Bonus
`remove_account` gained an `assume_yes` flag (default off, so the CLI is unchanged). The TUI already confirms before removing, so it no longer double-prompts.

## Screenshots

Main menu (with the new Watch entry):

menu

`List accounts` rendered in-place — a scrollable panel, no more stdout dump:

list

The Watch dashboard:

watch

## Notes

- **No new dependencies** — rendering uses the stdlib `curses` module.
- Graceful fallback to bold/dim/plain on no-color or very small terminals.
- Tests added for the ANSI tokenizer, color mapping, the pager, output capture, the watch loop, and `assume_yes`. Full suite: **626 passed, 3 skipped**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ cswap --upgrade                 # Upgrade claude-swap to the latest version
 cswap --purge                   # Remove all claude-swap data
 ```
 
+In the interactive TUI, read-only and switch/remove actions now render in a scrollable panel **inside** the menu (no more scrolling stdout). The **Watch** item opens a live dashboard of the active account and per-account usage that auto-refreshes on a configurable interval (`+`/`-` to change it, `r` to refresh now, `q` to exit). Usage itself is cached for 15s, so very short intervals re-render cached numbers rather than re-fetching.
+
 ## Tips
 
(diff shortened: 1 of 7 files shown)


realiti4:
Hi, thanks for the PR. It looks good, I've merged it. I was also thinking about updating the TUI.

PR #76, opened 2026-06-28
feat: auto-switch accounts by usage threshold (cswap autoswitch)

questbibek:
## What

Adds `cswap autoswitch` — a background watcher that polls each account's quota and switches to the account with the most headroom **before** the active one hits its limit, so you keep working past a single account's rate limit without switching by hand.

```bash
cswap autoswitch          # interactive setup wizard
cswap autoswitch start    # run the watcher in the background
cswap autoswitch status   # rules + per-account usage vs threshold
cswap autoswitch check    # evaluate once now and switch if over threshold
cswap autoswitch stop     # stop the watcher
```

## How

- **New `autoswitch` module.** Persisted rule set in `autoswitch.json` — a common/global usage-% threshold plus optional **per-account overrides**.
- **Reuses the existing engine.** The actual switch is delegated to `switch(strategy="best")`; autoswitch only adds the *gate* (don't switch until the active account's usage — the higher of its 5h/7d window — reaches its threshold). It therefore never moves onto a worse account, and with one/zero accounts it stays put and logs why.
- **Cross-platform detached watcher.** `start`/`stop`/`status` via a PID file, reusing `process_detection.is_pid_alive`; POSIX `start_new_session`, Windows `DETACHED_PROCESS`. Polls on an interval, logs each decision to `autoswitch.log`, survives transient/network errors.
- **CLI.** `autoswitch` (alias `auto`) is pre-dispatched like `run`, with sub-verbs `start`/`stop`/`status`/`check` and a bare wizard.

A swap rewrites the credentials file Claude Code reads, so it takes full effect on Claude's **next** start (a running session keeps the credentials it loaded) — documented in the README.

## Tests

`tests/test_autoswitch.py` (19 tests): config round-trip + defaults, `effective_threshold` (override beats global), `evaluate_and_switch` for over/under-threshold, per-account override, binding-window = max(5h, 7d), usage-unavailable, no-active-account, single-account stays put; watcher start/stop/refuse-double-start/stale-PID cleanup; and the wizard via scripted input.

## Note

I'm aware there are already a few auto-switch proposals open (#69, #65, #50, #57). This one is intentionally minimal and CLI-first: it doesn't add a daemon framework or UI, it just gates the existing `--strategy best` logic behind a configurable threshold and runs it on a timer. Happy to align with whichever direction the maintainer prefers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,27 @@ Or let claude-swap auto-pick by remaining quota — `cswap --switch --strategy b
 
 **Note:** You usually don't need to restart — on Linux/Windows the new account is picked up automatically, and on macOS after the Keychain cache expires. To apply it instantly, restart Claude Code or reopen the VS Code extension tab. See [Tips](#tips) for the per-platform details.
 
+### Auto-switch by usage
+
+Instead of switching by hand, let claude-swap watch your quota and hop to the freshest account *before* the active one hits its limit. Run the setup wizard:
+
+```bash
+cswap autoswitch
+```
+
+The wizard is a small menu — set a **common threshold** that applies to every account (e.g. switch at 85% usage), optionally **override individual accounts** (e.g. keep your work account until 95%), set the poll interval, and start/stop the background watcher. Each change is saved as you go.
+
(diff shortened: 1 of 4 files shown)


realiti4:
Hi, thanks a lot for the PR! I'd been wanting to build this one myself, and it's now merged as #81 which also takes claude code `.claude.lock` into account for safety. Your PR was really good as well also delegating actual switch to `switch(strategy="best")`

Since that is merged, I'll close this in favor of that. But I'd happily accept new PRs for new ideas on top of the new `AutoSwitchEngine`. Thanks again!

PR #77, opened 2026-06-28
fix(macOS): show "keychain unavailable" instead of "no credentials" when the active Keychain is locked, and retry the read

ynaamane:
On macOS, `cswap --list` / `--status` shows the active account as **"no credentials"** when the login Keychain is merely *transiently unreadable* — e.g. locked just after wake, or contended with Claude Code's own statusline polling the same `Claude Code-credentials` item. On a Keychain-only setup (no `~/.claude/.credentials.json` fallback present), the single failed `security` read collapses to `""`, which is indistinguishable from a genuinely empty slot — so the user is nudged into an unnecessary re-login even though the credential is intact.

## What this does

1. **Retry the active OAuth Keychain read once** (bounded, short backoff) to ride out a transient lock/contention before falling back. A genuinely absent item (rc-44, returned as `None` without raising) is **not** retried.
2. **Classify the read** via a new `ActiveCredentials(value, keychain_unavailable)` result. When the OAuth Keychain read fails *and nothing else covers it* (no plaintext file, no managed API key), the active usage row now renders:

   ```
   keychain unavailable — locked or in use; try again
   ```

   instead of the misleading `no credentials`. JSON `--list`/`--status` get a distinct `usageStatus: "keychain_unavailable"` (sibling of the existing `token_expired` / `api_key` / `no_credentials`).

A genuinely absent credential still reads as `no credentials`, so there is no behavior change for the empty-slot case.

## Notes

- `_read_credentials` keeps its historic `str | None` contract; the switch paths are unchanged. The classification lives in a new `_read_active_credentials()` that `_read_credentials` delegates to.
- The retry is an I/O backoff between attempts of an external CLI (`security`), not a sleep papering over an internal race; a hard lock simply surfaces the new "keychain unavailable" line.

## Tests

- +5 tests: retry-then-succeed, keychain-unavailable-with-no-fallback, failure-covered-by-file, rc-44-absent-is-not-unavailable, and an end-to-end `--list` rendering check; plus a `usage_fields` mapping assertion.
- Full suite: **631 passed, 3 skipped**.

### DIFF
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -23,8 +23,9 @@ import logging
 import os
 import sys
 import tempfile
+import time
 from pathlib import Path
-from typing import Protocol
+from typing import NamedTuple, Protocol
 
 from claude_swap import macos_keychain
 from claude_swap.exceptions import CredentialWriteError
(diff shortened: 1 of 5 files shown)


realiti4:
Hi, I've experienced the same thing. Thank you for the fix, it is looking good. I've merged it

PR #80, opened 2026-07-02
feat: add --share-history flag to cswap run for unified conversation history

### LINKED ISSUE #74: [Feature request] CLI flag for `cswap run` to allow 2 account share the same conversation history
Currently, `cswap run 1` and `cswap run 2` don't share same conversation history. I'm wondering if this is techical limitation?
If not, can we have it supported?

maddada:
# PR description written by Human:

I wanted to be able to share the sessions that I had between my work and personal accounts but this wasn't possible before. Worked with Fable to implement this as a flag. Hope you don't mind me making this PR.

Closes #74

---

# PR description written by Fable high:

## Summary

Adds an opt-in `--share-history` flag to `cswap run` that shares conversation
history (`projects/` and `history.jsonl`) from `~/.claude` into the session
profile, so every account sees — and `claude --resume` lists — the same
conversations.

```bash
cswap run 2 --share-history     # one conversation history across all accounts
cswap run 2 --no-share-history  # back to per-account history (the default)
```

## Motivation

Per-account history is a deliberate default and stays that way. But when you
use multiple accounts as *one person* (e.g. switching accounts to spread usage
limits), per-account history splits your conversations across profiles:
`--resume` under account 2 can't see the session you started under account 1.
This flag lets those users opt into a single unified history while keeping
customization sharing (`--no-share`) an independent choice.

## What it does

- **New flag** on `cswap run`: `--share-history` /
  `--no-share-history` (default: off). Independent of `--no-share`, so
  `--no-share --share-history` gives a bare profile with unified history.
- **Sharing mechanism**: the same symlink machinery as `SHARED_ITEMS`, extended
  with a new `HISTORY_ITEMS` tuple (`projects/`, `history.jsonl`). Links are
  recorded in the existing `.cswap-shared.json` manifest, so removal only ever
  touches cswap-created links, never user data.
- **No data loss on first use**: if the profile already accumulated its own
  history before the flag existed, it is merged into `~/.claude` first —
  transcript directories merge file-by-file (filenames are UUIDs, so a
  collision means the identical session; first writer wins), and
  `history.jsonl` merges by appending lines not already present. Nothing
  disappears from `--resume`.
- **Safe with live sessions**: merging moves files out from under any `claude`
  still running in the profile, so if the profile has live sessions the merge
  is deferred with a notice and retried on the next launch.
- **Toggling off is clean**: `--no-share-history` removes the links (data stays
  in `~/.claude`) and the profile goes back to accumulating its own history.
- **Fresh installs work**: if `~/.claude` has no history yet, an empty share
  target is seeded so there is something to link.

## Windows

Not supported in this PR, and rejected explicitly with a clear error. On
Windows, sharing uses re-synced copies instead of symlinks — a copied history
would *fork* on first write rather than stay shared, which is worse than not
having the feature. `_sync_sharing` additionally drops history links
defensively in copy mode, so a profile moved from POSIX to Windows self-heals.

**Windows support is feasible as a follow-up PR** — nothing here blocks it. This PR focuses on macOS/Linux because that is what we can actually test, and the failure mode of getting Windows wrong is silent history forking — exactly what the feature exists to prevent.

A sketch for whoever picks it up:

- `projects/` (the transcripts) can be shared with an NTFS **directory
  junction** (`_winapi.CreateJunction` / `mklink /J`) — works on every
  Windows install with no admin rights or Developer Mode. The link-detection
  code (`dest.is_symlink()`, `_remove_managed`) needs to become
  junction-aware (`Path.is_junction()` on Python 3.12+, `st_reparse_tag`
  before that).
- `history.jsonl` needs a real **file symlink**, which requires Developer
  Mode (or elevation). A capability probe at launch can pick between linking
  it, or sharing `projects/` alone with a notice — unified transcripts are
  the point; prompt history is a nicety.
- **Hardlinks should not be used**: if Claude Code ever rewrites
  `history.jsonl` via temp-file-and-rename, a hardlink silently forks.
- The merge logic (`_merge_history_into_source`) is already
  platform-neutral and needs no changes.

The main cost is verification on real Windows, not the code itself.

## Design notes

- `projects/` and `history.jsonl` were deliberately excluded from
  `SHARED_ITEMS` ("per-account history is a feature") — this PR keeps that
  default and moves them behind an explicit opt-in rather than changing the
  exclusion.
- The merge raises `OSError` on any failure and leaves remaining files in
  place, so a partial merge is retried (not lost) on the next launch.
- Manifest reads still filter to names cswap could have created, now
  `SHARED_ITEMS + HISTORY_ITEMS` — a corrupted or hand-edited manifest can
  never cause removal of arbitrary entries.
- Concurrency model is unchanged: lock-free on the reuse path, last-writer-wins
  between concurrent runs with different flags, self-healing on next launch.

## Testing

- `643 passed, 3 skipped` (full suite, POSIX).
- New tests:
  - CLI: flag parsing for `--share-history` / `--no-share-history`, defaults,
    and that args after `--` are still forwarded verbatim.
  - Session (POSIX): not shared by default; links created and recorded in the
    manifest; missing source seeded; existing profile history merged (both
    `projects/` and `history.jsonl`, including dedup and collision handling);
    merge deferred while the profile has live sessions; toggling off removes
    links but keeps data; independence from `--no-share`.
  - Windows: `run` rejects the flag with a clear error; `_sync_sharing` never
    links history in copy mode.

## Docs

README updated: new example line plus a paragraph explaining the behavior,
the merge-on-first-use guarantee, `--no-share-history`, and the Windows
limitation. Module docstring and CLI `--help` cover the same ground.

https://claude.ai/code/session_012JvGuwHK7mdPbxE9bHTs95

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,12 @@ cswap run 2                     # launch Claude Code as account 2, here only
 cswap run user@example.com      # by email
 cswap run 2 -- --resume         # everything after '--' is forwarded to claude
 cswap run 2 --no-share          # don't share your ~/.claude customizations
+cswap run 2 --share-history     # one conversation history across all accounts
 ```
 
-Your `~/.claude` customizations (settings, keybindings, CLAUDE.md, skills, commands, agents) are shared into the session by default — use `--no-share` for a bare profile. Conversation history stays per-account.
+Your `~/.claude` customizations (settings, keybindings, CLAUDE.md, skills, commands, agents) are shared into the session by default — use `--no-share` for a bare profile.
+
+Conversation history stays per-account by default. Pass `--share-history` to share `projects/` and `history.jsonl` from `~/.claude` instead, so every account sees (and `--resume` lists) the same conversations. Any history the profile already accumulated is merged into `~/.claude` on first use — nothing disappears. `--no-share-history` returns the profile to per-account history. Not supported on Windows yet (sharing uses copies there, which would fork the history rather than share it).
 
 ### Refresh expired tokens
 
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -47,6 +47,7 @@ Examples:
   cswap run 2
   cswap run user@example.com
   cswap run 2 --no-share
+  cswap run 2 --share-history
   cswap run 2 -- --resume
         """,
     )
@@ -62,6 +63,18 @@ Examples:
             "Don't share settings/keybindings/CLAUDE.md/skills/commands/agents "
             "from ~/.claude into the session profile (and remove previously "
             "shared items)"
+        ),
+    )
+    parser.add_argument(
+        "--share-history",
+        action=argparse.BooleanOptionalAction,
+        default=False,
+        help=(
+            "Share conversation history (projects/ and history.jsonl) from "
+            "~/.claude into the session profile, so every account sees one "
+            "unified history. History the profile already accumulated is "
+            "merged into ~/.claude first. --no-share-history restores "
+            "per-account history (the default). Not supported on Windows."
         ),
     )
     parser.add_argument(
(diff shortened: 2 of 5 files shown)


realiti4:
Hi, thanks for the PR looking good. I've merged it.

Issue #82, opened 2026-07-03
Show usage details by fable use


ljmerza:
One of the main reasons I switch is for fable usage limits but I can't see this when using the --list flag. Is it possible to add fable usage metrics? Thanks.

ljmerza:
Opened #83 for this.

The data turns out to already be in the `/api/oauth/usage` response — the newer `limits` array carries `weekly_scoped` entries with `scope.model.display_name` (e.g. `"Fable"`), but `build_usage_result()` only kept `five_hour`/`seven_day`/`extra_usage`, so the per-model windows were fetched and then discarded.

The PR parses those into per-model lines in `--list` (and the `--json` projection), with a `(!)` when a model is at/over its limit:

```
└ Fable: 100%   resets Jul 5 08:59   in 1d 19h  (!)
```

Additive and backward-compatible — responses without a `limits` array yield no change, and existing 5h/7d/spend output is untouched. Tests included.

realiti4:
Hi, I think it's a good addition. However, I'm not sure what will happen after July 7th, when Fable access ends for subscription plans. Let's wait until the 7th and see how they handle this field afterward. When they switch to credit based, they might remove that. If it's only temporary, there's no need to add it to the codebase for just 4 days.

ljmerza:
this isnt just for fable, its any extra models provided from the api. if fable isnt there then it doesnt show

realiti4:
Sorry, I didn't look at it in detail before. Well thought out and implemented, I've merged it. Thanks again!

PR #83, opened 2026-07-03
feat: show per-model weekly usage (e.g. Fable) in --list (#82)

ljmerza:
## What

Adds per-model weekly usage windows (e.g. **Fable**) to `cswap --list`, closing #82.

## Why

The `/api/oauth/usage` endpoint now returns a `limits` array whose `weekly_scoped` entries carry `scope.model.display_name` (e.g. `"Fable"`), alongside the legacy `five_hour`/`seven_day` keys. `build_usage_result()` only kept `five_hour`/`seven_day`/`extra_usage`, so the per-model data was fetched and then discarded — never shown. Switching accounts to dodge a per-model (Fable) limit is exactly the use case #82 describes, but the number that drives the switch wasn't visible.

## What changed

- **`oauth.build_usage_result()`** — parse `weekly_scoped` entries from `limits[]` into `result["scoped"]` (a list of `{name, pct, resets_at, countdown, clock}`).
- **`switcher._format_usage_lines()`** — render one line per model; a `(!)` flags a model at/over its limit.
- **`json_output.usage_to_json()`** — project `scoped` into the `--json` output (camelCase, carrying `name`).

## Compatibility

Additive and backward-compatible: responses without a `limits` array yield no `scoped` key, and existing `5h`/`7d`/`spend` output is unchanged.

## Example (verified against live accounts; emails/names redacted)

`cswap --list`:

```
Accounts:
  1: user1@example.com [Personal]
     ├ 5h:   0%
     ├ 7d:  62%   resets Jul 5 08:59   in 1d 19h
     └ Fable: 100%   resets Jul 5 08:59   in 1d 19h  (!)

  2: user2@example.com [Work] (active)
     ├ 5h:  10%   resets 16:39         in 3h 28m
     ├ 7d:  72%   resets 17:59         in 4h 48m
     └ Fable: 100%   resets 17:59         in 4h 48m  (!)
```

`cswap --list --json` (per account):

```json
"scoped": [
  { "pct": 100.0, "resetsAt": "…", "countdown": "1d 19h", "clock": "Jul 5 08:59", "name": "Fable" }
]
```

## Not included (happy to add if wanted)

- `account_headroom()` (used by `cswap auto`) still considers only `five_hour`/`seven_day`, so auto-rotate won't trigger on a Fable-only limit. Left out to keep this PR display-only; scoped windows could be wired into headroom in a follow-up.
- `severity` / `is_active` from `limits[]` aren't surfaced, to keep the diff minimal.

## Tests

`757 passed, 3 skipped`. New coverage: `build_usage_result` scoped extraction + a no-`limits` negative (`test_oauth.py`), `_format_usage_lines` per-model rendering incl. the marker (`test_switcher.py`), and the `--json` scoped projection (`test_json_output.py`).

### DIFF
--- a/src/claude_swap/json_output.py
+++ b/src/claude_swap/json_output.py
@@ -37,6 +37,13 @@ def _window_to_json(entry: dict) -> dict:
     return out
 
 
+def _scoped_window_to_json(entry: dict) -> dict:
+    """Project a per-model scoped weekly window, carrying its model name."""
+    out = _window_to_json(entry)
+    out["name"] = entry["name"]
+    return out
+
+
 def usage_to_json(usage: dict) -> dict:
     """Convert the internal usage dict to its camelCase JSON projection.
 
@@ -63,6 +70,8 @@ def usage_to_json(usage: dict) -> dict:
         if "clock" in spend:
             spend_out["clock"] = spend["clock"]
         out["spend"] = spend_out
+    if "scoped" in usage:
+        out["scoped"] = [_scoped_window_to_json(w) for w in usage["scoped"]]
     return out
 
 
--- a/src/claude_swap/oauth.py
+++ b/src/claude_swap/oauth.py
@@ -238,6 +238,31 @@ def build_usage_result(data: dict) -> dict | None:
             except (TypeError, ValueError) as e:
                 _logger.debug("extra_usage parse failed: %r", e)
 
+    # Per-model weekly limits live in the newer ``limits`` array as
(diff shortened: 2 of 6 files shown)


realiti4:
Hi, thanks for the PR! I think it's a good addition. However, I'm not sure what will happen after July 7th, when Fable access ends for subscription plans. Let's wait until the 7th and see how they handle this field afterward. If it's only temporary, there's no need to add it to the codebase for just 4 days.

PR #89, opened 2026-07-04
Quick-switch accounts from the Watch view

7GMA:
Adds `[s]` to the Watch view so you can switch accounts without leaving it.

- `s` → select mode: ↑/↓ highlight an account row, **Enter** switches and stays in Watch, **Esc/q/s** cancels.
- Auto-refresh pauses while selecting so rows don't shift under the cursor; a forced refresh after the switch shows the new active account.
- Reuses `switcher.switch_to`; no account logic added. Selection is derived from the existing `list_accounts()` output (a small `_watch_account_rows` helper) and highlighted via an optional `extra` attr on `_addstr_ansi`.
- Tests added; full suite green.

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -1790,6 +1790,10 @@ class ClaudeAccountSwitcher:
         print(bolded("Accounts:"))
         for i, (num, email, org_name, org_uuid, is_active, _) in enumerate(accounts_info):
             tag = self._get_display_tag(email, org_name, org_uuid)
+            # NOTE: the TUI watch view (tui._watch_account_rows) parses this
+            # output to map rows to accounts for quick-switch: it relies on the
+            # uncolored ``  {num}: `` prefix and the ``(active)`` marker below.
+            # Keep them intact when tweaking this line, or update that parser.
             if is_active:
                 marker = f" {bold_accent('(active)')}"
                 print(f"  {num}: {email} {muted(f'[{tag}]')}{marker}")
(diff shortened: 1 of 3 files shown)


realiti4:
Hi, thanks for the PR! I think quick switching from the `watch` view would be nice.

The main one: failed switches are currently invisible. `_capture(...)` swallows the output, so if `switch_to` errors out, for example due to a lock timeout, keychain issue, or an account being removed meanwhile, the user presses Enter and nothing visibly happens. Could we surface the captured output when it contains an error, e.g. flash it in the meta line, or fall back to the pager like `_do_switch` does?

Also, since `_watch_account_rows` parses the human `list_accounts()` output, let's add a short comment next to the `print(f" {num}: ...")` lines in `switcher.py` noting that the `watch` view depends on that format. Otherwise, a cosmetic tweak there could silently break this.

A small nice-to-have would also be starting the selection on the currently active row instead of `row 0`.

7GMA:
Thanks for the review! Addressed all three in the latest push:

- **Failed switches surfaced** — `switch_to` errors (lock timeout, keychain, account removed) now fall back to the pager like `_do_switch`; successful switches still stay silently in Watch.
- **Starts on the active row** — selection begins on the `(active)` account instead of row 0.
- **Format note added** — comment next to the `list_accounts()` print lines flagging that the watch parser depends on the `  {num}: ` prefix + `(active)` marker.

Added tests for each; full suite green.

realiti4:
Thanks for the adjustments, looking good merged it.

PR #92, opened 2026-07-05
Fix stale reset countdown in list/status: recompute from resets_at at render time

jackfuchs:
## Bug

`cswap list` can print a reset countdown that contradicts the reset clock on the same line:

```
7d:     62%   resets Jul 6 15:59   in 17h 0m   (!) · 2h ago
```

At the time of that output the reset was ~15h away, not 17h. The clock (`Jul 6 15:59`) was correct; the countdown was frozen.

## Root cause

`build_usage_result()` computes the `countdown` and `clock` display strings once at **fetch** time and stores them with the measurement. When the usage store serves last-good data later (stale-served entries, e.g. inactive accounts), `_format_usage_lines()` prints those frozen strings verbatim — so the countdown is off by exactly the measurement's age. The `clock` string has the same issue in a subtler form: a same-day `resets 15:30` silently starts meaning *yesterday* 15:30 once the date rolls over, and a reset that already passed still renders as pending.

The TUI already solved exactly this (`tui/data.py::reset_text`, "the countdown the API sent was correct at *fetch* time and drifts as the measurement ages") — the CLI list/status path just never got the same treatment.

## Fix

`_format_usage_lines()` now derives both strings from the stored `resets_at` at render time via the existing `format_reset()`, for the 5h/7d windows, per-model scoped windows, and the spend row. Entries persisted without `resets_at` (older versions) fall back to the fetch-time strings, so old cache files keep rendering.

Before/after with a measurement fetched 2h ago (reset 15h away):

```
before: 7d:  62%   resets 18:31  in 17h 0m
after:  7d:  62%   resets 16:31  in 14h 59m
```

## Tests

- countdown/clock recomputed from `resets_at` (stale cached strings ignored)
- fallback to cached strings when `resets_at` is absent (old persisted entries)
- fallback on unparseable `resets_at`
- spend row clock recomputed

Full suite: 891 passed, 3 skipped.

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -100,6 +100,27 @@ _FETCH_STAGGER_S = 0.25
 _USAGE_AGE_NOTE_S = 90.0
 
 
+def _reset_cell(window: dict) -> tuple[str, str] | None:
+    """``(clock, countdown)`` for one usage window, or None when unknown.
+
+    Recomputed from ``resets_at`` at render time: the strings cached at fetch
+    time drift as the measurement ages (a countdown frozen 2h ago overstates
+    the remaining wait by those 2h, and a same-day "15:30" clock silently
+    starts meaning yesterday). Entries persisted without ``resets_at`` fall
+    back to the fetch-time strings — stale beats blank.
+    """
+    resets_at = window.get("resets_at")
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for the fix, merged it!

PR #93, opened 2026-07-06
feat(autoswitch): add consume-first strategy

deathemperor:
Follow-up to #65 (as discussed there): brings the menu bar's **consume-first** idea into the core auto-switch engine as a selectable `--strategy`, so it works from the CLI/TUI too — not buried in the GUI.

## What it does
A second target-selection strategy alongside the default `best` (most headroom). `consume-first` proactively keeps you on the account whose **weekly (7d) window resets soonest** — use-it-or-lose-it — so perishable quota isn't wasted.

- **Below the threshold:** switches to a candidate whose weekly window resets *strictly sooner* than the active account's (with room to spare, past the hysteresis bar, respecting the cooldown). Once you're on the soonest-resetting account it stays put (`already-consuming-soonest`).
- **At/over the threshold or on failover:** must move, and picks the soonest-resetting candidate *with headroom* instead of the most-headroom one.

The engine now honors `settings.strategy` (previously `"best"` was effectively hardcoded — the field existed but wasn't read).

## Interface
- `cswap auto --strategy consume-first`
- `cswap config set autoswitch.strategy consume-first`

## Tests
- Engine: below-threshold switch to soonest reset, stay-when-already-soonest, at-limit prefers soonest over max-headroom, skips exhausted-but-sooner accounts, cooldown suppression, and a `best`-strategy regression guard.
- Settings: `consume-first` is a valid strategy (load + `set`), `merged_with_cli` override.

Full suite green (896 passed).

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ cswap auto --threshold 80      # switch earlier
 cswap auto --model Fable       # also switch when the Fable weekly limit is hit
 cswap auto --once              # single check-and-switch, for cron/scripts
 cswap auto --dry-run           # log what it would do, never switch
+cswap auto --strategy consume-first   # burn the soonest-resetting account first
 ```
 
 <details>
@@ -95,6 +96,7 @@ cswap auto --dry-run           # log what it would do, never switch
 
 - Runs safely alongside Claude Code: switches take the same credential locks Claude Code uses, so a swap never collides with a token refresh.
 - A cooldown (default 5 min) and a hysteresis margin stop it flip-flopping near the threshold: a proactive switch only lands on an account that's below the threshold *and* better than the current one by the margin — a candidate that clears the margin is always taken, but two accounts hovering at the line never ping-pong. When every account is exhausted it sleeps until the first one becomes usable again.
+- **Strategies** (`--strategy`, or `cswap config set autoswitch.strategy`): `best` (default) stays put until the active account nears its limit, then moves to the account with the most quota left. `consume-first` proactively keeps you on the account whose **weekly window resets soonest** — use-it-or-lose-it — switching to a sooner-resetting account (with room to spare) even below the threshold, so perishable weekly quota isn't wasted.
 - Usage polling is adaptive — a couple of accounts per check, busy alternates watched more closely, exhausted ones left alone until they reset — so API traffic stays flat no matter how many accounts you manage.
 - It fails safe: if a usage check errors it keeps trusting the last-known numbers while retries back off, and an expired token on an idle machine makes it hold rather than fail over (Claude Code refreshes the token on your next message).
 - An account whose refresh token has died is quarantined and reported until you log in with it and re-run `cswap add --slot N`. API-key accounts are never rotated onto unless you pass `--include-api-key-accounts`.
--- a/src/claude_swap/autoswitch.py
+++ b/src/claude_swap/autoswitch.py
@@ -369,6 +369,20 @@ _earliest_future_reset_ts = poll_policy.earliest_future_reset_ts
 _parse_reset_ts = poll_policy.parse_reset_ts
 
 
+def _seven_day_reset_ts(usage: dict | str | None) -> float | None:
+    """Epoch of an account's 7-day (weekly) window reset, or None if unknown.
+
+    The consume-first strategy ranks by this — the weekly window is the
+    perishable quota (the 5-hour one recycles too fast to be worth planning
+    around).
+    """
+    if isinstance(usage, dict):
+        window = usage.get("seven_day")
+        if isinstance(window, dict):
+            return _parse_reset_ts(window.get("resets_at"))
+    return None
+
+
 def _ref(number: str, email: str) -> dict:
     return {"number": int(number), "email": email}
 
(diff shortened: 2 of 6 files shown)


kmizzi:
Really like this — it's the right feature and the engine wiring is clean (ranking by `_seven_day_reset_ts`, the strictly-sooner guard on proactive moves, and the `already-consuming-soonest` hold instead of treating "nothing sooner with room" as a block). Two thoughts on making "consume-first" actually **consume it all**, which is where the naive soonest-first rule leaves quota on the table.

### The gap: soonest ≠ most-salvageable

The ranking key is `(reset_ts, -h)` — reset order first, headroom only as a tie-break — and nothing checks whether the soonest account's remaining weekly quota is *reachable before its reset*. The weekly window can only be drained **through** the 5-hour window: every ~5h you can push 7d utilization up until the 5h window caps, then you're throttled until it refills. So the realistically-consumable perishable quota is bounded by `windows_left × (5h throughput)`, not by the raw 7d headroom.

That produces two misfires against the "burn it before it resets" goal:

1. **Stranded tail.** An account that resets soonest but has more 7d quota left than the remaining 5h windows can deliver parks you there to "save" quota that's structurally unreachable — you lose the same quota you switched to protect.
2. **Burn rate ignored.** An account resetting a bit *later* but fully drainable in the time left can realize more perishable quota than the soonest-but-stranded one. Today it's always ranked behind.

(The instantaneous case is already handled — since `account_headroom` is `100 − max(5h, 7d, scoped)`, an account whose 5h window is maxed *right now* has `h ≤ 0` and is correctly skipped. This is about the **rate** over the remaining time, not the current block.)

### A burn-rate-aware refinement

The usage API gives 5h/7d as percentages of their own caps, not absolute tokens, so you can't statically convert "one 5h window" into "X% of weekly." But you don't need to — the engine already polls every ~60s, so the **measured** `d(weekly_pct)/dt` while an account is active *is* the burn rate. Concretely:

1. Persist the last `(sampled_at, seven_day_pct)` per account (the poller already has both).
2. `burn_rate = Δ7d_pct / Δt` from the two most recent *active* samples (fall back to a conservative prior until there are two).
3. `reachable = min(100 − seven_day_pct, burn_rate × hours_to_reset)` — how much perishable quota you can actually still consume there.
4. Rank consume-first by **most `reachable`**, tie-broken by soonest reset (inverting today's key). Soonest-resetting-*and*-drainable wins; soonest-but-stranded correctly loses to a slightly-later account you can finish.
5. **Time hysteresis:** if `hours_to_reset` is under ~one 5h window, a proactive switch spends a cooldown for near-zero realizable burn — hold instead. This is the "don't blindly chase the nearest reset" guard.

Nice property: with no burn-rate data yet, `reachable` collapses to `100 − seven_day_pct` and the ranking degrades gracefully to your current soonest-first behavior — so it can fold straight into `consume-first` rather than needing a separate strategy name (though a `consume-first-adaptive` choice is also fine if you'd rather keep this one simple).

### Offer

Happy to push this as a follow-up commit on `feat/consume-first-strategy` (or a stacked PR on top of yours), keeping your selection logic as the baseline and layering the burn-rate projection + time-hysteresis on top, with engine tests for the stranded-tail and later-but-drainable cases. Only if you and @realiti4 are up for it — your call. Heads up too that the branch currently shows as conflicting against `main` and'll need a rebase.

Vesely:
+1. I have two accounts with weekly windows resetting Wednesday and Saturday. Under `best` I regularly end up on the freshly-reset account while the other still has quota expiring in two days, and that quota is lost. This fixes exactly that.

The burn-rate idea above looks sensible, but with two accounts there's only one alternative to rank, so the stranded-tail case doesn't really apply here. Hope this can go in as-is and the adaptive version follows later.

realiti4:
Merged — thanks @deathemperor 🎉 Sorry for the wait.

Post-merge, I pushed two hardening commits to `main` for opt-in edge cases the review turned up:

* 019a6a5 extends the under-lock cooldown recheck to consume-first so two concurrent engines can't double-switch; adds distinct hold reasons, including a new `reset-unknown` when the active account's weekly reset isn't reported; and brings exit-code parity with best for zero-candidate ticks.
* 2fe9a50 adds the same parity when the only peers are API-key accounts.

Nothing behavioral changes for best.

One more fix is in progress: below-threshold selection currently runs on stored candidate data (up to ~10 minutes old), so before an actual consume-first switch fires, we'll refetch candidates and re-verify the choice, and treat already-past `resets_at` values as unknown. That keeps the adaptive-polling traffic flat while making the decision data trustworthy.

@kmizzi — yes, would love that. The stranded-tail observation is real, and the burn-rate projection is the right refinement. On the naming question: agreed with your first instinct — fold it into `consume-first` with the graceful degradation you described; one strategy that gets smarter beats a menu. Now that this is merged, main makes a clean base, and the freshness fix above should be useful groundwork since burn-rate estimation reads the same stored samples.

@Vesely — the merged behavior covers your two-account case as-is; the adaptive version will only sharpen target choice when there are multiple alternatives.

kmizzi:
Thanks @realiti4 — glad this is in, and glad the burn-rate refinement is welcome. I'll pick it up as the adaptive step: fold it into `consume-first` (ranking by *reachable* perishable quota, degrading gracefully to today's soonest-first when there's no rate data yet), with engine tests for the stranded-tail and later-but-drainable cases, based on current `main`.

One coordination question before I build, since it overlaps your in-progress freshness fix. The burn-rate estimate needs a persisted prior `(sampled_at, seven_day_pct)` per account to compute `Δ7d_pct / Δt` — but `UsageStore` currently keeps only the current `lastGood` snapshot, no earlier sample. That's the same "stored samples" surface your refetch/re-verify fix touches. So:

1. **Sequencing** — would you rather I wait for the freshness fix to land and build on top of it, or is current `main` a fine base and you'll rebase that fix over mine? Either works for me; I just don't want to churn the storage shape under you.
2. **Storage shape** — does your fix already retain a timestamped weekly sample I can read, or should the adaptive PR add the minimal previous-sample field itself? Happy to match whatever shape you're introducing.

No rush on my end — I'd rather land it cleanly on your groundwork than race it.

realiti4:
@kmizzi Good timing — the freshness fix just landed as cf4d63b, so current `main` is a clean base, no sequencing needed. It's a two-phase commit (decide on the stored snapshot, refetch and re-verify before a switch fires, past `resets_at` treated as unknown) and it deliberately doesn't touch the store schema — still just the latest `lastGood`/`fetchedAt` per account. So the previous-sample field is best added in your PR — a single `prev`-style field rolled over on each successful record would do it, and store fields are additive by convention, so it's a safe change.

Two things worth knowing as you build: samples can be sparse (candidates can go ~10 min between polls, and only the active account burns quota — so your conservative-prior fallback matters), and the re-verify re-ranks on refetched data right before a switch, so computing `reachable` inside the ranking key gets you that re-check for free.

kmizzi:
Opened it as #178 — burn-rate gate on top of the merged soonest-first (with the additive `prev` sample in the store, since `cf4d63b` left the schema alone as you noted). Degrades to today's behavior until a rate is measured. Details + a note on why I made it a gate rather than re-sorting the key are in the PR.

Issue #85, opened 2026-07-04
Repeated Re-queries Cause Usage Reporting Lapse


jparr721:
I previously posted anecdotally about this, but I have some actual evidence now, given that I've been running the auto functionality. 

```
20:53:38  Account-3 (jparr721@pm.me): usage unknown (switch at 90%) | others: #1: ?, #2: ?, #4: ?
20:53:38  no switch: active-usage-unknown (1/3 before failover)
20:54:37  Account-3 (jparr721@pm.me): 53% used (switch at 90%) | others: #1: 100%, #2: 33%, #4: 54%
20:54:37  no switch: below-threshold (53% < 90%)
20:55:43  Account-3 (jparr721@pm.me): usage unknown (switch at 90%) | others: #1: ?, #2: ?, #4: ?
20:55:43  no switch: active-usage-unknown (1/3 before failover)
20:56:38  Account-3 (jparr721@pm.me): 53% used (switch at 90%) | others: #1: 100%, #2: 33%, #4: 54%
20:56:38  no switch: below-threshold (53% < 90%)
20:57:41  Account-3 (jparr721@pm.me): usage unknown (switch at 90%) | others: #1: ?, #2: ?, #4: ?
20:57:41  no switch: active-usage-unknown (1/3 before failover)
20:58:47  Account-3 (jparr721@pm.me): 53% used (switch at 90%) | others: #1: 100%, #2: 33%, #4: 54%
20:58:47  no switch: below-threshold (53% < 90%)
20:59:44  Account-3 (jparr721@pm.me): usage unknown (switch at 90%) | others: #1: ?, #2: ?, #4: ?
20:59:44  no switch: active-usage-unknown (1/3 before failover)
21:00:44  Account-3 (jparr721@pm.me): usage unknown (switch at 90%) | others: #1: ?, #2: ?, #4: ?
21:00:44  no switch: active-usage-unknown (2/3 before failover)
```

After running for a period of time, typically at least 20 minutes, the system stops reporting the usage, and it becomes pretty inconsistent. I'm trying to figure out what would cause this, considering you mentioned that the system caches, but you can see that there's multiple times in a row, typically, where it will fail to report the usage and, as a result, fail to do the automatic swap over.

jparr721:
Note this is pretty flaky. I'll get periods of time where it will be off, and then it will suddenly come back and start reporting again. I'm not sure what makes this go in and out.

realiti4:
Hi, thanks for pointing this out.

[...]

PR #94, opened 2026-07-06
fix(usage): global 429 backoff so the rate limiter can recover

deathemperor:
Addresses the recurring `http-429` loop on the usage endpoint (related to #85).

## Problem
The usage endpoint rate-limits **per IP**, but the store's backoff is **per-account only**. With multiple accounts polling on staggered schedules, whichever account's backoff just expired keeps poking the shared limit — so it never quiets. Observed live: all accounts cycling through 429s every ~2 minutes indefinitely. A single account's occasional success also resets its `consecutiveFailures`, keeping its backoff pinned at the 30s floor.

## Fix
A **shared 429 gate** in `UsageStore`:
- Any `http-429` in a fetch pass sets `rateLimit.blockedUntil = now + expo(consecutive429)` (server `Retry-After` as a floor), and the collector (`_collect_usage_entries`) skips **all** fetching until it passes — serving last-known-good meanwhile.
- A pass with a clean success **and no 429** clears it; a **mixed** pass (one 429 + one success) still backs off, so an intermittent success can't defeat the gate.
- It grows on each re-trip until it outlasts the IP window, then a probe succeeds and clears it.
- Persisted in `usage.json` (`rateLimit`) so the menubar engine, TUI, and cron `cswap auto --once` all honor one backoff.
- Per-account backoff stays for genuinely per-account errors.

## Tests
- `TestGlobalRateLimitBackoff` (10 cases): trips on 429, grows, clears on clean success, mixed-pass still backs off, non-429 ignored, Retry-After floor, persistence across instances, expiry, coexistence with per-account state.
- Collector: `test_global_backoff_skips_all_fetching` — no endpoint call while gated, last-good served.
- Updated `test_active_in_backoff_keeps_trusted_headroom` to the new semantics (a 429 quiets all polling, not just the one account).

Full suite: 898 passed, 3 skipped.

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -1710,15 +1710,23 @@ class ClaudeAccountSwitcher:
                 sentinels[num] = static
 
         entries = store.entries(identities)
-        to_fetch = [
-            num
-            for num in info_by_num
-            if num not in sentinels
-            and (fetch is None or num in fetch)
-            and not entries[num].fresh(now)
-            and not entries[num].in_backoff(now)
-            and not entries[num].claimed(now)
-        ]
+        # A prior 429 stands *all* usage fetching down for a growing interval:
+        # the endpoint rate-limits per IP, so quieting every account (not just
+        # the one that saw the 429) is what lets the limiter recover. Last-good
+        # is served meanwhile.
+        to_fetch = (
+            []
+            if store.in_global_backoff(now)
+            else [
+                num
+                for num in info_by_num
+                if num not in sentinels
+                and (fetch is None or num in fetch)
+                and not entries[num].fresh(now)
+                and not entries[num].in_backoff(now)
+                and not entries[num].claimed(now)
+            ]
+        )
 
         if to_fetch:
             store.claim(to_fetch, identities)
--- a/src/claude_swap/usage_store.py
+++ b/src/claude_swap/usage_store.py
@@ -54,6 +54,10 @@ TRUST_MAX_AGE_S = 3600.0
 BACKOFF_BASE_S = 30.0
 BACKOFF_CAP_S = 600.0
 
+# The fetch-error kind (from oauth._classify_usage_error) that means the usage
+# endpoint rate-limited us — an IP-wide signal, hence the global backoff.
+RATE_LIMIT_ERROR = "http-429"
+
 # (email, organizationUuid) — the identity a slot number currently maps to.
 Identity = tuple[str, str]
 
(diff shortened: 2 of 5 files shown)


realiti4:
Thanks for the PR and the tests. I dug into the premise because my earlier comment on #85 said this looked per-IP, and that may have sent you in the wrong direction.

I tested the usage endpoint with two accounts from the same IP. For the 429s I could reproduce, the scope looks per-account/per-org, not per-IP. There seem to be two cases:

- `Retry-After: 0`: sustained/edge limiting, intermittent 429s, tied to that account's overall Claude Code activity.
- nonzero `Retry-After`: burst limiting; about 5 rapid requests on one token caused a 300s block, with `Retry-After` counting down accurately.

In the clean isolation tests, one account being blocked did not block the other account from the same IP. So I don't want to merge a global all-account gate as-is; it can stand down healthy accounts and interfere with near-threshold refreshes that failover relies on.

That said, your symptom is real, and I checked #65: your auto-switch integration is doing the right thing. With the engine running, the display reads store-only and the engine paces fetches. The remaining gap is display-only mode: with auto-switch off, `_snapshot()` runs an on-demand pass every `refresh_interval`, so every stale account can become fetch-eligible on each display refresh. With the store's failure backoff, a limited account retries about every 2 minutes, which fits your "cycling every ~2 minutes" observation. That's not a mistake on your side; there was no paced fetch path for a display to use. I’ve just exposed one on main in `1817dcc`: `claude_swap.snapshot_source.SnapshotSource`, the same discipline the TUI uses: active account plus one due candidate per pass, with the rest served from the store. After rebasing #65, display-only mode can drive its refresh timer through one persistent `SnapshotSource.take()` instance, and pass `store_only=True` while the engine runs.

I'm also taking two smaller improvements from this: `Retry-After: 0` 429s now keep a tighter retry cadence, and nonzero `Retry-After` 429s are logged as burst-style blocks that normal cswap polling should not trigger.

Was auto-switch on or off when you saw the cycling, and what's your `refresh_interval`? If it was off, the paced snapshot source should make this disappear. And if you ever see a long-rested account 429 on its first poll, that's the log snippet that reopens the per-IP question.

deathemperor:
You're right, and my per-IP call was circumstantial. Your two-account isolation test is the real signal, and my own logs actually support per-account: my most-used account (pinned at 5h 100%) 429s while the other two succeed from the same IP — the "all three failing together" stretch I read as per-IP was just all three being heavily used at once.

The global gate's downside you describe is real: it stands down healthy accounts and interferes with the near-threshold refreshes failover relies on. My own updated test (`test_active_in_backoff_keeps_trusted_headroom` becoming "fetches nothing") was that behavior change staring back at me. Closing #94.

Your questions: auto-switch was **on**, `refresh_interval` 60 — so in my case the display was already store-only and it wasn't display-only mode. The cycling was the engine's near-threshold escalation: the active account sat at 5h 100% / 7d 96% against a 98% threshold, so escalate-all fired each tick and re-fetched candidates as their ~2-minute per-account backoffs expired. I did not see a long-rested account 429 on its first poll, so I have nothing that reopens the per-IP question.

`SnapshotSource` is the right fix for the display path. I'll rebase #65 on main and drive the menu bar's refresh timer through one persistent `SnapshotSource.take()` — `store_only=True` while the engine runs, and the paced active-plus-one-candidate pass when it's off — which removes the "every stale account eligible each refresh" path in display-only mode. The two upstream tweaks (`Retry-After: 0` cadence and burst-block logging) also cover the noise on the auto-on side.

PR #97, opened 2026-07-06
Fetch usage for session-owned accounts from the session profile's credentials (#96)

### LINKED ISSUE #96: Usage for session-mode accounts silently freezes once claude rotates tokens inside the profile
## Symptom

Accounts that have been used with `cswap run` stop updating in `list`/`watch`: their bars freeze at the last good measurement and the age note just keeps growing ("· 21h ago", "· 1d 8h ago", …). A maxed weekly window keeps showing `100% (!)` long after it actually reset. The log shows the cause repeating every pass:

```
WARNING - Usage fetch failed for account 1: http-401
WARNING - Usage fetch failed for account 2: http-401
```

## Root cause

1. `cswap run` seeds the session profile from backup, refreshing once at bootstrap (session.py `_bootstrap`) — from that point the profile and the backup share one OAuth token family.
2. While the session runs, claude rotates that family *inside the profile* — on macOS in the profile's hashed keychain entry, elsewhere in the profile's `.credentials.json`. Nothing syncs it back to the backup store (the reuse path in `setup_session` deliberately skips bootstrap).
3. Refresh tokens are single-use, so the backup now holds a consumed older generation. Every usage fetch for the account uses the backup credential and 401s; the refresh machinery in `try_fetch_usage_for_account` fails permanently too (`invalid_grant`), but only at DEBUG level — at default log level it looks identical to a plain 401.
4. The stale-serving store (correctly) keeps showing last-good data, so the user sees silently frozen bars with no hint that they can never update again.

While a session is live, `_fetch_account_usage` already deliberately skips refresh ("worst case its usage shows as unavailable until the session exits") — but the state persists *after* the session exits too, because the backup family is already dead by then.

Verified on macOS with two session-mode accounts: `--token-status` shows both backups expired (`Jul 5 12:48` / `Jul 5 23:18`), while the profiles' hashed keychain entries hold newer generations (e.g. expiry `Jul 6 07:14`) — claude rotated past the backup, exactly as above.

## Proposal

For accounts with a session profile, source usage fetches from the profile's current credential (keychain entry first, plaintext seed as fallback), strictly read-only — rotation stays claude's job. That makes usage live again whenever the profile's token is fresh (which it is whenever the session is actually being used), and turns the idle-expired case into the existing token-expired sentinel instead of an eternal silent 401 loop.

PR incoming.

Possible follow-up (separate discussion): syncing the profile's credential back to the backup store would also heal `--switch-to` for such accounts, but that touches the overwrite-safety invariants from #79, so it is left out of the PR.

jackfuchs:
Fixes #96 — full root-cause analysis there. Short version: once `cswap run` has bootstrapped a profile, claude rotates the token family inside the profile and nothing syncs it back, so the backup's refresh token is a consumed generation the server 401s forever. Usage for that account then freezes silently at the last pre-session measurement.

## Change

- **session.py**: new `read_session_credentials(session_dir)` — best-effort read of the profile's *current* credential JSON: the hashed keychain entry first (it shadows the plaintext seed from claude's first write; the service name comes from the existing `keychain_service_name`), then the plaintext `.credentials.json`. Read-only by design — cswap keeps never writing the hashed entry.
- **switcher.py** `_fetch_account_usage`: for inactive accounts, prefer the session profile's credential and fetch with `is_active=True` (no refresh, no persist) — rotating the profile's family would log the next `cswap run` out the same way the backup got burned. Two edges:
  - profile credential locally expired + live session → `USAGE_TOKEN_EXPIRED` sentinel instead of burning a request that would 401 (mirrors the owned-active rule in `_fetch_active_usage`; the running claude refreshes lazily on its next API call, after which fetches succeed again);
  - profile credential expired + no live session → fall through to the existing backup path unchanged, so a backup family that is still alive keeps serving and healing exactly as before.

Accounts without a session profile are untouched (covered by a regression test).

## Verified

On the affected setup from #96: the profile's keychain entry is found and used; a live-but-idle session now reports the honest token-expired sentinel instead of frozen bars, and flips to live data as soon as the session's claude refreshes. Full suite: 907 passed, 3 skipped.

### DIFF
--- a/src/claude_swap/session.py
+++ b/src/claude_swap/session.py
@@ -182,6 +182,36 @@ def delete_macos_keychain_entry(session_dir: Path) -> None:
         pass  # best-effort; absent entry is already success (rc 44)
 
 
+def read_session_credentials(session_dir: Path) -> str | None:
+    """Best-effort read of a session profile's *current* credential JSON.
+
+    Once a session has run, the profile — not the backup store — holds the
+    newest generation of the account's token family: claude rotates tokens
+    in place, and nothing syncs them back to backup. On macOS the rotated
+    credential lives in the profile's hashed keychain entry (which shadows
+    the plaintext seed from the moment claude first writes it), elsewhere in
+    the profile's ``.credentials.json``. Read-only by design: writing either
+    location stays claude's job (see the module docstring on why cswap never
+    writes the hashed entry). Returns ``None`` when the profile has no
+    readable credential material.
+    """
+    if not session_dir.is_dir():
+        return None
+    if Platform.detect() == Platform.MACOS:
+        try:
+            creds = macos_keychain.get_password(
+                keychain_service_name(session_dir), _keychain_account_name()
+            )
+            if creds:
+                return creds
+        except KeychainError:
+            pass  # locked/denied/timeout — the plaintext seed is the next-best truth
+    try:
+        return (session_dir / ".credentials.json").read_text()
+    except OSError:
+        return None
+
+
 def live_sessions_for(session_dir: Path) -> list[ClaudeSession]:
     """Live Claude instances running against a session profile."""
     if not session_dir.exists():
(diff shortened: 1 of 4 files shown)


realiti4:
Hi, I think it is a good solution and thanks for the fix. I've merged it.

PR #99, opened 2026-07-07
fix(switch): never back up an empty current credential (Keychain timeout)

deathemperor:
## The bug (credential loss)
On macOS a Keychain read that times out (`security find-generic-password timed out after 5.0s`) returns `""` rather than raising, so `_read_credentials()` yields an empty string. `_perform_switch` only rejected a `None` read (`if original_creds is None`), then wrote the empty value over the **departing** account's backup — destroying its stored credential.

I hit this in the wild running the menu bar's auto-switch: two accounts' backups were wiped to **0 bytes** during background switches while the Keychain was timing out, and re-adding them from a terminal was immediately undone by the next background switch-away.

```
13:29:37  Added account 2          ← restored
13:30:31  Switched from 2 to 3     ← backed up account 2 EMPTY (Keychain timeout) → wiped
13:30:56  Added account 1          ← restored
13:31:38  Switched from 1 to 3     ← backed up account 1 EMPTY → wiped
13:31:46  Switch failed: Account-2 has no stored credentials
```

## The fix
Reject an **empty** current-credential read too, before the backup step, so the switch fails harmlessly (leaving the good backup intact) instead of overwriting it. The caller retries once the Keychain settles. One-line guard alongside the existing `is None` check.

## Test
`test_switch_refuses_to_overwrite_backup_with_empty_current_creds`: a switch whose current-credential read returns `""` (as a Keychain timeout does) now raises `CredentialReadError` and leaves the departing account's backup byte-for-byte intact. Fails on `main`, passes with the fix. Full suite: 908 passed, 3 skipped.

## Note
This only guards the normal (backup-then-switch) path — the direct-activation path already snapshots for rollback only and never writes a backup. A separate follow-up could re-probe the Keychain on a transient timeout so the switch can succeed rather than defer, but preventing the data loss is the priority here.

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -2743,6 +2743,16 @@ class ClaudeAccountSwitcher:
                 original_creds = self._read_credentials()
                 if original_creds is None:
                     raise CredentialReadError("Failed to read current credentials")
+                if not original_creds:
+                    # An empty read (e.g. a macOS Keychain `security` timeout,
+                    # which returns "" rather than raising) must NOT be written
+                    # over the departing account's backup — that would destroy
+                    # its stored credential. Fail the switch; the backup stays
+                    # intact and the caller can retry once the Keychain settles.
+                    raise CredentialReadError(
+                        "Current account credential is empty (Keychain unreadable?); "
+                        "refusing to overwrite its backup"
+                    )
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for the fix, and nice catch! I’ve merged it.

PR #100, opened 2026-07-07
feat(auto): switch on a per-model weekly limit via `--model` / `autoswitch.model`

tomer-dev:
## Problem

`account_headroom` — the single function that decides *when* `cswap auto` switches and *which* account it picks — only looks at the account-wide **5h/7d** windows. Anthropic now also enforces **per-model weekly limits**, which the usage layer already parses into `scoped` windows (e.g. `{"name": "Fable", "pct": 100}`). Those were never consulted, so:

- An account whose **model quota is exhausted** but whose 5h/7d still have headroom reads as "plenty left" → `cswap auto` **never switches off it**, even though every request for that model is blocked.
- On the live shape where the usage API returns **only** a per-model window (no 5h/7d yet), `account_headroom` returns `None` → the engine reads the active account as **"usage unknown"** and sits there.

I hit this working exclusively on one model: the account was model-maxed for a day while cswap happily stayed on it.

## What this does

Adds an **opt-in** model filter — off by default, so nothing changes for anyone who doesn't set it (a per-model cap must not evict users who don't use that model).

- `oauth.account_headroom(usage, models=())` — when `models` is non-empty, each named `scoped` window is folded into the binding `max(pct)` alongside 5h/7d (case-insensitive `display_name` match). Empty tuple = today's behavior, byte-for-byte.
- `AutoSwitchSettings.model` + a new `"string"` `SettingSpec` kind — validated, round-trips through `cswap config set/get/unset autoswitch.model`; a null/garbage value in `settings.json` disables the filter rather than crashing.
- `cswap auto --model NAME` flag (overrides the setting, like the other `auto` flags).
- Threaded through the **engine only** — so both the leave-decision and target-selection agree. Manual `--switch --strategy best` and the poll-cadence heuristics deliberately stay on 5h/7d (noted in code); happy to extend those in a follow-up if you'd like.

## Example

```bash
cswap auto --model Fable
# or persist it:
cswap config set autoswitch.model Fable
```

Now an account with `5h 1% · 7d 40% · Fable 100%` is treated as at-limit and the engine moves to the account with the most model headroom.

## Tests

- `test_oauth.py`: the `models` math — folds a named window, ignores unlisted models, case-insensitive, only-scoped shape, still `None` when nothing matches.
- `test_settings.py`: string-kind set/get/unset round-trip, empty-value rejection, garbage→`None` clamp; the existing registry/defaults tests cover the new field automatically.
- `test_config_cli.py`: key-count assertions bumped 7→8.
- `test_autoswitch.py`: new `TestModelAwareSwitch` — switches when the model is maxed despite session headroom, **stays put** with no model configured on the same usage, and still respects a maxed 5h window.

Full suite: **921 passed, 3 skipped**.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Let claude-swap watch your usage and switch for you. When the active account's 5
 ```bash
 cswap auto                     # foreground loop, polls every 60s
 cswap auto --threshold 80      # switch earlier
+cswap auto --model Fable       # also switch when the Fable weekly limit is hit
 cswap auto --once              # single check-and-switch, for cron/scripts
 cswap auto --dry-run           # log what it would do, never switch
 ```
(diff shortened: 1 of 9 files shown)


realiti4:
Merged, thank you.

PR #101, opened 2026-07-07
fix(keychain): re-probe after a cooldown so a daemon recovers from timeouts

deathemperor:
## Problem
The macOS Keychain capability cache drops to file mode on the first `security` failure and **sticks for the whole process** (so one CLI invocation can't split-brain between backends — correct for a sub-second command). But in a long-running daemon (the menu bar, `cswap watch`), a single transient `security find-generic-password timed out after 5.0s` disables the Keychain **permanently**: the active account's credential becomes unreadable and the dashboard shows "keychain unavailable" / "no credentials" indefinitely — while a fresh `cswap list` in a terminal reads it fine.

Observed running the menu bar: a timeout cluster left the active account stuck showing the error until the app was relaunched.

## Fix
Record a re-probe deadline when a failure drops to file mode (`KEYCHAIN_RECHECK_COOLDOWN_S = 60s`). `_use_keychain()` re-probes once the cooldown elapses:
- A CLI command runs far shorter than the cooldown → it never re-probes mid-invocation, so the no-split-brain guarantee is preserved.
- A daemon self-heals within a minute of the Keychain recovering.
- A directly-forced file mode with no deadline stays sticky (unchanged).

## Tests
- `test_kc_call_failure_schedules_a_recheck`, `test_keychain_recovers_after_cooldown`, `test_keychain_stays_file_mode_during_cooldown`.
- Existing `test_capability_cache_sticky_false` / `_is_process_local` still pass (a forced `False` with no deadline stays sticky). Full suite: 910 passed.

Pairs with #99 (empty-credential backup guard): #99 stops a Keychain timeout from *destroying* credentials on a switch; this stops it from making them *unreadable* for the daemon's lifetime.

### DIFF
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -161,7 +161,10 @@ class CredentialStore:
             result = fn(*args)
         except macos_keychain.KEYCHAIN_ERRORS:
             self._keychain_usable_cache = False
-            self._keychain_disabled_until = time.time() + KEYCHAIN_RECHECK_COOLDOWN_S
+            # Monotonic so a wall-clock jump can't expire the cooldown early/late.
+            self._keychain_disabled_until = (
+                time.monotonic() + KEYCHAIN_RECHECK_COOLDOWN_S
+            )
             raise
         if self._keychain_usable_cache is None:
             self._keychain_usable_cache = True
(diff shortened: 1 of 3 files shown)


realiti4:
Hi, thanks a lot for the PR. This is a real issue.

There is one edge I’d want to close before merge. The re-probe also re-enables Keychain reads after an active credential write fell back to the file. If a switch happens while the Keychain is flaky, we write the fresh creds to `.credentials.json`, but the best-effort delete of the old Keychain item may also fail. Once the cooldown elapses, a read can prefer the Keychain again and resurrect that stale item, showing the wrong account. Today that residual stays buried because file mode is sticky, and the re-probe brings it back.

Scoping the re-probe to read failures alone does not fully cover it either, since a deadline scheduled by an earlier read can still be pending when a later write leaves the stale item behind.

Could you update the PR to pin file mode after an active write falls back, so we never re-probe onto a Keychain we could not verify-clear? That should preserve the pure read-timeout recovery this PR is targeting. I’d also switch the cooldown to `time.monotonic()` so a wall-clock jump cannot expire it early or late.

Once that guard is in, I think this should be in good shape to merge.

deathemperor:
Good catch — that resurrection edge is real. Done:

- `_pin_file_mode()` now pins file mode (drops to `False`, clears any pending re-probe deadline) whenever an active OAuth write falls back to the file — so once we couldn't verify-clear the old Keychain item, we never re-probe onto it. This also covers your second point: a deadline a prior read scheduled is cleared by the write, since the pin runs in `_write_oauth_credentials`'s file-fallback path regardless of how file mode was entered.
- Pure read-timeout recovery is untouched — a read that times out (no write) still re-probes after the cooldown.
- Cooldown switched to `time.monotonic()`, so a wall-clock jump can't expire it early or late.

Tests: `test_write_keychain_failure_pins_file_mode` (Keychain write fails → fall back → pinned, no re-probe despite the failed op scheduling one) and `test_write_fallback_clears_pending_read_reprobe` (already in file mode with a pending read re-probe → a write with a failed stale-item delete clears it). Existing read-recovery tests still pass. Full suite: 912 passed.

realiti4:
Thanks for the update. The OAuth fallback guard and monotonic cooldown look good.

I found one remaining variant of the same stale-Keychain case on the managed API-key path. In `_write_managed_credentials`, if `set_password` fails, we fall back to writing the fresh key to `primaryApiKey` and mark the backend as `"file"`, but we don’t pin file mode there. That leaves the re-probe deadline from the failed Keychain op in place.

Since managed-key reads check the Keychain before `primaryApiKey`, a daemon can re-probe after the cooldown and read an old Keychain item instead of the fresh fallback value. I reproduced this on the latest PR head with a stale Keychain key and fresh config key; after the cooldown, the read returned the stale key.

Could you mirror the OAuth fix in the managed-key fallback path? After the fallback config write succeeds, call `_pin_file_mode()` when `self._host.platform == Platform.MACOS` and `not wrote_to_keychain`, and add a managed-key version of the fallback regression test.

After that, I think this is ready to merge.

deathemperor:
Fixed — mirrored the guard on the managed-key path. `_write_managed_credentials` now calls `_pin_file_mode()` after the fallback config write when `self._host.platform == Platform.MACOS and not wrote_to_keychain`, so a failed `set_password` that falls back to `primaryApiKey` can't leave a re-probe pending onto a stale "Claude Code" Keychain item.

Added `test_managed_key_write_fallback_pins_file_mode` (Keychain write fails → falls back to config → pinned, no re-probe). Full suite: 913 passed.

realiti4:
Thanks for the adjustments, looking good I've merged it.

PR #104, opened 2026-07-07
Recover the display faster from edge-429s, then throttle a saturated account (#103)

### LINKED ISSUE #103: Active account's usage bars freeze for minutes during sustained-edge (retry-after:0) 429s
## Symptom

When you're actively working on the currently-active account, its usage bars in `cswap watch` (and the `list` age) can freeze for several minutes — bars go grey (stale-dimmed), the "· Nm ago" note keeps climbing — before they snap back. It happens exactly when you most want live numbers: while you're burning through that account's budget. Meanwhile claude.ai's own usage page updates in real time, because it reads usage internally rather than through the rate-limited `/usage` endpoint cswap has to poll.

## Root cause

The active account is polled every pass. When its own Claude Code activity has its budget at the edge, the usage endpoint answers `HTTP 429` with `Retry-After: 0` — the sustained/edge rule (retries are penalty-free, ~every other one succeeds). The current edge backoff still runs the exponential curve, capped at `EDGE_BACKOFF_CAP_S = 120s`:

```
consecutive edge miss:   1    2    3    4    5    6 ...
backoff (s):            30   60  120  120  120  120
```

So a run of edge misses parks the account for up to 120s per attempt. Over a handful of consecutive misses that adds up to ~10 minutes before a poll gets through — measured on a real account (12× `retry-after:0` in one burst while the account was in active use).

The irony: the backoff meant to protect against 429s makes the *active* account — the one you can't avoid re-polling — the one that freezes longest.

## Proposal

Since edge (`retry-after:0`) misses are penalty-free and ~half succeed, keep the first few retries very short so the display recovers in seconds, then ramp the wait up only once the account is genuinely saturated (so we neither hammer a stuck limit nor let the active account poll into the wall every pass):

```
consecutive edge miss:   1    2    3    4    5    6    7    8
backoff (s):            15   15   15   30   45   60   75   90
```

Worst-case recovery over a 6-miss burst drops from ~9.5 min to ~3 min; a single miss from 30s to 15s. Burst-rule 429s (`Retry-After: N>0`) are unchanged.

PR incoming.

Runs absolute smooth now:

Image

jackfuchs:
Fixes #103.

Edge 429s (`Retry-After: 0`) are penalty-free and ~half succeed, so the exponential-to-120s curve was the wrong shape: it froze the active account's bars for minutes exactly during heavy burn. This keeps the first `EDGE_FAST_RETRIES` (3) misses at `EDGE_BACKOFF_FAST_S` (15s) — the display recovers in seconds — then ramps toward a lowered `EDGE_BACKOFF_CAP_S` (90s), which both stops hammering a genuinely stuck limit and throttles the active account so it isn't polling into the wall every pass.

```
consecutive edge miss:   1    2    3    4    5    6    7    8
before (s):             30   60  120  120  120  120  120  120
after  (s):             15   15   15   30   45   60   75   90
```

Only the `retry-after:0` branch of `_failure_backoff_s` changes. The exponential curve for generic errors, the burst rule (`retry-after > 0`, honored as the floor), and the retry-after safety cap are all untouched.

Tests: the edge-backoff test now asserts the fast-then-ramp sequence, plus a guard that a single edge miss recovers at 15s. Full suite green.

### DIFF
--- a/src/claude_swap/usage_store.py
+++ b/src/claude_swap/usage_store.py
@@ -58,14 +58,21 @@ BACKOFF_CAP_S = 600.0
 # Retry-After value:
 # - "Retry-After: 0" = the sustained/edge rule: the account's overall Claude
 #   Code activity has its budget at the edge; retries are penalty-free and
-#   ~every other one succeeds. Long exponential backoff only costs freshness
-#   during exactly the heavy-burn periods where it matters most, so edge
-#   backoff is capped low.
+#   ~every other one succeeds. So the first few retries stay very short
+#   (EDGE_BACKOFF_FAST_S) to keep the display fresh during exactly the
+#   heavy-burn periods where it matters most; only after EDGE_FAST_RETRIES
+#   consecutive misses — the account is genuinely saturated, not just busy —
+#   does the wait ramp up (toward EDGE_BACKOFF_CAP_S) so we stop hammering a
+#   limit that clearly isn't clearing. This ramp also throttles the active
+#   account, which is polled every pass and would otherwise poll straight into
+#   the wall.
 # - "Retry-After: N>0" = the burst rule (~5 rapid requests on one account →
 #   hard 300s block; measured: accurate, counts down, not extended by
 #   probing). Honored as the wait, up to a safety cap so a pathological
 #   header can never park an account for hours.
-EDGE_BACKOFF_CAP_S = 120.0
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for the PR. Findings shared here: https://github.com/realiti4/claude-swap/issues/103#issuecomment-4948419182

Closing in favor of 245f5bae8444daaecc5fbc97007e6f6af43989a3

PR #106, opened 2026-07-07
Quarantine dead-refresh-token accounts and surface 're-login needed' (#105)

### LINKED ISSUE #105: Accounts with a dead refresh token freeze silently and retry forever (no 're-login needed' signal)
## Symptom

An account whose stored refresh token has died (e.g. after its session profile rotated the token family, or the backup credential aged out) does two unhelpful things:

1. Its usage bars in `cswap watch`/`list` freeze at the last-good measurement and just keep aging — "· 18h ago", "· 1d 8h ago", … — with **no indication** that they can never update again. It looks identical to a briefly-stale account.
2. cswap keeps polling it every backoff cycle (~10 min) forever. Each attempt refreshes → the server rejects it → and the repeated attempts additionally draw fresh `429`s on the account.

Measured on a real dead account, the token endpoint answers definitively:

```
HTTP 400  {"error": "invalid_grant", "error_description": "Refresh token not found or invalid"}
```

That is a permanent rejection — no transient 429/timeout/network blip produces it — yet the fetch loop treats it like any other failure and retries indefinitely.

## Proposal

Detect a dead refresh-token lineage and quarantine the account:

- Surface `invalid_grant` distinctly from a transient refresh failure. One server-confirmed `invalid_grant` is definitive, so it's enough to mark the token dead (transient errors never do).
- Once dead, stop fetching the account (no more endless 401/429), and show an explicit **"⚠ re-login needed"** state instead of silently-aging bars — while still keeping the last-seen measurement visible for context.
- Clear the quarantine automatically when the credential is refreshed (`cswap add` / re-add), so recovery needs no manual cache surgery.

Rendered, the quarantined account reads like this instead of frozen bars:

```
 3  user@example.com  [personal]   · 19h ago
    ⚠ re-login needed — refresh token dead; run: cswap add
    └ last seen 53% used · 19h ago
```

Related to #96, where this silent-freeze mode first came up as a follow-up. PR incoming.

jackfuchs:
Fixes #105.

A refresh-token lineage the server rejects with `invalid_grant` can't recover on its own, but the fetch loop retried it forever — silently aging the bars and drawing fresh 401/429s each cycle. This detects it and quarantines the account.

## Change

- **oauth**: `try_fetch_usage_for_account` now surfaces `invalid_grant` distinctly from a transient `refresh-failed` (via `try_refresh_oauth_credentials`). The proactive-refresh path short-circuits on `invalid_grant` **without hitting the usage endpoint**, so a dead token stops drawing 429s immediately.
- **usage_store**: a consecutive-`invalid_grant` strike count. One server-confirmed `invalid_grant` ("Refresh token not found or invalid") is definitive, so `AUTH_DEAD_STRIKES = 1`; transient errors (429/timeout) neither advance nor reset it. `token_dead()` gates `due_candidate`; `clear_dead_token()` lifts the quarantine; a success resets the count.
- **switcher**: the collector surfaces `USAGE_RELOGIN_REQUIRED` for dead tokens — which both drives the display and (via `num not in sentinels`) stops the fetch loop. `add_account`/`add_account_from_token` clear the quarantine on a fresh credential, so recovery is just a re-add — no cache surgery.
- **TUI/list**: the sentinel renders like the other warn states (⚠ + note + last-seen line); no per-surface special-casing.

## Notes

- Purely additive: healthy accounts and every existing sentinel path are untouched.
- `AUTH_DEAD_STRIKES` is a one-line knob — raise to 2 if you'd prefer a buffer against a one-off misclassification (the trade-off is a ~10-min-slower verdict, one failure-backoff between strikes).

Tests cover: invalid_grant propagation (proactive short-circuit + 401-retry) vs. transient staying `refresh-failed`; strike counting/reset/transient-immunity; `due_candidate` skipping a dead token; the collector surfacing the sentinel and skipping the fetch; and re-add clearing the quarantine. Full suite green (960 passed).

### DIFF
--- a/src/claude_swap/json_output.py
+++ b/src/claude_swap/json_output.py
@@ -27,6 +27,11 @@ USAGE_API_KEY = "api key"
 # with no plaintext fallback — distinct from a genuinely empty slot, so the user
 # isn't misled into an unnecessary re-login.
 USAGE_KEYCHAIN_UNAVAILABLE = "keychain unavailable"
+# The stored refresh-token lineage is dead (repeated ``invalid_grant``). The
+# account is quarantined from fetching until a re-login (``cswap login`` / ``add``)
+# replaces the credential; distinct from "token expired" (which Claude Code can
+# refresh on its own) because only the user can fix it.
+USAGE_RELOGIN_REQUIRED = "re-login needed"
 
 
 def _window_to_json(entry: dict) -> dict:
@@ -99,6 +104,8 @@ def usage_fields(entry: dict | str | None) -> tuple[str, dict | None]:
         return "api_key", None
     if entry == USAGE_KEYCHAIN_UNAVAILABLE:
         return "keychain_unavailable", None
+    if entry == USAGE_RELOGIN_REQUIRED:
+        return "relogin_required", None
     if isinstance(entry, str):
         return "no_credentials", None
     return "unavailable", None
--- a/src/claude_swap/oauth.py
+++ b/src/claude_swap/oauth.py
@@ -408,12 +408,20 @@ def try_fetch_usage_for_account(
         and oauth.get("refreshToken")
         and is_oauth_token_expired(oauth.get("expiresAt"))
     ):
-        refreshed = refresh_oauth_credentials(working_credentials)
-        if refreshed:
-            working_credentials = refreshed
+        refresh = try_refresh_oauth_credentials(working_credentials)
+        if refresh.credentials:
+            working_credentials = refresh.credentials
             _persist(persist_credentials, account_num, email, working_credentials)
             oauth = extract_oauth_data(working_credentials) or oauth
             access_token = oauth.get("accessToken") or access_token
+        elif refresh.error == "invalid_grant":
+            # The refresh-token lineage is server-rejected — permanently dead.
+            # Don't hit the usage endpoint with a token we know is expired
+            # (that just adds a 401/429 to a lost cause): report the permanent
+            # failure distinctly so the store can quarantine the account.
+            return UsageOutcome(None, error="invalid_grant")
+        # A transient refresh failure falls through to try the (expired) token;
+        # the 401 path below retries the refresh.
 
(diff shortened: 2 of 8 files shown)


jackfuchs:
Cross-ref for context: #108 (standalone `cswap login`) adds a `login_and_add` path that, if both land, should also call the `clear_dead_token` helper introduced here, so a re-login clears the quarantine. This PR is self-contained without #108; noted the one-line follow-up over on #108.

realiti4:
Hi, thanks a lot for the PR! I think this would be a really good improvement.

First, the PR says `add_account_from_token` clears the quarantine too, but I don’t think the patch currently does that. `cswap add-token ... --email user@example.com` can refresh the same account in place, but the old `authDeadStrikes` row remains, so the account can stay stuck at `re-login needed` and never fetch the new token. Can you call `clear_dead_token()` from the `add-token` update/new-write paths as well?

Second, the new message says run: `cswap login`,  can we say something like “log in with Claude Code, then run: `cswap add`.” in case we don't land #108 and this PR is fully self-contained.

Small optional polish: after a fetch returns `invalid_grant`, the same `list`/`watch` refresh doesn't render `re-login needed` until the next pass. It would be nicer if it showed immediately.

jackfuchs:
Thanks — all three are in.

**1.** You're right, the body oversold it. `clear_dead_token()` now fires on both `add-token` paths (refresh-in-place and new/overwrite), matching `add_account`. Added tests for a refreshed token lifting the quarantine and for a reused slot dropping a stale strike row.

**2.** Message is now `log in with Claude Code, then run: cswap add` — self-contained regardless of #108.

**3.** Did the polish too: after a fetch returns `invalid_grant` the collector re-checks the fetched slots and surfaces `re-login needed` in the same pass instead of waiting for the next refresh.

Full suite green (963 passed).

realiti4:
Thanks for the adjustments! It is looking good, I've merged it.

Issue #107, opened 2026-07-07
Offer: standalone 'cswap login' (browser OAuth) to skip the log-out/log-in/add dance


jackfuchs:
## Upfront, so you can judge it fairly

I built this for myself, out of pure convenience, and I'm fully aware it's on shakier ground than the rest of what I've sent — it reuses Claude Code's own OAuth client id and endpoints, so it's a reverse-engineered flow, not a supported API, and it'll break the day Anthropic changes it. It doesn't touch any of the other features (the backoff and dead-token PRs stand on their own). But it gives me such a big day-to-day win that I wanted to at least put it on the table and let you decide. Totally fine if it's a no.

## The itch

Adding an account today means the ping-pong: log into Claude Code as account A → `cswap add` → log out → log in as account B → `cswap add` → … For someone juggling several accounts (and, in my case, two orgs under the same email) that's a lot of session churn.

`cswap add-token` sidesteps the session dance, but a `claude setup-token` is inference-only server-side — an account added that way shows **no usage** in `list`/`watch`, which is most of why I run cswap.

## What this does

```
cswap login              # add an account straight through the browser, no ping-pong
cswap login --private    # open the OAuth page in an incognito window, so a second
                         # account logs in without disturbing the claude.ai session
                         # your normal browser is already signed into
```

It runs the same PKCE authorization-code flow Claude Code uses, so the resulting credential carries the **full scopes** — refresh token, expiry, and the account/org metadata usage tracking needs. So unlike `add-token`, an account added via `cswap login` keeps showing its 5h/7d/Fable usage. And it never touches the live `~/.claude` login — it just writes a slot.

`--private` is the part I actually reach for daily: with two accounts on one email, an incognito window lets me pick the *other* one cleanly without logging my main session out.

## Honest caveats (so this isn't a surprise later)

- **Reverse-engineered.** Client id + endpoints are Claude Code's. If you'd rather not carry that maintenance/ToS surface in the project, I completely understand.
- **Fragile by nature.** If Anthropic changes the flow, `login` breaks (the rest of cswap doesn't).
- **Self-contained.** It's one new module + a slot-writer; nothing else depends on it, so it's easy to drop or gate behind a flag if you want it opt-in.

PR is up if you want to look, but no pressure to take it — I mostly wanted to share it in case it's useful to others too.

realiti4:
[...]

PR #108, opened 2026-07-07
Add 'cswap login': standalone browser OAuth login with optional --private (#107)

jackfuchs:
For #107 — please read the issue first for the honest framing: I built this for my own convenience, I know it's a reverse-engineered flow that can break if Anthropic changes it, and it doesn't touch any other feature. Offering it in case it's useful; a "no thanks" is completely fine.

## What it adds

```
cswap login              # add an account via the browser, skipping log-out/log-in/add
cswap login --private    # open the OAuth page in an incognito window
```

Runs Claude Code's own PKCE authorization-code flow, so the credential has the full scopes — refresh token, expiry, account/org metadata — and the added account keeps showing usage in `list`/`watch` (unlike `add-token`'s inference-only setup-tokens). Never touches the live `~/.claude` login; it only writes a slot, mirroring `add_account`'s identity/slot handling (existing (email, org) refreshes in place, else next free slot / `--slot`).

`--private` picks the first installed incognito-capable browser (Chrome/Brave/Edge/Chromium/Vivaldi via `--incognito`/`--inprivate`, Firefox via `--private-window`) and launches the binary directly, so the flag is honored even when the browser is already running; it warns and falls back to manual paste when none is found (Safari/Arc have no reliable private-window CLI flag).

## Containment

- One new self-contained module (`standalone_login.py`) + a `login_and_add` slot-writer; nothing else in the codebase depends on it.
- Easy to gate behind an opt-in flag or drop entirely if you'd prefer not to carry the reverse-engineered surface — happy to adjust.

## Tests

PKCE challenge/URL/code-parse/credential-build; the private-browser resolver preference order + fallback (no real window opened); `run_login_flow` private vs. non-private path; and the slot-writer (new slot without hijacking active, same-email-different-org → separate slots, re-login in place, empty-email rejected). Full suite green (967 passed).

### DIFF
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -45,6 +45,7 @@ _SUBCOMMAND_FLAGS = {
     "status": "--status",
     "add": "--add-account",
     "add-token": "--add-token",
+    "login": "--login",
     "remove": "--remove-account",
     "rm": "--remove-account",
     "export": "--export",
@@ -720,6 +721,22 @@ The original flag spellings (%(prog)s --switch, %(prog)s --list, ...) keep worki
         const="",
         help=argparse.SUPPRESS,
     )
+    group.add_argument(
+        "--login",
+        action="store_true",
+        help=(
+            "Add an account via a standalone browser OAuth login — no Claude Code "
+            "log-out/log-in dance. Full scopes, so usage still shows in list."
+        ),
+    )
+    parser.add_argument(
+        "--private",
+        action="store_true",
+        help=(
+            "With --login: open the browser in a private/incognito window so you "
+            "can log in as another account without touching your normal session."
+        ),
+    )
 
     args = parser.parse_args(argv)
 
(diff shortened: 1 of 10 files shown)


jackfuchs:
Heads-up on the interaction with #106 (dead-token quarantine), in case both land:

This PR's `login_and_add` deliberately does **not** call `clear_dead_token` — that helper is introduced in #106, and each PR is self-contained on its own. But if both merge, `login_and_add` should also lift the quarantine on re-login; otherwise a `cswap login` that refreshes a dead-token account wouldn't clear its "re-login needed" state (the disabled fetches can't clear the strike count on their own). It's a one-liner — happy to fold it into whichever of the two merges second, or send it as a small follow-up once the merge order is known.

jackfuchs:
Pushed a small follow-up (4d051f1):

**--help:** `login` was never listed in the commands block — the subcommand worked but was undocumented, and its `--login` mirror flag was left visible (unlike `--add-token`/`--menubar`/etc.). Now `login` is documented with an epilog example, `--login` is suppressed to match the other subcommand flags, and the real `--private` modifier stays visible.

**Quarantine (the #106 interaction):** folded in the one-liner as promised — `login_and_add` now lifts a dead-token quarantine on the refreshed slot (both in-place and new/overwrite paths), so a `cswap login` can actually recover a dead-token account. `clear_dead_token` lives in #106, so the call is guarded via `getattr`: a no-op until that helper is present, then it activates once both land. Keeps this PR self-contained and mergeable in either order — trivial to simplify to a direct call once #106 is in.

Full suite green.

PR #119, opened 2026-07-09
Ignore a session profile that drifted to another account when fetching usage

jackfuchs:
Follow-up to #97. The session-aware fetch prefers a session profile's credential as the slot's freshest token family — correctly so. But an in-session `/login` re-points the profile at a *different* account (a real workflow: the slot's account hits its 5h limit mid-session, and re-logging inside the session is the only way to keep the conversation). The profile directory keeps claiming the original slot, so the fetch records the other account's usage under this slot's label: two rows render byte-identical bars (down to the same reset second), and the slot's real account silently stops being tracked.

The profile's `.claude.json` `oauthAccount` records who claude is actually logged in as, right next to the credential. Before using session credentials, the fetch now compares it against the slot's identity — email must match, org only when both sides have a value (same leniency as `_is_session_valid`). On a confirmed mismatch it falls back to the backup credential, with refresh allowed: the live session no longer holds this slot's family, so rotating the backup can't log anything out. An unreadable identity is not treated as drift, so a profile with a broken `.claude.json` keeps today's behavior.

Deliberately *not* gated on the `.cswap-stale-credentials` marker: that is also set on benign same-account rotations, where the profile is still the right source.

Tests cover the identity/drift semantics (incl. same-email-different-org and lenient org matching), the drifted-profile fallback (refreshable backup fetch), and the matching/unreadable cases keeping the read-only profile fast path. Full suite green (983 passed).

### DIFF
--- a/src/claude_swap/session.py
+++ b/src/claude_swap/session.py
@@ -212,6 +212,51 @@ def read_session_credentials(session_dir: Path) -> str | None:
         return None
 
 
+def read_session_identity(session_dir: Path) -> tuple[str, str] | None:
+    """Best-effort read of the account identity a session profile is logged in as.
+
+    Claude records the logged-in account in the profile's ``.claude.json``
+    ``oauthAccount`` and rewrites it on every (re-)login, so this reflects the
+    profile's *current* identity — which an in-session ``/login`` can re-point
+    at a different account than the slot the profile was created for. Returns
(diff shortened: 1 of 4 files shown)


realiti4:
Thanks for fixing this case. It looks good, and I’ve merged it.

PR #120, opened 2026-07-10
feat(menubar): weekly-limit display — per-model limits (Fable) + reflect passed resets

deathemperor:
The usage API's `limits` array carries per-model weekly windows, which `oauth.build_usage_result` already normalizes into `usage["scoped"]` (each `{name, pct, resets_at, ...}`, e.g. `name: "Fable"`). The CLI `cswap list` and the TUI already render these, but the menu bar's `usage_summary` didn't — so a maxed model wasn't visible there.

This appends each scoped window to the account row after 5h/7d, matching the CLI's at-limit `(!)` marker and the menu bar's live reset countdown:

```
5h 82% · 7d 12% · Fable 4% (1d 11h)
```

Menu-bar-only change (no core touched — the data was already parsed). Tests: `usage_summary` with a scoped window, the `(!)` over-limit marker, and multiple scoped models with a countdown. Full suite: 974 passed, 3 skipped.

### DIFF
--- a/src/claude_swap/menubar.py
+++ b/src/claude_swap/menubar.py
@@ -20,7 +20,7 @@ import re
 import threading
 import time
 from dataclasses import asdict, dataclass, fields
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 
 from claude_swap.exceptions import ClaudeSwitchError, CredentialReadError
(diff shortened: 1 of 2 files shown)


deathemperor:
Pushed a second, related commit and retitled the PR — both changes are about the menu bar's **weekly-limit display**.

**Reflect a passed weekly reset from the static 7-day cadence.**

Weekly limits (the `7d` window, and the per-model `scoped` windows this PR adds) reset on a fixed 7-day cadence. Once the stored `resets_at` is in the past, the window has objectively rolled over — the cached pct belongs to a window that no longer exists. Previously the menu bar kept showing that stale pre-reset pct *and* dropped the countdown entirely (`_live_countdown` returns `None` for a passed reset). So after every weekly reset you couldn't see the new window until something forced a fresh fetch of that account — which, for a passive account under the paced `SnapshotSource`, could be a long while.

`_rolled_weekly_window()` advances a passed reset to its next 7-day boundary and shows the window as reset (`pct` 0), so the display follows the static schedule with no refetch:

```
before:  7d 96%          ·  Fable 100% (!)     (stale pct, countdown gone)
after:   7d 0% (5d 0h)   ·  Fable 0% (5d 0h)
```

Applied to the `7d` window and the `scoped` windows in `usage_summary`, and to the title's `7d` percentage. The `5h` window is deliberately untouched — its reset is session-relative, not a fixed cadence, so it can't be rolled. A real fetch still overwrites the rolled values as soon as one lands.

Full suite: 980 passed, 3 skipped.

realiti4:
Thanks for the PR, looking good merged it.

PR #123, opened 2026-07-11
feat(tui): themeable dashboard — cswap-dark + 4 Catppuccin flavors

### LINKED ISSUE #122: TUI has no theming support (one hardcoded theme, command palette disabled)
The TUI dashboard (`cswap tui`/`cswap watch`) only has one look: `CSWAP_DARK` in `tui/theme.py` is the sole registered `Theme`, and the command palette is explicitly disabled in `tui/app.py`:

```python
# No command palette: actions live in the dashboard's nested menu, in
# their own context — not in a global searchable list. This also drops
# Textual's system commands (theme picker included; there is one theme).
ENABLE_COMMAND_PALETTE = False
```

There's no way to match the dashboard to a terminal color scheme (e.g. Catppuccin, Solarized, etc.) short of hand-patching the installed package.

I have a working fix on a fork branch: registers `cswap-dark` (unchanged default) plus the four Catppuccin flavors (Latte, Frappé, Macchiato, Mocha) as selectable Textual themes, re-enables the command palette so Textual's built-in "Change theme" picker becomes available, persists the choice in `settings.json` (`tui.theme`), adds a `--theme` override flag for `tui`/`watch`, and makes the CLI's own accent color (`printer.py`) follow the same setting.

Opening a PR shortly with the implementation.

anhede:
Closes #122.

## Summary

- Registers `cswap-dark` (unchanged default) plus Catppuccin Latte/Frappé/Macchiato/Mocha as selectable Textual themes.
- Re-enables the command palette (`ctrl+p`) so Textual's built-in "Change theme" command becomes available — the old comment justifying disabling it ("there is one theme") no longer holds.
- Persists the choice as a new `tui.theme` section in `settings.json`, readable/writable through the existing generic `cswap config` machinery.
- Adds a `--theme NAME` flag to `cswap tui`/`cswap watch` that overrides the persisted setting for one session.
- The CLI's own accent color (`printer.py`, previously a fixed 256-color escape) now follows the same setting, using a 24-bit truecolor escape since the Catppuccin accents aren't xterm-palette colors.

### A note on scope

`widgets.py`/`autoview.py` previously imported `theme.py`'s colors (`ACCENT`, `MUTED`, `SEV_WARN`, ...) as fixed module constants and used them directly as Rich `Text` styles — bypassing Textual's `$variable` CSS system entirely. Just registering new `Theme` objects would only have recolored Textual's own chrome (borders, footer keys, block cursor) and left the usage bars, account list, and severity colors — most of what's actually on screen — stuck on `cswap-dark`'s palette regardless of the selected theme. Both modules now resolve colors from the active theme at render time (`current_theme_colors()`), falling back to `cswap-dark`'s colors outside a running app (e.g. the existing unit tests that call these render functions directly), so switching themes actually recolors everything.

Theme hex data lives in a new dependency-free `claude_swap/palette.py`, imported by both `tui/theme.py` (which builds the Textual `Theme` objects) and `settings.py`/`printer.py`. Those two are on the hot path for every plain CLI invocation (`cswap list`, `cswap auto --once`, ...), and `tui/__init__.py` already documents a deliberate policy of keeping textual/rich imports out of that path — `palette.py` keeps that intact.

Catppuccin hex values cross-checked against the official palette (`catppuccin/palette`, `palette.json`).

## Test plan

- [x] `uv run pytest` — 1109 passed, 3 skipped
- [x] Pilot-driven manual check: all 5 themes register, command palette opens/closes without crashing, `current_theme_colors()` reflects the active theme and updates on switch, `cswap-dark` unchanged after reverting
- [x] `cswap config set tui.theme <name>` + `cswap config unset` round-tripped through `load_tui_settings`/`effective_settings`
- [x] `printer.accent()` verified to follow `tui.theme` across `catppuccin-mocha`, `catppuccin-latte`, and unset (falls back to `cswap-dark`'s original terracotta) — exact truecolor escapes checked against each flavor's official peach hex
- [x] Persisted-setting vs `--theme` CLI override precedence verified directly against `CswapApp`

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ Run `cswap` on its own (or `cswap tui`) for the full-screen dashboard: live usag
 
 <img src="assets/tui-watch.png" width="760" alt="cswap watch — live 5h/7d usage bars for every account, with reset times and the active account marked">
 
+Open the command palette (`ctrl+p`) and run "Change theme" to pick between `cswap-dark` (default) and the four Catppuccin flavors (Latte, Frappé, Macchiato, Mocha) — the choice persists via `cswap config set tui.theme <name>`, or override it for one session with `cswap tui --theme <name>` / `cswap watch --theme <name>`. The CLI's own accent color follows whichever theme is set.
+
 
 ### Refresh expired tokens
 
@@ -213,6 +215,7 @@ cswap config                              # list effective settings ("(default)"
 cswap config get autoswitch.threshold
 cswap config set autoswitch.threshold 80  # validated: rejects out-of-range values loudly
 cswap config set autoswitch.model Fable   # per-model switching (see "auto"); Fable,Opus for several
+cswap config set tui.theme catppuccin-mocha  # persisted TUI theme; also sets the CLI's accent
 cswap config unset autoswitch.threshold   # back to the default
 cswap config path                         # where settings.json lives
 ```
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -11,6 +11,7 @@ from claude_swap import __version__
 from claude_swap.exceptions import ClaudeSwitchError
 from claude_swap.json_output import error_envelope
 from claude_swap.printer import dimmed, error, force_utf8_output, muted
+from claude_swap.settings import TUI_THEME_CHOICES
 from claude_swap.switcher import ClaudeAccountSwitcher
 
 
@@ -626,6 +627,14 @@ The original flag spellings (%(prog)s --switch, %(prog)s --list, ...) keep worki
         ),
     )
     parser.add_argument(
+        "--theme",
+        choices=TUI_THEME_CHOICES,
+        help=(
+            "With 'tui'/'watch': theme for this session only, overriding the "
+            "persisted tui.theme setting"
+        ),
+    )
+    parser.add_argument(
         "--slot",
         type=int,
         metavar="NUM",
@@ -788,6 +797,9 @@ The original flag spellings (%(prog)s --switch, %(prog)s --list, ...) keep worki
             "'switch --strategy next-available'"
         )
 
+    if args.theme is not None and not (args.tui or args.watch):
+        parser.error("--theme can only be used with 'tui' or 'watch'")
+
     if args.slot is not None and not (args.add_account or args.add_token is not None):
         parser.error("--slot can only be used with 'add' or 'add-token'")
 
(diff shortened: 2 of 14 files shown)


realiti4 (changes requested):
Hi @anhede, thanks. The render-time color handling looks right. palette.py, the settings section generalization, and the --theme override also look good, and I'd like to merge this.

One direction change: I disabled the command palette deliberately and want to keep it off. Actions in this TUI live in the dashboard's nested menu and stay in context. The palette adds a second, parallel action system, where `Quit`, `Keys`, `Screenshot`, and `Maximize` become searchable commands alongside the menu.

Could you move the picker into the menu instead? A `Theme…` entry in `_root_entries()` (`dashboard.py`, around line 80) could open a submenu containing the five theme names. `Remove account…` already uses this pattern, so the menu-stack plumbing is there.

Selecting a theme can set `app.theme`, and the existing `watch_theme` hook should persist it to `settings.json` regardless of where the change came from. Previewing on highlight and committing on selection, like Textual's own picker, would be a nice touch if you're up for it.

This also avoids the main bug in the current version: Textual 8.x registers its 21 built-in themes in `App.__init__()`, while `register_theme()` only replaces the four matching `catppuccin-*` names. As a result, `Change theme` offers roughly 22 themes instead of five.

Selecting something like `nord` only partially themes the UI: the chrome changes, while the bars and rows continue using the `current_theme_colors()` fallback for `cswap-dark`. Then this call from `watch_theme`:

```python
set_setting(backup_root, "tui.theme", "nord")
```

fails choice validation and is silently swallowed, so the selection does not survive a restart.

The menu picker prevents unsupported themes from being selected, but `watch_theme` should still explicitly guard against values not in `THEME_NAMES` rather than relying on the swallowed `ConfigError`.

A few smaller things:

* I'd drop the CLI accent coupling from this PR. A key named `tui.theme` changing the output of `cswap list` is a naming mismatch, and `_resolve_accent()` adds a `settings.json` read to every normal CLI invocation. Happy to discuss CLI theming separately.
* If it does stay, the truecolor escape is a portability regression. `\033[38;2;…` is not safe on legacy Windows consoles and some older terminals, while the previous `38;5;173` value was. It would need a `COLORTERM` check or a nearest-256-color fallback.
* Please remove the `Safe for the target terminal (kitty)` comment from `printer.py` and the `this rice's own plymouth-themes/…` reference from `palette.py`. They look unrelated to this project.
* `MenuItem` (`widgets.py`, around line 322) resolves its colors during construction, so menu labels retain the previous theme until the menu is rebuilt. The theme submenu will make this visible immediately, so these colors should also be resolved at render time like the other widgets.

Everything else looks good as-is.

realiti4:
Hi, thanks again for the PR! Theming support ended up landing via #151, which follows the direction we discussed here (a curated set as a menu setting), so I'm closing this one. If you have follow-up ideas a PR is always appreciated!

PR #126, opened 2026-07-12
feat(menubar): optional per-model weekly limits (e.g. Fable) in the title

asgeirtj:
The dropdown account rows gained scoped per-model windows (e.g. Fable) recently, which is great — but the menu-bar **title** still only renders `five_hour`/`seven_day`. Right now the number many of us are actually rationing is the Fable weekly bucket, and it's a click away.

This adds an opt-in `title_scoped` setting (default **off**, so existing titles are unchanged) with a toggle under **Settings → Show model limits in title**. When enabled, each scoped window is appended to the title as `<name> <pct>%`, after the 5h/7d segments, reusing `_rolled_weekly_window` so a passed weekly reset renders as 0% just like the dropdown rows.

Combined with `title_pct: off` you can make the title track only the model bucket:

```
⇄ asgeirtj6 · Fable 24%
```

or with `title_pct: both`:

```
⇄ asgeirtj6 · 24% · 15% · Fable 24%
```

- Settings round-trip via the existing `MenuBarSettings` load/save (unknown-key-tolerant, so older configs are unaffected)
- 3 new tests in `tests/test_menubar.py`; full suite passes
- Running it live on my own menubar since this afternoon

Happy to adjust naming (`title_scoped` vs `show_scoped_in_title`) or fold it into `TITLE_PCT_CHOICES` if you'd rather model it that way.

### DIFF
--- a/src/claude_swap/menubar.py
+++ b/src/claude_swap/menubar.py
@@ -45,6 +45,7 @@ class MenuBarSettings:
 
     show_account_name: bool = True
     title_pct: str = "both"  # one of TITLE_PCT_CHOICES
+    title_scoped: bool = False  # append per-model weekly limits (e.g. Fable) to the title
     refresh_interval: int = 60
     auto_switch_enabled: bool = False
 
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for the PR, it looks good, merged it!

Issue #125, opened 2026-07-12
feat: percent-remaining toggle + usage pace/reserve indicator


jslitzkerttcu:
## Summary

`pct` everywhere (menu bar, `cswap list`, JSON) is percent **used**, sourced straight from the usage API's `utilization`/`percent` fields (`oauth.py:394-449`). There's no way to see percent remaining, and no sense of whether current burn rate is sustainable until the next reset.

Two related but separable asks, inspired by how steipete/CodexBar handles this for its usage meters:

## 1. Used/remaining display toggle

CodexBar exposes a simple flip (`Sources/CodexBarCore/UsageFormatter.swift:91-97`, `usageLine(remaining:used:showUsed:)`): the same underlying number, just framed as `"62% used"` vs `"38% left"`.

For claude-swap this would be:
- A setting, e.g. `usage.display = "used" | "remaining"` (`settings.py`, same shape as the existing `autoswitch.*` keys).
- `menubar.py`'s `usage_summary`/`format_title` and `switcher.py`'s `list_accounts` line-building read the setting and render `100 - pct` with a `left` suffix instead of `pct` with a `used` suffix when set.
- Menu bar: a settings-menu item alongside the existing `Title percentage` picker.
- No new data needed — `pct` is already there; this is pure formatting.

## 2. Pace / reserve indicator

The more interesting piece. CodexBar's `Sources/CodexBar/UsagePace.swift` computes, for each window: `expectedUsedPercent = elapsed / windowDuration × 100`, compares it to `actualUsedPercent` (the real `pct`), and buckets the delta into a stage (`onTrack`, `slightlyAhead`, `ahead`, `farAhead`, and the `behind` mirror set), plus an ETA-to-exhaustion / "will last to reset" projection from the current burn rate.

Applied to claude-swap:
- Both window durations are fixed constants (5h = 18000s, weekly = 604800s), and every window already carries `resets_at` (`oauth.py:394-449`), so `elapsed = duration - (resets_at - now)` needs no new API data.
- A small `usage_pace.py` (mirrors the shape of `poll_policy.py`) with something like `compute_pace(pct, resets_at, duration_s, now) -> PaceResult(stage, delta_pct, expected_pct, eta_seconds)`.
[...]

PR #127, opened 2026-07-12
feat: usage.display setting for percent used vs. remaining (#125 part 1)

mliem2k:
## Problem

`pct` everywhere (`cswap list`, `cswap status`, the menu bar) is percent **used**. There's no way to see percent remaining, which is closes issue #125's first ask: a `used`/`remaining` display toggle (the second half of that issue, a pace/reserve indicator, is a separate follow-up not included here).

## What this does

Adds a `usage.display` setting (`"used"`, the default, or `"remaining"`):

```
cswap config set usage.display remaining
```

- `cswap list`/`cswap status`: every window (5h, 7d, spend, per-model scoped) reframes as `100 - pct` with a `left` suffix, e.g. `5h 62%` becomes `5h 38% left`. The `last seen` fallback note reframes the same way.
- Menu bar: title, dropdown rows, and the usage log line all reframe together. A new **Settings → Usage display** submenu mirrors the existing Title percentage picker.
- **JSON output is unchanged** — `pct` stays the raw utilization, so scripts can derive either framing themselves without a schema bump.
- The over-limit `(!)` marker still keys off the raw utilization regardless of display mode (a window is exhausted at `pct >= 100` either way).

## Implementation

- New `usage` section in `settings.json`, parallel to `autoswitch` (not a new `AutoSwitchSettings` field — it's a display preference, not engine policy). `SettingSpec.default` now resolves through a section → dataclass map so `cswap config get/set/unset/list usage.display` work through the existing generic plumbing without special-casing.
- `reframe_pct(pct, display)` in `settings.py` is the one place that does the `100 - pct` math and clamps at 0 (an over-limit window reads `0% left`, never negative).
- `_format_usage_lines`/`last_seen_note` (switcher.py) and `usage_summary`/`format_title`/`format_account_label`/`format_usage_log` (menubar.py) all take an optional `display` param defaulting to `"used"`, so every existing call site (and the TUI, untouched) keeps working byte-for-byte.

## Tests

- `test_settings.py`: `TestUsageSettings` (load/round-trip/garbage-degrades/section-isolation/invalid-choice/unset) + `TestReframePct` (identity, inversion, clamp, unknown-mode fallback).
- `test_switcher.py`: `TestUsageDisplayMode` (remaining-mode rendering, over-limit clamp, unknown-mode fallback, `last_seen_note` both ways).
- `test_menubar.py`: remaining-mode coverage for `usage_summary`, `format_account_label`, `format_title`, `format_usage_log`.
- `test_config_cli.py`/`test_settings.py`: updated the two settings-count assertions (8 → 9 keys).
- Full suite: 1226 passed, 3 skipped (up from 1205 passed on main).

### DIFF
--- a/src/claude_swap/menubar.py
+++ b/src/claude_swap/menubar.py
@@ -24,7 +24,10 @@ from datetime import datetime, timezone
 from pathlib import Path
 
 from claude_swap.exceptions import ClaudeSwitchError, CredentialReadError
+from claude_swap.settings import USAGE_DISPLAY_CHOICES, USAGE_DISPLAY_USED, reframe_pct
 from claude_swap.switcher import SENTINEL_NOTES
+
+USAGE_DISPLAY_LABELS = {USAGE_DISPLAY_USED: "Used", "remaining": "Remaining"}
 
 ICON = "⇄"
 REFRESH_CHOICES: tuple[int, ...] = (30, 60, 300)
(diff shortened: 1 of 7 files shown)


realiti4:
Thanks a lot for the PR! To be clear, #148 covers the other half of #125 (the pace indicator) — the used/remaining toggle itself is the part I decided to skip for now, see my reasoning on #125 (everything in the tool speaks percent used, so a "38% left" display next to a "switch at 90" config felt like a recipe for confusion).

Closing as not planned. Thanks again!

PR #128, opened 2026-07-12
feat: per-account aliases (#110)

### LINKED ISSUE #110: feat: per-account aliases (short names) in list, switch and menu bar
## Use case

I manage 3 Claude accounts on one machine, each with a distinct role (e.g. one for site development, two for content generation). They live on similar-looking emails, so telling them apart in `cswap list` and the menu bar means reading full addresses.

## Request

A user-defined short alias per account, e.g.:

```bash
cswap alias 1 dev
cswap alias 2 content-a
```

Shown wherever accounts are displayed, and accepted wherever accounts are addressed:

- `cswap list` / TUI / menu bar: `1: dev (claude@example.com) [CCT]`
- `cswap switch dev`, `cswap run content-a`, `cswap remove dev`

## Notes

- Slot numbers already work as short handles for switching, but they carry no meaning and reshuffle mentally when accounts are added/removed; a named handle is self-documenting.
- Alias is display/addressing sugar only — no change to stored credentials; could live in `settings.json` (or a sidecar map keyed by account email) in the backup root.
- Related to the display work in #102, but orthogonal: this is about naming, not detail rows.

Thanks for the tool — the auto-switcher saved my batch pipeline from mid-run rate limits on day one.

mliem2k:
Closes #110.

## What this adds

```
cswap alias 2 dev            # set (also: cswap alias user@example.com dev)
cswap alias dev prod         # rename (identifier can itself be an alias)
cswap alias 2 --unset        # remove
cswap alias                  # list current aliases
cswap add --alias dev        # convenience at add time
```

Aliases work anywhere an identifier does: `cswap switch dev`, `cswap remove dev`, `cswap run dev`.

## Following the spec worked out in the issue thread

Resolution precedence is **number → alias → email**. Validation: letters/digits/`-`/`_`/`.`, non-empty, normalized to lowercase, never purely numeric (reserved for slot numbers), unique across accounts.

The issue thread flagged three specific gaps in a first-draft design, all addressed here:

1. **`switch`/`remove` would reject aliases before resolution.** Both `switch_to` and `remove_account` validate a non-digit identifier with `_validate_email()` *before* `_resolve_account_identifier` is ever called, so `cswap switch dev` would fail with "Invalid email format". Both now check `_is_known_alias()` first and skip that gate for a recognized alias.
2. **The real display plumbing is the `accounts_info` 7-tuple (now, was 6), not `AccountInfo`.** Threaded through `_build_accounts_info` → `list_accounts`'s print loop (alias renders as `alias (email)` *after* the `  {num}: ` prefix the TUI watch-view parser depends on, never replacing it), `AccountSnapshot` (TUI account cards + remove menu), `account_row` in `json_output.py`, and the menu bar's `format_title`/`format_account_label`/dropdown/remove menu.
3. **Export/import and `add` re-run.** `transfer.py` carries `alias` through export and validates it on import (rejects a malformed alias, rejects a duplicate alias within one export file; a collision against an *already-present local* alias warns and drops just the alias rather than aborting the whole import, since alias is display sugar, not critical account data). Re-running `cswap add` on an already-managed account (the refresh-in-place branch) never touches the alias field at all; moving an aliased account to a new explicit `--slot` preserves it instead of silently wiping it.

## Storage

An optional `alias` field on each account record in `sequence.json`, alongside the existing `email`/`uuid`/`organizationUuid`/`organizationName`/`added`. Missing/`null` for every pre-existing account, so no migration is needed.

## Tests

45 new tests:
- `test_switcher.py` (`TestAccountAliases`, 20 tests): set/unset/list, number/alias/email resolution precedence, duplicate/numeric/invalid-character rejection, rename via an existing alias as the identifier, `switch_to`/`remove_account` no longer gate aliases on email format, `add_account` preserves alias across a re-slot and via `--alias`, refresh-in-place never touches it.
- `test_cli.py` (`TestAliasCommand` + `TestAddAliasFlag`, 15 tests): the full `cswap alias` command including usage-error argument combinations, root guard, main() dispatch, and an end-to-end `cswap switch dev` drive proving the identifier reaches `main()` without an "Invalid email format" error.
- `test_json_output.py`: `alias` present when set, absent (not null) when unset.
- `test_tui.py`: `account_card_text`/`mini_account_text` render `alias (email)`.
- `test_menubar.py`: `format_account_label`/`format_title` prefer alias; `_adapt_snapshot` carries it through.
- `test_transfer.py` (`TestAliasTransfer`, 4 tests): round-trip, invalid-alias rejection, same-export duplicate rejection, cross-machine collision drops the alias without aborting the import.

Full suite: 1250 passed, 3 skipped (up from 1205 passed on main).

### DIFF
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -12,6 +12,7 @@ from claude_swap.exceptions import ClaudeSwitchError
 from claude_swap.json_output import error_envelope
 from claude_swap.printer import (
     accent,
+    bolded,
     dimmed,
     error,
     force_utf8_output,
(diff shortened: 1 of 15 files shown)


jslitzkerttcu:
Hey @mliem2k, didn't see this land until after I'd already opened #129 for the same issue (I'm assigned to #110). Since we've both built the same feature independently, wanted to flag the overlap rather than let it sit as a silent duplicate, and pass along one thing from reviewing both implementations that's worth fixing regardless of which one lands:

Your alias validation regex (`^[a-z0-9_.-]+$` in models.py) allows a leading hyphen, e.g. "-dev". Once set, that alias can never be passed back into any command; argparse reads a leading-hyphen positional as an option flag, so `cswap switch -dev` would fail with "unrecognized arguments" instead of resolving. Worth rejecting at validation time.

@realiti4, happy to let you pick whichever implementation you'd prefer, or point to specific things from either one you'd like carried into the other. Not trying to make this a competition, just want to save you the trouble of discovering the duplicate/the hyphen bug yourself.

mliem2k:
Thanks for the catch, and sorry for the collision, @jslitzkerttcu. I opened this without noticing #110 already had an assignee, and you clearly had the head start here (the design in the issue thread is yours, worked out directly with @realiti4).

Confirmed the leading-hyphen bug: my `_ALIAS_RE` (`^[a-z0-9_.-]+$`) does allow `-dev` as a set-able alias, which then can never be passed back into any command since argparse reads it as a flag. Good catch. Since #129 also fixes that plus the `--unset`-with-no-target and self-import-collision edge cases, closing this in favor of that one rather than asking @realiti4 to pick between duplicates.

PR #129, opened 2026-07-13
feat: per-account aliases (short names)

### LINKED ISSUE #110: feat: per-account aliases (short names) in list, switch and menu bar

jslitzkerttcu:
## Summary

Closes #110. Implements the alias design we proposed in the issue, refined per your feedback (thanks for the detailed review!):

```
cswap alias 2 dev          # set (also: cswap alias user@example.com dev)
cswap alias dev prod       # rename (identifier can itself be an alias)
cswap alias 2 --unset      # remove
cswap alias                # list current aliases
cswap add --alias dev      # convenience at add time
```

Resolution precedence is number → alias → email, matching the spec. Validation: `^[a-z0-9._-]+$` after lowercasing, not purely digits, no leading hyphen (argparse would read it as a flag), unique across accounts.

## The three gaps you flagged

1. **`switch_to`/`remove_account` gates**: both previously ran `_validate_email()` on any non-digit identifier before `_resolve_account_identifier` ever saw it, so `cswap switch dev` failed with "Invalid email format". Fixed by checking for an alias match first; the interactive duplicate-email prompt is now skipped for a resolved alias (aliases are unique by construction, so it never needs disambiguation).

2. **Display plumbing**: went through the real hot paths you named instead of the dead `AccountInfo.display_label`. Widened `_build_accounts_info`'s tuple (6 → 7, alias appended) and threaded it through `list_accounts`, `accounts_snapshot` → `AccountSnapshot` (TUI cards and the remove-account menu), `account_row` (JSON), and `_build_status_payload`. Menu bar's `format_title`/`format_account_label` prefer the alias over the email-derived label when set. Every surface leads with the alias ("alias (email)") since the whole point is not having to read full addresses to tell similar-looking accounts apart.
   - Also found the `list_accounts` print-loop comment claiming a `tui._watch_account_rows` parser depends on the exact output format was stale; that function no longer exists anywhere in the repo (it predates the Textual TUI rewrite). Removed the comment since there's no actual constraint left.

3. **Export/import + `add_account` alias preservation**: `alias` is carried through the export entry builder, import validation/normalization, and the `new_record` write. `add_account`'s explicit-slot full-record-write branch reads the existing record first and carries its alias forward instead of dropping it on refresh/migrate, and the refresh-in-place path applies a passed `--alias` immediately rather than ignoring it.
   - Import policy: an incoming alias that collides with an existing local alias is dropped with a warning rather than failing the whole import (aliases are a display nicety, not identity).

## Architecture

- `normalize_alias()` in `models.py` is the single source of truth for validation, shared by the CLI, `add_account`, and import validation.
- `set_alias()`/`unset_alias()` are dedicated methods with a clear API; `list_aliases()` returns structured `(num, alias, email)` tuples so the CLI (or any future caller) formats its own output rather than the switcher printing directly.

## Testing

- Full suite: 1205 → 1261 tests, all green.
- Coverage includes: alias validation, resolution precedence (including rename-via-alias), the gate-fix regressions, `add --alias` on fresh/refresh/migrate paths, export/import round-trip + collision handling (including a self-import edge case), JSON payload fields, menu bar and TUI formatting (including the remove-account menu), and a handful of edge-case fixes an independent review pass surfaced (leading-hyphen rejection, `--unset` with no target, and an empty-string identifier that could otherwise silently match an arbitrary account).
- Manually smoke-tested the CLI end-to-end.

## Test plan

- [x] `pytest tests/`: 1261 passed, 3 skipped
- [x] Manual: `cswap alias 1 dev`, `cswap alias`, `cswap list` (alias shown, leading), `cswap alias 2 dev` (duplicate, errors), `cswap alias dev prod` (rename via alias), `cswap remove dev`, `cswap alias 1 --unset`, `cswap alias --unset` (no target, errors), `cswap alias 1 -dev` (leading hyphen, errors)

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ Or switch to a specific account:
 ```bash
 cswap switch 2
 cswap switch user@example.com
+cswap switch dev                # or by alias, once set with `cswap alias 2 dev`
 ```
 
 Not sure which one? `cswap list` is the dashboard — every account's 5-hour and 7-day usage and reset times at a glance:
@@ -168,7 +169,11 @@ cswap config                    # Show or edit settings (see Configuration below
 cswap list                      # Show all accounts with 5h/7d usage and reset times
 cswap status                    # Show current account
 cswap add --slot 3              # Add account to a specific slot (prompts before overwrite)
+cswap add --alias dev           # Add account and give it a short alias
 cswap remove 2                  # Remove an account
+cswap alias 2 dev               # Give an account a short alias (usable anywhere NUM|EMAIL is)
+cswap alias 2 --unset           # Remove an account's alias
+cswap alias                     # List all aliases
 cswap tui                       # Interactive dashboard (also: bare `cswap`)
 cswap watch                     # Dashboard, opened on the live watch page
 cswap upgrade                   # Upgrade claude-swap to the latest version
@@ -286,6 +291,8 @@ cswap switch 2 --json
 Every payload carries a `schemaVersion` (currently `1`); on a handled error stdout is `{"schemaVersion":1,"error":{...}}` with a non-zero exit code. `--switch`/`--switch-to` report `{"switched": true|false, "from": …, "to": …, "reason": …}`.
 
 Usage is served from a per-account cache: when the usage API is briefly unreachable, the last-known numbers are shown instead of nothing (the human view marks them with their age, e.g. `· 2m ago`). Rows with usage carry additive `usageFetchedAt`/`usageAgeSeconds` fields telling you how old the measurement is.
+
+An account row also carries an additive `alias` field once one is set with `cswap alias` (e.g. `"alias": "dev"`); accounts without one simply omit the key.
 
 </details>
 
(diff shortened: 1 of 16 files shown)


realiti4:
Looking really good! Thanks for the PR, I've merged it.

PR #130, opened 2026-07-13
Add `disable`/`enable` to hold an account out of auto-rotation

kmizzi:
## What

Adds two commands to park a managed account out of automatic rotation without removing it (and losing its stored login):

```bash
cswap disable <num|email>   # hold out of auto-rotation
cswap enable  <num|email>   # put it back
```

The legacy flag spellings `--disable-account` / `--enable-account` work too, matching the rest of the CLI.

## Why

There's currently no way to say "keep this account managed, but don't auto-switch onto it right now" — a work account you don't want touched, one you're resting, or a temporarily-flaky login. The only options are `remove` (destroys the stored login) or nothing. This adds a reversible toggle.

## Behavior

A disabled account is skipped by:
- the **auto-switch engine** (every candidate list routes through `switchable_account_numbers()`),
- bare **`cswap switch`** rotation, and
- the **`best`** and **`next-available`** strategies.

It stays fully managed and remains a valid **explicit** `cswap switch <num|email>` target — disabling only affects *automatic* selection, so you can still deliberately jump to a parked account. The fresh-machine bootstrap path (activating a login right after `--import`) is intentionally untouched.

`cswap list` shows a `(disabled)` marker; `--list --json` carries an additive `"disabled": true` on the row (absent otherwise, so existing consumers are unaffected).

## Implementation

- State is a per-account `disabled: true` flag in `sequence.json`. It's preserved across the org-field migration (which mutates records in place) and is naturally cleared when a slot is removed and re-added.
- New switcher API: `set_account_disabled()`, `is_account_disabled()`, `disabled_account_numbers()`, and a `_disabled_from_data()` helper. `switchable_account_numbers()`, `_select_best_switchable()`, and the `switch()` rotation loop gained the `and not disabled` predicate (rotation prints a `Skipping Account-N (disabled)` line / JSON warning, mirroring the existing broken-slot skip).
- Resolution reuses `resolve_account()` (NUM or EMAIL; hard error on ambiguous email or unknown account). Disabling the active account or the last remaining rotation target prints a heads-up but never blocks.

## Tests

21 new tests: flag storage + accessors, rotation/`best` exclusion, explicit `switch <num>` still activating a disabled slot, re-add clearing the flag, idempotent toggles, active-account / last-account warnings, the `(disabled)` list marker, the JSON field, and CLI dispatch for both the subcommands and legacy flags. Full suite green (`1226 passed, 3 skipped`).

## Follow-up (not in this PR)

The TUI/menu-bar automatically *respect* disabled state (their switch actions call the same `switch()` paths), but don't yet surface a `(disabled)` marker or an in-app toggle — a natural next step.

https://claude.ai/code/session_01XCcSh6s4UiLHrwKLd1UNaX

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,7 @@ cswap auto --dry-run           # log what it would do, never switch
 - Usage polling is adaptive — a couple of accounts per check, busy alternates watched more closely, exhausted ones left alone until they reset — so API traffic stays flat no matter how many accounts you manage.
 - It fails safe: if a usage check errors it keeps trusting the last-known numbers while retries back off, and an expired token on an idle machine makes it hold rather than fail over (Claude Code refreshes the token on your next message).
 - An account whose refresh token has died is quarantined and reported until you log in with it and re-run `cswap add --slot N`. API-key accounts are never rotated onto unless you pass `--include-api-key-accounts`.
+- To hold an account out of rotation yourself — a work account you don't want touched, one you're resting — run `cswap disable <num|email>`; `cswap enable <num|email>` puts it back. Disabled accounts are skipped by auto-switch, bare `cswap switch`, and the `best` / `next-available` strategies, but stay fully managed and remain a valid explicit `cswap switch <num|email>` target. They show a `(disabled)` marker in `cswap list`.
 - By default only the account-wide 5h/7d windows drive switching. If you work on one model and hit its **weekly per-model limit** first (e.g. Fable), add `--model Fable` (or `cswap config set autoswitch.model Fable`) to fold that model's window into the decision, so it switches off an account whose model quota is spent even while its 5h/7d windows still have room.
   - **Model names** are Anthropic's own per-model `display_name`s, matched case-insensitively. The exact strings for your accounts are the per-model rows in `cswap list` (e.g. a line reading `Fable: 100%`).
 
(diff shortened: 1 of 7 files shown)


realiti4:
Hi, thanks for the PR. I think this is a great idea and would be very useful.

It looks good overall. However, the alias feature PR landed just before this one. Could you also add support for aliases and resolve any conflicts? Once that is done, we should be able to merge it. Thanks!

kmizzi:
Thanks! Done — rebased onto the current `main` (with the alias feature) and added alias support:

- **Alias resolution comes for free.** `set_account_disabled` resolves its target through `resolve_account` → `_resolve_account_identifier`, so `cswap disable dev` / `cswap enable dev` (or any alias) work the same as a number or email — no separate resolution path.
- **Conflicts resolved in the list output.** `list_accounts` and `_build_list_payload` now carry the alias `label` and the `(disabled)` marker together, and `account_row` keeps both the new `alias` and my additive `disabled` JSON field.
- **New test** `test_disable_and_enable_by_alias` proves the alias path (aliases an account, disables/enables it by alias, asserts rotation membership).

Full suite green: **1283 passed, 3 skipped**. Ready for another look.

realiti4:
Thanks for the adjustments! I've merged it.

PR #133, opened 2026-07-13
feat: surface and toggle disabled state in the TUI and menu bar

kmizzi:
Follow-up to #130. That PR added `cswap disable` / `enable` to hold accounts out of auto-rotation, and the interactive surfaces already **respected** the flag (their switch actions call the same `switch()` paths) — but they didn't yet **surface** a `(disabled)` marker or offer an in-app toggle. As the #130 description put it:

> The TUI/menu-bar automatically respect disabled state … but don't yet surface a `(disabled)` marker or an in-app toggle — a natural next step.

This is that next step.

## What changed

- **Snapshot** — `AccountSnapshot.disabled`, populated by `accounts_snapshot` from the sequence store, so every interactive view reads the flag in the same coherent one-pass view as usage and active-detection (no second read that could disagree).
- **TUI** — a `(disabled)` marker on the active account card and on the inactive one-line rows, plus a **Disable / enable account…** submenu. Selecting an account toggles it in place: the action reads the live state to pick the direction, and there's no confirm modal since it's fully reversible.
- **Menu bar (macOS)** — the same `(disabled)` marker on each account row, and a **Disable / enable account** submenu whose rows are check-marked when disabled.
- Both toggles route through the existing `set_account_disabled`, i.e. the same path the CLI uses — so number/alias/email resolution and the JSON store stay identical across surfaces.

## Behavior notes

- Disabled accounts remain fully managed and stay a valid **explicit** switch target from these menus (matching the CLI: only *auto*-rotation skips them). The marker is informational; the toggle is the only new mutation.
- No new keybindings — the toggle lives in the menu, consistent with the dashboard's "actions live in the nested menu" design.

## Tests

- TUI: `(disabled)` marker on card + mini rows; submenu toggle disables without a modal and pops back to root; a disabled row offers *enable* and flips back on.
- Menu bar: `format_account_label` disabled marker; `_adapt_snapshot` tuple carries `disabled` through.
- Switcher: disable/enable by alias resolves like number/email.

Full suite green (`uv run pytest`): 1287 passed, 3 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ cswap auto --dry-run           # log what it would do, never switch
 - Usage polling is adaptive — a couple of accounts per check, busy alternates watched more closely, exhausted ones left alone until they reset — so API traffic stays flat no matter how many accounts you manage.
 - It fails safe: if a usage check errors it keeps trusting the last-known numbers while retries back off, and an expired token on an idle machine makes it hold rather than fail over (Claude Code refreshes the token on your next message).
 - An account whose refresh token has died is quarantined and reported until you log in with it and re-run `cswap add --slot N`. API-key accounts are never rotated onto unless you pass `--include-api-key-accounts`.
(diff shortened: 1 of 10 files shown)


realiti4:
Thank you for the follow-up, looks good. I've merged it.

PR #134, opened 2026-07-13
fix: skip disabled accounts after fresh import

ShiroKSH:
## Summary

- Skip disabled slots when a fresh imported profile uses bare `cswap switch`.
- Return an actionable error when no enabled, switchable account remains.

## Root cause

The fresh-machine path checked backup integrity but not disabled state, unlike normal rotation.

## Validation

- `uv run pytest -q`


realiti4:
Thanks for the fix! Looks good, I've merged it.

PR #142, opened 2026-07-15
feat: cswap move & swap — assign an account to any slot number

PavloPopravkin:
## What

A small `cswap swap <a> <b>` command that exchanges two accounts' slot numbers, so they trade places in `cswap list` and as numeric targets. Targets accept the usual NUM | EMAIL | ALIAS forms.

```
$ cswap swap 1 2
Swapped Account 1 and Account 2:
  1: work@example.com
  2: personal@example.com
```

## Why

Slot numbers become muscle memory (`cswap switch 1`, scripts, mappings read in list order), but they're assigned by add-order. After adding accounts in the "wrong" order the only way to reorder today is remove + re-add, which costs the stored credentials.

## How

Everything keyed by the slot number moves with the swap:

- sequence records themselves (aliases live inside the record, so they travel with their account)
- per-slot credential and config backups (read both slots up front, write under the new keys, delete the old keys — skipped when both accounts share an email, where old and new keys fully overlap)
- rotation order in `sequence` and `activeAccountNumber`
- each slot's session profile directory, so `cswap run` history survives (best-effort move staged through a temporary name for the shared-email case; a skipped move only costs that slot's session history, since setup re-bootstraps from the relocated backups)

Not touched, by design:

- directory mappings — they key on (email, org) precisely to survive renumbering
- usage-cache rows — they key on the slot number but carry the account identity, so a swapped row fails the identity check and self-heals on the next poll

Live session-mode instances block the swap, same as `--remove-account`.

## Tests

`tests/test_swap_accounts.py`: swap by number/email/alias, active-number follows the account, rotation order updated, backups relocated (including a one-sided/never-backed-up slot), session profile moved, same-account and unknown-identifier rejected. Full suite: 1324 passed, 3 skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ cswap enable 2                  # Return a disabled account to rotation
 cswap alias 2 dev               # Give an account a short alias (usable anywhere NUM|EMAIL is)
 cswap alias 2 --unset           # Remove an account's alias
 cswap alias                     # List all aliases
+cswap swap 1 2                  # Exchange two accounts' slot numbers (list order)
 cswap tui                       # Interactive dashboard (also: bare `cswap`)
 cswap watch                     # Dashboard, opened on the live watch page
 cswap upgrade                   # Upgrade claude-swap to the latest version
--- a/README.md
+++ b/README.md
@@ -187,6 +187,7 @@ cswap alias 2 dev               # Give an account a short alias (usable anywhere
 cswap alias 2 --unset           # Remove an account's alias
 cswap alias                     # List all aliases
 cswap swap 1 2                  # Exchange two accounts' slot numbers (list order)
+cswap move 2 1                  # Assign an account to a slot (relocates to an empty slot, swaps if taken)
 cswap tui                       # Interactive dashboard (also: bare `cswap`)
 cswap watch                     # Dashboard, opened on the live watch page
 cswap upgrade                   # Upgrade claude-swap to the latest version
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -333,6 +333,51 @@ def _unmap_command(argv: list[str]) -> None:
         sys.exit(130)
 
 
+def _swap_command(argv: list[str]) -> None:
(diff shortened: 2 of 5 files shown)


realiti4:
Hi, thanks for the PR!

Great idea, but I think we shouldn't restrict this to just swapping. Your muscle memory comment made me realize this would be better as an "assign to a shortcut" kind of behavior.

I'm thinking something like this:

```
cswap move work@example.com 1
cswap move dev 1
cswap move 2 1
```

So it works with email, alias, or number, and we'd be able to move an account to any slot with it including an empty one, which plain swap can't do. If the target slot is occupied, we show a notice and swap them (the displaced account takes the vacated slot, so nothing is ever lost).

Nice side effect: with that rule, swap a b is exactly move a <b's slot>, so your implementation should carry over almost unchanged. You could even keep swap as a two-way spelling of move if you like.

What do you say? If this sounds good to you as well, I'd happily accept follow-up commits on this PR.

PavloPopravkin:
Love it, that framing is clearly better. Assigning an account to a slot is the real operation; swapping is just the special case where the slot is already taken. And it covers the empty-slot case plain swap can't: `remove` leaves sparse slots, so "move to an unused number" is a genuine thing.

Pushed a follow-up commit:

- `cswap move <NUM|EMAIL|ALIAS> <slot>` — target is a slot number. Empty target relocates the account and frees its old slot; occupied target swaps the two (the displaced account takes the vacated slot, so nothing is lost); same slot is a no-op.
- The occupied path just calls the existing `swap_accounts`, so that logic carried over unchanged and its tests stay green.
- Kept `swap a b` as the two-way spelling — it's exactly `move a <b's slot>`, and there's a test asserting the two land identical state.
- New `tests/test_move_accounts.py` covers relocation, active/rotation/backup/session-profile follow, swap-equivalence, no-op, and bad-target validation. Full suite green.

I made the target a slot *number* only (not an identifier) so the empty-slot case stays unambiguous. Happy to also accept an account identifier there (so `move a b` == `swap a b`) if you'd prefer — easy to add.

realiti4:
Merged thank you.

PR #143, opened 2026-07-16
feat: per-window switch triggers + soonest-reset strategy

jmilrc:
## Summary

Two additive, backward-compatible `autoswitch` options:

**1. Per-window OR'd switch triggers** — new `threshold5h` / `threshold7d` settings (defaults 95 / 98). Switch when the **5h** window **or** the **7d** window crosses its *own* threshold, rather than a single `threshold` on the binding (max) window. The 7d window is a week-long limit, so it's useful to let it sit closer to the wall than the self-refilling 5h window. Landing eligibility, the `--threshold` flag, and the TUI slider all honor the per-window model; the existing `threshold` still governs folded per-model windows and poll-cadence escalation.

**2. `strategy = "soonest-reset"`** — among eligible target accounts, switch to the one whose **7-day window resets soonest** (burn the most-perishable weekly budget first), instead of `"best"` (most headroom). `"best"` remains the default, so existing behavior is unchanged unless opted in.

## Why

With several accounts, "switch to the account with the most quota left" isn't always what you want — it preferentially spends the budget that's *furthest* from resetting. `soonest-reset` spends the most-perishable weekly budget first. And a single binding-window threshold can't express "ride the 5h window to X% but switch the precious 7d window earlier."

## Compatibility

- Fully additive. `"best"` stays the default; unset per-window keys inherit sensible defaults.
- Downgrade-safe: the settings loader already ignores unknown keys and degrades an unknown `strategy` value to the default with a logged warning, so reverting to an older version is clean.

## Tests

Adds unit tests (`test_poll_policy.py`: per-window trigger/landing helpers, soonest-reset ordering incl. missing/past reset timestamps) and engine-level tests (`test_autoswitch.py`: per-window trigger asymmetry, `soonest-reset` vs `best` selection). Full suite passes.

### DIFF
--- a/src/claude_swap/autoswitch.py
+++ b/src/claude_swap/autoswitch.py
@@ -48,7 +48,10 @@ from claude_swap.locking import FileLock
 from claude_swap.poll_policy import (
     ESCALATION_MARGIN_PCT,
     RESET_SLACK_S,
+    account_landing_ok,
+    account_triggered,
     binding_pct,
+    soonest_7d_reset_ts,
 )
 from claude_swap.settings import AutoSwitchSettings, atomic_write_json, parse_model_names
 from claude_swap.switcher import ClaudeAccountSwitcher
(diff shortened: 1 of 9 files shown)


jmilrc:
Closing — opened in error, not intended for submission.

PR #145, opened 2026-07-16
fix: preserve MCP OAuth sessions across account switches

### LINKED ISSUE #135: Account switch clobbers live MCP OAuth tokens (mcpOAuth) with the target slot's stale snapshot
### Summary

On macOS, Claude Code stores MCP server OAuth tokens (`mcpOAuth`) in the **same** keychain item as the account credential (`Claude Code-credentials`), alongside `claudeAiOauth`. When cswap switches accounts, it writes the target slot's stored snapshot over the live item verbatim, which replaces the current `mcpOAuth` tokens with whatever was in the blob the last time that account was backed up. If any MCP token rotated in the meantime, the switch regresses it to an older (often already rotated-out) generation, and the affected MCP servers demand re-authentication.

### Environment

- claude-swap 0.20.0 (uv tool install), macOS (Darwin 25.5.0), keychain backend
- Two managed accounts, switching via `cswap --switch-to <slot>`

### Mechanism (v0.20.0 source)

- `switcher.py:3690` (Step 3 of `_perform_switch`): `self._write_credentials(target_creds)` writes the stored backup blob wholesale. Only `oauthAccount` in `~/.claude.json` gets a targeted splice; the keychain blob does not.
- There are no references to `mcpOAuth` anywhere in the package — the credential is treated as an opaque per-account blob, but Claude Code overloads that item as a shared token store: `claudeAiOauth` is per-account, `mcpOAuth` is account-independent (tokens for third-party MCP servers like Todoist, Cloudflare, GitHub, etc.).
- The refresh path (`oauth.py`, `try_refresh_oauth_credentials`) correctly preserves sibling keys — but it only runs for `claudeAiOauth`, never at switch time for `mcpOAuth`.

### Evidence from a live system

Live item vs stored slot backups (SHA-256 of the sorted `.mcpOAuth` object, secrets never printed):

```
slot 1 backup: 58 mcp servers, mcpOAuth hash 3606bde16ad7
slot 2 backup: 58 mcp servers, mcpOAuth hash a02f21e2d4e0   <- diverged
live:          58 mcp servers, mcpOAuth hash 3606bde16ad7
```

Per-server comparison of slot 2's backup vs live shows the regression concretely — the Cloudflare MCP token in the backup is an older generation than the live one:

```
cloudflare: slot2-backup expiresAt=1783989344998, live expiresAt=1783989606433 (token blob DIFFERS)
```

Switching to slot 2 without any mitigation writes the older token back. When the MCP server has rotated the refresh token server-side, that older generation is dead, and the user has to re-run OAuth for each affected MCP server. With ~58 MCP servers in the blob, this is a recurring papercut after most switches. The failure is intermittent by nature (depends on whether a given server rotated tokens since the backup was taken), which made it annoying to trace.

### Suggested fix

Before Step 3, splice the live blob's account-independent keys into the target credential:

```python
live = json.loads(original_creds)
target = json.loads(target_creds)
if "mcpOAuth" in live:
    target["mcpOAuth"] = live["mcpOAuth"]
target_creds = json.dumps(target)
```

(Conceptually: swap `claudeAiOauth`, carry `mcpOAuth` live — mirroring what the config path already does by splicing only `oauthAccount` into `~/.claude.json` instead of replacing the file.)

A conservative variant is a key allowlist/denylist so future account-independent keys Claude Code adds to the item are carried too, with `claudeAiOauth` being the only key taken from the stored backup.

### Workaround

I currently wrap `cswap` in a shell function that snapshots `.mcpOAuth` from the live item before the switch and merges it back afterwards. It works, but it can't cover the menubar app, `cswap auto`, or TUI-initiated switches — which is why I think this belongs in cswap itself.

Happy to test a patch on a real two-account macOS setup.

pencount:
## Problem

Switching Claude accounts replaces the entire active credential object with the destination slot's saved copy. Claude Code stores both the account login and machine-shared OAuth integrations (notably `mcpOAuth`, the MCP server logins) in that one object.

If an MCP refresh token rotates after the destination slot was backed up, activating that slot restores the older token generation, and the affected MCP servers demand re-authentication.

## Root cause

The credential object contains state with two different owners and lifetimes:

- `claudeAiOauth` is owned by an account slot
- sibling fields, currently including `mcpOAuth`, are shared by the machine and may rotate independently

A destination slot can authoritatively supply its `claudeAiOauth`, but its sibling fields are frozen at backup time. The live credential's siblings are by definition the current generation.

## Change

One commit, ~85 production lines: compose every activation from the two owners instead of copying the stored blob verbatim.

- the destination slot supplies only its `claudeAiOauth` login
- the sibling fields come from the live credential (live wins over the slot's stale snapshot)
- when there is no live JSON credential object to take siblings from (fresh machine, or a managed API key is active), the stored blob activates unchanged — exactly the pre-existing behavior
- managed API keys and opaque legacy credential shapes are never composed; they activate byte-for-byte as before

Both activation paths (normal switch and direct activation) go through the same composer. Account backups, exports, imports, and session profiles are untouched.

## Out of scope (unchanged behavior)

Preserving MCP state across an **OAuth → managed API key → OAuth round trip** needs an account-independent snapshot store, since there is no live OAuth object to take siblings from while the API key is active. That is a larger change (new storage identity, its own failure modes); a complete, tested implementation exists on [`feat/shared-oauth-store`](https://github.com/pencount/claude-swap/tree/feat/shared-oauth-store) and I'm happy to follow up with it if you want that case covered.

## Validation

- unit: live siblings win over stale target siblings; API-key live state, absent live state, API-key targets, and opaque legacy shapes all activate verbatim
- end-to-end: a normal switch and a direct activation both preserve the live `mcpOAuth` over the slot's stale copy
- `uv run pytest -q` — 1,355 passed, 3 skipped
- `git diff --check`

Closes #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,9 @@ The original flag spellings (`cswap --switch`, `cswap --list`, ...) keep working
 ## How it works
 
 - Backs up OAuth tokens and config when you add an account
-- Swaps credentials when you switch accounts
+- Swaps only the account-specific Claude login when you switch accounts;
+  live account-independent OAuth state (such as MCP server logins) is
+  preserved instead of being overwritten by a slot's older snapshot
 - Account credentials stored securely using platform-appropriate methods
 - Switches (manual and automatic) hold Claude Code's own credential locks while writing, so a swap never interleaves with a token refresh
 - Auto-switch freshens a target's token before activating it, and quarantines accounts whose refresh token has died (recover by re-adding it with `cswap add --slot N`, or by replacing its stored credentials from a known-good export with `cswap import backup.cswap --force`)
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -99,6 +99,54 @@ def looks_like_api_key(credentials: str | None) -> bool:
     return text.startswith("sk-ant-api") and not text.startswith("{")
 
 
+def _credential_object(credentials: str | None) -> dict | None:
+    """Parse a JSON credential object, excluding managed API keys."""
+    if not credentials or looks_like_api_key(credentials):
+        return None
+    try:
+        data = json.loads(credentials)
+    except (json.JSONDecodeError, TypeError):
+        return None
+    return data if isinstance(data, dict) else None
+
+
+def shared_credential_fields(credentials: str | None) -> dict | None:
+    """Return account-independent fields from a Claude OAuth credential object.
+
+    Claude Code keeps machine-shared OAuth integrations (notably ``mcpOAuth``)
+    as siblings of the slot-owned ``claudeAiOauth`` field in one object.
+    ``None`` means the input is not a JSON credential object (missing,
+    malformed, or a managed API key). A dictionary — including ``{}`` — is
+    authoritative: every sibling of ``claudeAiOauth`` is machine-shared.
+    """
+    data = _credential_object(credentials)
+    if data is None:
+        return None
+    return {
+        key: value
+        for key, value in data.items()
+        if key != "claudeAiOauth"
+    }
+
+
+def merge_shared_credential_fields(
+    target_credentials: str, shared_fields: dict
+) -> str:
+    """Compose a target Claude login with independently-owned shared fields.
+
+    Returns ``target_credentials`` unchanged when it is not a JSON credential
+    object carrying a Claude login (managed API keys and opaque legacy shapes
+    stay activatable verbatim).
+    """
+    target = _credential_object(target_credentials)
+    if target is None or "claudeAiOauth" not in target:
+        return target_credentials
+
+    composed = dict(shared_fields)
+    composed["claudeAiOauth"] = target["claudeAiOauth"]
+    return json.dumps(composed)
+
+
 def approved_form(api_key: str) -> str:
     """The value Claude Code stores in ``customApiKeyResponses.approved``.
 
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -110,40 +110,62 @@ def _credential_object(credentials: str | None) -> dict | None:
     return data if isinstance(data, dict) else None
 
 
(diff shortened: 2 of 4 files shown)


realiti4 (changes requested):
Thanks for this the implementation is polished, and the #144 → #145 slim-down was the right call. Compose-at-activation is the right overall shape.

There is one ownership case we need to address before merge: not every sibling is machine-shared.

The credential item can also contain `trustedDeviceToken`, which backs Claude Code's remote-control sessions and is sent as the `X-Trusted-Device-Token` header. Unlike the MCP tokens, it is account-bound: Claude Code enrolls it against the logged-in account after `/login` and deliberately re-enrolls on every login so a previous account's token is not sent with the new account. Since `cswap` switches do not trigger re-enrollment, live-siblings-win would carry account A's device token into account B.

Could we invert the composer to use an explicit allowlist of the known shared keys `mcpOAuth`, `mcpOAuthClientConfig`, `mcpXaaIdp`, `mcpXaaIdpConfig`, and `pluginSecrets` while leaving unknown fields target-owned? An unknown shared field defaulting to target-owned risks a stale restore; an unknown account-scoped field defaulting to live-owned risks crossing account boundaries. The allowlist therefore fails safer.

A test asserting `trustedDeviceToken` and an unknown sibling are taken from the target would pin that contract down.

**Non-blocking edge case:** In direct activation, the composer only receives live credentials when `~/.claude.json` contains an identity. A live credential object without a config identity is therefore still replaced verbatim. Reading active credentials unconditionally in that path would close the gap, either here or in a follow-up.

For the API-key round trip, my preference is to defer `feat/shared-oauth-store` until there is real demand the additional storage identity introduces meaningful failure surface, and that implementation would need the same allowlist treatment.

With the allowlist change in place, I think #145 is ready to merge.

realiti4:
Thanks for the adjustments again, looks good I've merged it.

PR #148, opened 2026-07-19
feat: weekly usage pace indicator (closes #125)

### LINKED ISSUE #125: feat: percent-remaining toggle + usage pace/reserve indicator

jslitzkerttcu:
## Summary

Scoped to the pace indicator only. #125 had two separable asks (a used/remaining display toggle and the pace indicator), and I've left the toggle out here, matching the "skip it for now, revisit if more people ask" call in the thread (a `38% left` display next to a `switch at 90` threshold felt like a small permanent source of confusion for a cosmetic gain, and every usage surface would need to stay consistent with it).

The pace indicator itself follows the implementation notes from the thread:

- `compute_pace()` is a pure consumer of the stored snapshot: no fetching, no poll-cadence influence, and it computes elapsed against the snapshot's `fetched_at` rather than `now()`.
- Applies only to weekly windows, never 5h (that window is "ahead of pace" almost by definition early on, and poll staleness is negligible against a week but not against 5h).
- A marker shows only when meaningfully ahead (≥15pp over the expected-so-far percentage); on-track/behind stay silent to avoid noise in already-dense rows.
- Stage/delta stays in the UI; ETA numbers (`projectedExhaustionAt`, plus a `willLastToReset` boolean) are `--json`-only, since linear ETA has wide error bars and would look falsely precise in the UI.
- The first ~24h after a weekly reset is suppressed, since `elapsed == 0` alone isn't a sufficient guard against the near-zero-elapsed false positive.
- Workday filtering is left for v2.

## Implementation

- New `pace.py`: `compute_pace()`, `projected_exhaustion_ts()`, `will_last_to_reset()`, applies only to `seven_day` plus every `scoped` per-model window, never `five_hour`.
- Wired into every surface that renders weekly usage: `cswap list` / the TUI's shared row formatter, the TUI's full account card and minimized row, and the menu bar dropdown.
- `compute_pace()` also needed to stay correct when `resets_at` is several cycles stale (our poll-and-serve-last-good architecture can hand it genuinely old cached data). It derives the current window's start via modular arithmetic on `resets_at`, so it's correct no matter how many cycles behind the cached window is.

## Test plan

- [x] New `tests/test_pace.py`: pure fixed-clock tests for `compute_pace`/`projected_exhaustion_ts`/`will_last_to_reset` (elapsed math, suppression window, ahead threshold, stale/rolled resets_at handling)
- [x] Marker-injection tests added to `TestFormatUsageLines` (switcher), `test_menubar.py`, `test_tui.py`, `test_json_output.py`, including negative cases (no marker without `fetched_at`, no marker in the suppression window, no marker on 5h, `(!)` maxed marker wins over pace)
- [x] Full suite green on macOS/Linux/Windows (1406 passed; the only prior failures were 6 pre-existing, unrelated ones already present on `main` in `test_swap_accounts.py`/`test_move_accounts.py`)
- [x] Manually verified against my real accounts via `cswap list` / `cswap list --json` (no crashes, correctly showed nothing since my accounts are either behind pace or within the reset-suppression window), plus synthetic data through the actual rendering functions to confirm the marker text across all three surfaces
- [x] README's `--json` schema section updated to document the new additive fields

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,8 @@ Usage is served from a per-account cache: when the usage API is briefly unreacha
 
 An account row also carries an additive `alias` field once one is set with `cswap alias` (e.g. `"alias": "dev"`); accounts without one simply omit the key.
 
+Weekly windows (`sevenDay` and any per-model `scoped` entry — never `fiveHour`) additively carry `expectedPct`/`aheadOfPace`/`projectedExhaustionAt`/`willLastToReset` once enough of the week has elapsed to make pace meaningful: `expectedPct` is the "on schedule" usage for how far into the week you are, `aheadOfPace` is `true` only when meaningfully over that (the same signal the human views show as a `(pace)`/`(ahead of pace)` marker), and `projectedExhaustionAt`/`willLastToReset` are a linear-projection ETA and its yes/no summary for whether usage stays under 100% before the next reset — `--json`-only, since the projection has wide error bars and would look falsely precise in `cswap list`/the TUI/the menu bar.
+
 </details>
 
 `cswap auto --json` emits an event *stream* instead — one JSON object per line (`{"schemaVersion":1,"event":"switch","ts":…, …}` with kinds like `poll`, `switch`, `no-switch`, `account-quarantined`, `all-exhausted`, `error`). The contract is additive: new kinds and fields may appear, so scripts should ignore unknown ones.
--- a/src/claude_swap/json_output.py
+++ b/src/claude_swap/json_output.py
@@ -10,7 +10,7 @@ from __future__ import annotations
 
 from datetime import datetime, timezone
 
-from claude_swap import oauth
+from claude_swap import oauth, pace
 
 # Bump only on a breaking change to any payload shape. Scripts key off this.
 SCHEMA_VERSION = 1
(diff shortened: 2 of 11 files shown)


realiti4:
Thanks for this, it looks good.

Could we adjust a couple of small things before merge?

1. **(pace) → (ahead)** in the compact surfaces (menubar rows, TUI mini). (pace) can read as "on pace", which is exactly the state that deliberately shows nothing — (ahead) is one character more and unambiguous. The long-form (ahead of pace) surfaces are fine as they are.

2. **active_usage_fetched_at** in the menubar snapshot ends up unread — format_title doesn't render pace, so only the per-account tuple entries are consumed. Unless you have a title marker in mind as a follow-up, I'd drop the key (and its _STATE/_adapt_snapshot/test bits) to keep the snapshot lean.

Thanks again.

jslitzkerttcu:
Done on both, thanks for catching these:

1. Swapped (pace) to (ahead) in the menubar rows and the TUI's minimized row. Agreed it reads much more clearly against the "shows nothing when on pace" behavior. Left the long form (ahead of pace) surfaces as is per your note.
2. Dropped active_usage_fetched_at from the menubar snapshot along with the now unused bits in _adapt_snapshot and its test assertion, since no title marker is planned, there was no reason to carry it.

Pushed as a follow-up commit; full suite's still green (same 6 pre-existing failures on main as noted in the PR description, unrelated to this change). Let me know if anything else needs adjusting!

realiti4:
Thanks for the adjuements! Looks good, I've merged it.

PR #150, opened 2026-07-19
fix: menu bar notifications break after every uv/pipx reinstall

pencount:
rumps notifications need a resolvable bundle identifier. Command-line Python tools have no app bundle, so rumps falls back to an `Info.plist` beside the interpreter — and uv/pipx recreate that directory on reinstall or upgrade, silently breaking notifications until the plist is restored by hand.

This makes `cswap --menubar` self-healing: at launch it ensures `CFBundleIdentifier`/`CFBundleName` in the interpreter-adjacent `Info.plist`, preserving any keys already there. No-op off macOS; on a read/write failure it logs a warning and the app starts normally (notifications are the only casualty).

## Validation

- `uv run pytest -q tests/test_menubar.py` — 63 passed (two new tests: the plist is created/repaired with existing keys preserved, and the off-macOS no-op)
- `git diff --check`

### DIFF
--- a/src/claude_swap/menubar.py
+++ b/src/claude_swap/menubar.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import plistlib
 import re
 import sys
@@ -55,7 +56,10 @@ def ensure_notification_identity(
     data: dict = {}
     try:
         if path.exists():
-            loaded = plistlib.loads(path.read_bytes())
+            try:
+                loaded = plistlib.loads(path.read_bytes())
+            except Exception:
+                loaded = None  # unreadable/corrupt — rebuild from scratch
             if isinstance(loaded, dict):
                 data = loaded
         changed = False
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks a lot for the fix.

Before the merge, can we adjust two small things?

1. **A corrupt plist can crash the launch instead of degrading.** If the existing `Info.plist` *starts* like XML but is malformed (interrupted write, or one of the hand-restored plists this PR is rescuing people from), `plistlib.loads` raises `xml.parsers.expat.ExpatError` — which isn't a `ValueError`/`OSError`, so it escapes the `except` tuple, and the CLI only catches `ClaudeSwitchError`. Result: `cswap menubar` tracebacks on *every* launch until the file is deleted by hand — the exact failure mode we're trying to eliminate. I'd guard the parse separately and treat anything unparseable as empty, so the function rebuilds the file (heals) instead of giving up:

   ```python
   if path.exists():
       try:
           loaded = plistlib.loads(path.read_bytes())
       except Exception:
           loaded = None  # unreadable/corrupt — rebuild from scratch
       if isinstance(loaded, dict):
           data = loaded
   ```

   A test with a truncated plist would lock this in nicely.

2. **(Small, optional) make the write atomic** — write to a temp file and `os.replace()` it over `Info.plist`. An interrupted `write_bytes` is one way to produce exactly the corrupt file from point 1, so this closes the loop.

Everything else looks great. We can merge it after the adjustments. Thanks.

pencount:
Both addressed in 56f833b — the parse is now guarded separately so a corrupt plist rebuilds instead of crashing, the write goes through a temp file + `os.replace`, and there's a truncated-plist regression test.

realiti4:
Thanks a lot for the adjustments! I've merged it.

PR #151, opened 2026-07-20
Add a light theme (dark, light, auto) for the TUI and CLI

hex:
## Summary

Add a `ui.theme` setting (`dark`, `light`, or `auto`; default `auto`) that themes the Textual TUI and the plain CLI output.

## What's included

- A `cswap-light` Textual theme beside the existing `cswap-dark`. Rich renderables (usage bars, account cards, the auto-view log) read their colours from the active theme through a `Palette` object, so they recolour when the theme changes. The light severity and accent colours clear 4.5:1 contrast on the base background and on the highlighted and flash row backgrounds.
- A light ANSI palette in `printer.py`, selectable with `set_theme()`. The dark output stays the same, byte for byte.
- `auto` reads the terminal background with an OSC 11 query, run before the Textual input driver starts. It falls back to `dark` on a non-tty, a timeout, an unsupported terminal, or tmux, and never blocks or raises. It drains a late reply and does not honour `$TEXTUAL_THEME`.
- `Ctrl+T` cycles `dark`, `light`, `auto` while the app runs and saves the choice.
- The preference lives in a new `ui` section of `settings.json`, with `cswap config get/set/unset ui.theme`.

## Compatibility

`auto` resolves to `dark` when the terminal does not answer the OSC 11 query, so current users see no change unless their terminal reports a light background. The dark theme is the same as before in the TUI and the CLI.

## Testing

Tests cover the setting, palette resolution, the OSC 11 detection and its fall-back paths, the live toggle, and a contrast guard that checks 4.5:1 for six colours against three backgrounds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,17 @@ Run `cswap` on its own (or `cswap tui`) for the full-screen dashboard: live usag
 
 <img src="assets/tui-watch.png" width="760" alt="cswap watch — live 5h/7d usage bars for every account, with reset times and the active account marked">
 
+### Theme
+
+The TUI ships a dark theme and a light theme, both WCAG AA-contrast checked, plus `auto` (the default), which follows the terminal's background via an OSC 11 query. Pick one from the root menu's **Theme…** entry (the current one is marked), press `Ctrl+T` inside the TUI to cycle `dark → light → auto` live, or set it up front:
+
+```bash
+cswap config set ui.theme light   # or: dark, auto
+```
+
+The plain CLI output (outside the TUI) follows the same `ui.theme` setting. With `auto`, if the terminal doesn't answer the OSC 11 query, cswap falls back to the dark palette. Inside `tmux` or `screen` (which don't pass the query through) it skips the probe entirely and uses dark, so `auto` never adds a startup delay there.
+
+Toggling the theme live inside the TUI only affects new output — auto-view log lines already printed keep the colors they were written with; only lines added after the switch pick up the new theme.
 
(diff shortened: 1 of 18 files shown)


realiti4:
Hi, thanks a lot for the PR! This is a really nice take on theming — two curated themes plus auto, wired through `ui.theme` as a proper setting, and dark output staying exactly as it is. The OSC 11 fallback handling and the contrast guard test are great.

Could we adjust one thing before merge?

Can you add a **Theme…** entry to the root menu — a small submenu with **dark / light / auto**, with the current one marked? In the TUI, the menu is the discoverable path and keybindings are shortcuts, so the theme should be reachable from there too. The submenu plumbing from **Add account…** / **Remove account…** should cover it. `Ctrl+T` is great as the quick toggle; let's keep it.

One optional thought, fine as a follow-up: since `auto` is the default, terminals that never answer OSC 11 (`tmux` mainly) wait the full 150 ms on every CLI call — a `$TMUX` short-circuit could avoid that.

hex:
Done. Added a **Theme…** entry to the root menu with a `dark / light / auto` submenu (the current one marked), reachable the same way as Add/Remove account. Kept `Ctrl+T` as the quick toggle.

I also took the optional suggestion: it now skips the OSC 11 probe when `$TMUX` or `$STY` is set, so `auto` doesn't wait out the timeout under tmux/screen.

Thanks for the review!

realiti4:
Thanks a lot! Looks good, merged.

Issue #146, opened 2026-07-17
5h usage window reads incorrectly (some accounts underestimated, others flap) since 2026-07-16 evening — breaks autoswitch decisions


oconnorjoseph:
## What happened?

Starting last night (2026-07-16, evening), the 5h usage window `cswap` reports per-account has become unreliable — some accounts read **lower** than their real usage (autoswitch treats them as having headroom, then the very next request 429s with "per-token usage budget reached"), and others appear to flap between readings within seconds with no realistic way that much real usage could have changed. Net effect: `auto` (`--strategy best`, threshold 90) is picking bad targets and thrashing, which defeats the entire point of autoswitch — it's supposed to keep a session unblocked by rotating to an account with real headroom before the current one hits its limit, and right now it isn't reliably doing that.

This is a shared account pool (~14 Claude accounts on one company Anthropic org set, mostly Team plan seats) driving several concurrent Claude Code CLI sessions across parallel git worktrees, with `cswap auto` running continuously as a background launchd agent (`--strategy best`, threshold 90%, hysteresis 10%, cooldown 300s) — so autoswitch decisions matter a lot here.

**Evidence 1 — clean day, then sudden onset exactly matching "last night":**

Counting `Usage fetch failed ... http-429` lines (the hard rate-limit signal, not a cswap-side computation) per day in the log:

```
2026-07-13   26
2026-07-14   15
2026-07-15    0    <- zero all day
2026-07-16   42    <- resumes at 18:39 local and climbs through the evening
2026-07-17   61    <- (partial day so far, already exceeds the 16th)
```

First few after the gap:
```
2026-07-16 18:39:16 - WARNING - Usage fetch failed for account 16: http-429, retry-after 3600s (per-token usage budget reached; backing off)
2026-07-16 18:59:47 - WARNING - Usage fetch failed for account 16: http-429, retry-after 2369s (per-token usage budget reached; backing off)
2026-07-16 19:14:50 - WARNING - Usage fetch failed for account 16: http-429, retry-after 1466s (per-token usage budget reached; backing off)
```

No new `cswap` version was installed around this time — we're still on the latest release (`0.21.0`, shipped 2026-07-15), and the `cswap auto` daemon process itself has been running continuously since Monday (2026-07-13) with no restart. So this isn't a client-side regression from a new build landing; whatever changed, changed under a long-running, unmodified process.

[...]

PR #152, opened 2026-07-20
fix(usage): converge multi-machine 429 deadlock on a shared usage token

codeslake:
## Problem

When several machines poll the usage endpoint for the **same account**, they can deadlock each other into a permanent 429 state.

The usage endpoint budgets *usage requests* per token over a rolling ~1h window. That budget is shared by every machine polling that account, but no machine can see the others: there is no cross-machine coordination, and the response carries no remaining-request count — only a `Retry-After` once you are already blocked.

Three things then compound:

1. **A fixed post-429 floor cannot converge.** `POST_429_MIN_INTERVAL_S` puts every machine at the *same* interval, so N machines sum to N x the rate. If that exceeds the budget, they stay blocked indefinitely.
2. **`RETRY_AFTER_FLOOR_CAP_S = 900` re-probed mid-window.** The server's `Retry-After` counts down the full ~1h window accurately; capping the wait to 15 min meant probing again while still inside the window, and each probe re-arms the block, so the token never drains.
3. **Trust expiry turned a healthy account into an unusable target.** A 429 is a *polling* throttle — it does not move the account's real 5h/7d windows. But `TRUST_MAX_AGE_S` flipped a throttled account's `last_good` to "unknown" while it was plainly fine (e.g. 21% used), which drove failover flapping and "all accounts exhausted" sleeps. The flapping then re-polled the throttled token and re-armed the block.

## Fix

**AIMD on a contended token** (`poll_policy.py`). While 429s recur, each successful poll grows the interval multiplicatively (`POST_429_BACKOFF_MULT = 1.5`) toward a wider ceiling (`POST_429_MAX_INTERVAL_S = 1800`), so machines sharing a token each retreat until their combined rate fits the budget. Movement decays it back down. This is TCP-style congestion control: fair-sharing by reaction alone, with **no machine count and no shared state to configure**.

**Honor hour-scale `Retry-After`** (`usage_store.py`). `RETRY_AFTER_FLOOR_CAP_S` 900 -> 3600, so the wait spans the endpoint's actual window instead of re-arming it. Still bounded to one window, so a pathological header can never park an account for hours.

**Trust 429-stale data until its window resets** (`usage_store.py`). Usage rises monotonically within a window, so a frozen `last_good` is a valid *lower bound* on true usage right up to that window's reset. Trust it until then — data-driven, not a fixed clock. Once every window has reset the value is obsolete (usage was zeroed) and reads unknown. Rows with no `resets_at` fall back to a bounded `RATE_LIMIT_TRUST_MAX_AGE_S`. Non-429 failures are unchanged: a timeout or network error is no evidence `last_good` still holds.

This last part mirrors Claude Code 2.1.208, which shows last-known usage bars when `/usage` is rate-limited rather than erroring.

## Scope

- `src/claude_swap/poll_policy.py` +21/-1
- `src/claude_swap/usage_store.py` +75/-4
- `tests/` +123

No new configuration, no new dependencies, no platform-specific code. The constants are derived from the endpoint's observed behavior (the ~1h rolling window), not from any particular setup.

## Testing

- 7 new tests covering AIMD growth, the ceiling, the reset-driven trust window, and the no-`resets_at` fallback.
- `tests/test_poll_policy.py` + `tests/test_usage_store.py`: **83 passed**.
- Full suite: **1420 passed**, same 17 pre-existing failures as `main` at `1964097` (all in `test_switcher.py`, ANSI-color assertions unrelated to this change) — no regressions.

Observed in practice across two machines sharing one account: the 429 oscillation stops and the pollers settle at staggered intervals.

---

🤖 PR description and upstream rebase prepared with [Claude Code](https://claude.com/claude-code); the fix itself was authored and field-tested by the submitter.

### DIFF
--- a/src/claude_swap/poll_policy.py
+++ b/src/claude_swap/poll_policy.py
@@ -82,6 +82,20 @@ EDGE_BACKOFF_S = 300.0
 POST_429_MIN_INTERVAL_S = 360.0
 RECENT_429_WINDOW_S = 3600.0
 
+# AIMD on a contended token. The usage endpoint's per-token budget is shared
+# across every machine polling the same account, and none of them can see the
+# others (there is no cross-machine coordination, and the endpoint exposes no
+# remaining-request count — only a Retry-After once already blocked). So while
+# 429s recur on a token, each successful poll multiplicatively grows the
+# interval (×POST_429_BACKOFF_MULT) toward POST_429_MAX_INTERVAL_S — wider than
+# the normal candidate ceiling so several machines can each back off far enough
+# that their combined rate fits under the budget. Movement (a real success run
+# with no recent 429) decays it back down. This is TCP-style congestion control:
+# a token gets fair-shared by reaction alone, with no machine count or shared
+# state to configure.
+POST_429_BACKOFF_MULT = 1.5
+POST_429_MAX_INTERVAL_S = 1800.0
+
 # The engine escalates to a full candidate refresh when the active account is
 # within this margin of the threshold (decision policy, but the urgent-mode
 # cadence keys on the same band, so it lives with the cadence numbers).
@@ -188,7 +202,12 @@ def plan_after_fetch(
     ):
         interval = URGENT_INTERVAL_S
     if recent_429:
-        interval = max(interval, POST_429_MIN_INTERVAL_S)
+        # AIMD additive-increase: grow the interval multiplicatively from the
+        # last one toward the wider 429 ceiling, so machines sharing a
+        # contended token each retreat until their combined rate fits the
+        # budget. Floored at POST_429_MIN_INTERVAL_S for the first 429.
+        increased = max(base * POST_429_BACKOFF_MULT, POST_429_MIN_INTERVAL_S)
+        interval = min(POST_429_MAX_INTERVAL_S, max(interval, increased))
 
     next_poll = now + interval * (1.0 + JITTER_FRAC * (2.0 * rng() - 1.0))
     headroom = oauth.account_headroom(new_usage, models)
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -1384,6 +1384,7 @@ class ClaudeAccountSwitcher:
             num: (info.get("email", ""), info.get("organizationUuid", "") or "")
             for num, info in data.get("accounts", {}).items()
         }
+        # No models needed: only fetched_at is read, never scoped-window trust.
         return {
             num: entry.fetched_at
             for num, entry in self._usage_store.entries(identities).items()
@@ -2794,13 +2795,16 @@ class ClaudeAccountSwitcher:
             for num, email, _org_name, org_uuid, _active, _creds, _alias in accounts_info
         }
         info_by_num = {str(info[0]): info for info in accounts_info}
+        # Scoped-window models so the 429-stale trust bound honors per-model
+        # (e.g. Fable) resets, matching the poll planner's window view.
+        _threshold, models = self._poll_policy_inputs()
         sentinels: dict[str, str] = {}
         for num, info in info_by_num.items():
             static = self._static_usage_sentinel(info)
             if static is not None:
                 sentinels[num] = static
 
-        entries = store.entries(identities)
+        entries = store.entries(identities, models)
         # Dead refresh-token lineage: quarantine. Surfacing the sentinel here both
         # drives the "re-login needed" display and (via ``num not in sentinels``
         # below) stops the endless fetch loop that would otherwise 401/429 forever.
@@ -2825,7 +2829,7 @@ class ClaudeAccountSwitcher:
             for num, record in records.items():
                 if record.sentinel is not None:
                     sentinels[num] = record.sentinel
-            entries = store.entries(identities)
+            entries = store.entries(identities, models)
             self._persist_poll_plans(
                 records, pre, entries, info_by_num, identities
             )
@@ -2891,6 +2895,7 @@ class ClaudeAccountSwitcher:
         try:
             identities = {number: (email, org_uuid or "")}
             now = self._usage_store.clock()
+            # No models needed: only fetched_at/next_poll_at is read here.
             entry = self._usage_store.entries(identities).get(number)
             if entry is None or entry.fetched_at is None:
                 return
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -2863,11 +2863,7 @@ class ClaudeAccountSwitcher:
             before, after = pre.get(num), post.get(num)
             if after is None or after.fetched_at is None:
                 continue
-            recent_429 = (
-                before is not None
-                and before.last_429_at is not None
-                and (now - before.last_429_at) < poll_policy.RECENT_429_WINDOW_S
-            )
+            recent_429 = before is not None and before.recent_429(now)
             plans[num] = poll_policy.plan_after_fetch(
                 prev_interval_s=before.poll_interval_s if before else None,
                 prev_usage=before.last_good if before else None,
--- a/src/claude_swap/usage_store.py
+++ b/src/claude_swap/usage_store.py
@@ -35,6 +35,7 @@ from claude_swap.locking import FileLock
 from claude_swap import oauth
 from claude_swap.poll_policy import (
     EDGE_BACKOFF_S,
+    RECENT_429_WINDOW_S,
     SERVE_TTL_S,
     parse_reset_ts,
 )
@@ -181,6 +182,38 @@ class UsageEntry:
 
     def in_backoff(self, now: float) -> bool:
         return self.backoff_until is not None and now < self.backoff_until
+
+    def recent_429(self, now: float) -> bool:
+        """Whether this token 429'd recently enough to keep the post-429 cadence.
+
+        Recency is measured from when the 429's honored backoff *lifts*, not
+        from the 429 itself. An hour-scale Retry-After is honored as one long
+        backoff during which no attempt runs, so the only 429 stamp is at the
+        block's start; measuring recency from the stamp would make the first
+        post-block success (which can't happen until the backoff lifts) see a
+        window that has already fully elapsed — the AIMD growth and the
+        POST_429 floor would never engage, and machines sharing the token could
+        not converge. Anchoring on the backoff's end keeps "recent" true across
+        the first post-block success while a short (``Retry-After: 0``) block
+        still expires normally. Bounded by ``RECENT_429_WINDOW_S`` past the
+        anchor so it clears once the saturated window has aged out.
+        """
+        if self.last_429_at is None:
+            return False
+        anchor = self.last_429_at
+        # Use the backoff anchor only while the LIVE backoff is a 429 backoff.
+        # last429At is never cleared, but backoffUntil/lastError are rewritten by
+        # any later failure — so without the lastError guard an unrelated timeout
+        # on a token that 429'd long ago would install a fresh backoffUntil and
+        # spuriously re-arm the post-429 cadence. Success clears lastError and
+        # backoffUntil; only a 429 sets last429At and backoffUntil together.
+        if (
+            self.last_error == "http-429"
+            and self.backoff_until is not None
+            and self.backoff_until > anchor
+        ):
+            anchor = self.backoff_until
+        return now < anchor + RECENT_429_WINDOW_S
 
     def claimed(self, now: float) -> bool:
         """A collector stamped this entry moments ago (fetch may be in flight)."""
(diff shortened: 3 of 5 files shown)


realiti4:
Hey, thanks for this — and for field-testing it before submitting. Interesting timing: there have been similar hour-scale `Retry-After` regime while investigating #146 last week (it appears to have changed server-side around July 16), so your report usefully corroborates that observation.

I'd like to get parts of this in, but I found a few things while testing that need adjusting first:

**The `Retry-After` cap raise (`900 → 3600`)** — happy to take this as-is, with one correction to the comments: our measurements show probing does not re-arm the block. We have three probes from one episode whose `Retry-After` values all count down to the identical wall-clock deadline. The raise is still worth it (it stops wasting ~3 probes per block), but the "re-arming" rationale would enshrine a wrong model of the limiter, so could you reword those comments?

**The AIMD backoff** — there's an integration problem, and it's caused by the cap raise itself: `recent_429` is checked strictly against `RECENT_429_WINDOW_S` (3600s), and `last429At` only gets stamped on failed attempts. Once a 3600s `Retry-After` is honored, there's a single stamp and then no attempts for exactly 3600s — so the first post-block success sees `recent_429=False` and the growth never engages. It actually loses the existing 360s floor too (on main, the mid-block re-probes kept re-stamping `last429At`). The unit tests pass `recent_429=True` directly, which is why they don't catch it. I think the fix is to make the 429 memory survive the honored backoff (e.g. stamp at backoff expiry, or compare against it), plus one test that goes through the store:

```text
429 with Retry-After: 3600
→ advance clock
→ successful fetch
→ assert the floor/growth apply
```

**The trust extension** — the flapping problem you're fixing is real, but trusting until `max(reset_ts)` means a frozen snapshot can drive decisions until the 7-day reset, and there's no client-side ceiling at all — a malformed far-future `resets_at` would control trust indefinitely. There's also a hole with partial metadata: if 5h has no `resets_at` but 7d does, the 7200s fallback never applies — and 5h-without-`resets_at` is a shape the server actually sends right now (came up in #146). Could you clamp it to the earliest future reset with a bounded ceiling (the existing `RATE_LIMIT_TRUST_MAX_AGE_S` would do), and pass the configured models into `relevant_windows`? FWIW I made agent scan Claude Code 2.1.215 binary — the stale-fallback behavior you cite is real, but it's bounded at exactly 1h there too.

One question out of curiosity, because it would settle something we've been trying to pin down: do your two machines use the same exported/imported credentials, or did each log in separately? If separate logins were contending on one account, that tells us the endpoint's budget is no longer strictly per-token, which changes our cadence math.

codeslake (commented):
Thanks for the thorough review — all on point. Pushed three commits (on top of the current head), green (`NO_COLOR=1 pytest`, 1450 passed / 3 skipped). In order:

**1. `Retry-After` cap comment (`7d682a0`)** — reworded to the measured model: the server's `Retry-After` counts down to a fixed wall-clock deadline and is *not* re-armed by probing; the old 900s cap just wasted ~3 probes per block. No behavior change.

**2. AIMD integration bug (`ded82e7`)** — your diagnosis was exactly right. Fix: recency is now anchored on when the 429 backoff *lifts*, not the stamp, via `UsageEntry.recent_429(now)` — so the first post-block success still sees `recent_429=True` and the growth/floor engage. A `last_error == "http-429"` guard keeps an unrelated timeout from re-arming it (caught in self-review; regressioned). Store-level tests, as you asked: `test_recent_429_true_at_first_success_after_hour_block`, `test_floor_engages_at_first_post_block_success` (+ a `test_legacy_recency_would_drop_the_floor` documenting the old defect), `test_repeated_429_episodes_converge_to_the_wide_ceiling`, `test_unrelated_timeout_does_not_re_arm_recency`.

**3. Trust bound (`dfb47a5`)** — now `now < min(earliest relevant-window reset, fetched_at + RATE_LIMIT_TRUST_MAX_AGE_S)`: earliest reset not `max`; the ceiling always applies (kills far-future/malformed `resets_at`); a window missing `resets_at` can only tighten, never loosen (the partial-metadata shape you flagged); and `models` is now threaded into `relevant_windows`. Tests: `test_far_future_reset_is_clamped_to_the_ceiling`, `test_trust_keys_on_earliest_future_reset`, `test_partial_metadata_still_bounded_by_ceiling`, `test_ceiling_wins_when_reset_is_beyond_it`.

**4. Same vs separate tokens** — I fingerprint-compared the two machines. On the account that deadlocked the refresh tokens are **different** (separate logins), and a control account where I do sync came back byte-identical, so it's a real case of two distinct tokens on one account → the budget is **not strictly per-token**. You were right.

I deliberately kept the fix agnostic to this, though, since the answer isn't stable (sync vs separate login can change over time). `plan_after_fetch` never inspects token identity — it reacts only to `recent_429`, and AIMD converges for any pollers sharing any throttled counter, per-token or per-account alike. What *does* change with per-account is the constant tuning (the `~28-30/hr` numbers are single-machine), not the mechanism; `POST_429_MAX_INTERVAL_S` is the knob to widen for larger fleets. `TestBudgetInvariants` is left as-is — the N>1 margin is reasoned, not measured.

---

**Separate follow-up (out of scope here):** CC 2.1.216 no longer polls `/api/oauth/usage` for its own numbers — it reads `anthropic-ratelimit-unified-{5h,7d}-utilization`/`-reset` off the `/v1/messages` response headers. cswap can't do that directly (it sends no messages), but a forward-proxy in front of CC already captures those headers to disk (e.g. the claude-code-cache-fix `usage-log` extension, [#251](https://github.com/cnighswonger/claude-code-cache-fix/pull/251), appends them to `~/.claude/usage.jsonl`). So cswap could prefer that header-sourced usage and poll only as fallback — pushing the endpoint rate toward zero.

Big caveat, which is why it's not this PR: the headers only ever describe the **account whose session is sending messages**. The moment cswap switches away, that account stops emitting `/v1/messages`, so its header stream goes stale — and the *candidate* accounts cswap most needs fresh numbers for (to pick the next target) never send messages at all. So header-sourcing can cover the active account but the candidates still need polling; it shrinks the 429 surface rather than eliminating it, and it needs a stable path/schema contract so cswap doesn't couple to a proxy's internals. Happy to open an issue if it's of interest.

---

Both machines are running these commits now; I'll report back if the deadlock recurs.

realiti4:
Thanks a lot for the adjustments, it looks good and I've merged it.

And thanks for the token fingerprint comparison, so it looks like budget is per-account now.

Thanks for the header-sourcing pointer too. I'll be looking into that side myself as well. While poking at recent binaries it looks like Claude Code also persists a last-known usage snapshot client-side (in `~/.claude.json`), but it's only written when usage is actually viewed via command etc., so it's free opportunistic reading for the active account rather than something to rely on; agreed either way that candidate accounts will still need polling.

I'm curious to see how it behaves across both machines too, so please do.

codeslake:
Both machines are on the merged build now (0.23.0b1, identical `usage_store`), and I've armed a watch for the next hour-scale 429 episode. None since deploy, so nothing to report yet, but I'll post what the AIMD floor/growth actually does the first time it fires across both.

On the `~/.claude.json` snapshot: agree it's opportunistic-only — active account, refreshed just on an explicit usage view, so candidates still fall back to polling either way. And since cswap sends no `/v1/messages` of its own, header-sourcing doesn't help the candidate side for us either. Polling stays the floor.

Will circle back with the cross-machine numbers once it fires — thanks for the quick review and merge.

codeslake:
Field report from the two machines, as promised. Merged build (0.23.0b1, identical `usage_store`) on a Linux server and a Mac since 07-20, ~32h of continuous state sampling (`cache/usage.json` every 60s on both, ~3.5k samples) plus token-fingerprint sampling every 300s.

## 1. The integration bug you flagged is fixed in practice

The case you predicted — a 3600s `Retry-After` honored, then the first post-block success sees `recent_429=False` so growth never engages — engages now. One episode, verbatim from the sampler:

```
02:08  429 #1, Retry-After 3600 → honored to 03:08
03:12  first retry after the block → 429 #2
04:16  success → interval 600 → 900     ← past CANDIDATE_MAX, growth engaged
04:34  429 #3
05:35  429 #4
06:41  success → interval 900 → 1350    ← growth continued
07:04  recency window lapsed → back to 600
```

Pre-fix this account would have sat at 600 and re-probed on the old cadence. Floor/growth survives the honored block exactly as the store-level test asserts.

Equally important: **no 429 storm and no flapping through four consecutive blocks.** Each block was honored to the second, exactly one probe per block, and the peer machine kept polling the same account normally throughout. The deadlock this PR set out to fix did not reappear.

Provenance caveat: those 429s were **self-inflicted**. I was debugging an unrelated proxy on the Mac and its test processes hammered the usage endpoint with that account's token. Useful as a stimulus, not evidence about the endpoint's natural 429 rate. The Linux box, polling 180-600s and never loaded that way, saw **zero 429s in 32h**.

## 2. Two token-scope observations

**Usage-poll limit looks per-token, not per-account.** During the episode above, at `03:12:09` the Mac got a 429 for account A while 8 seconds later, at `03:12:17`, the Linux box fetched the same account A successfully. Same account, both machines, tokens already diverged. So the poll budget appears token-scoped while the message budget is account-scoped (the fingerprint comparison from last time) — different layers, probably worth separating in the cadence model.

**Lineage diverges on its own.** Refresh rotates the refresh token, so two machines starting from one exported credential stop sharing a lineage at the first refresh (three generations over 32h, ~8h apart). A re-login is distinguishable from a routine rotation by the refresh-token expiry: rotation preserves it, re-login resets it ~4 weeks out.

## 3. Two credential experiments

**A — does a re-login revoke the peer's grant?** No. Ran `/login` on Linux at 13:37, minting a fresh lineage; the Mac kept polling the same account successfully on its old grant six minutes later, `authDeadStrikes=0`. Grants are independent, so re-authing one box does not knock its peer offline. (Distinct from the *reuse* case in #164, where presenting a superseded token from the same lineage is what appears to revoke it.)

**B — same credential live on two machines.** Synced the Linux credential into the Mac keychain so both held one lineage, then watched. What surfaced was not a rotation race but a stall, and it is worth reporting precisely:

```
14:03  sync — both machines hold lineage L; cswap's own backup for that slot still holds L-1
21:37  access token expires
       cswap: an owner (a live Claude Code) is present → defers, does not refresh
       Claude Code: idle, issues no API request → does not refresh
       daemon: "headless daemon cannot complete OAuth", reschedules ~4h out
       → nobody refreshes; the access token stays dead
21:51  autoswitch: usage unknown ×3 → failover to the other account
       log: "Backed up account 2 (lineage differs from the stored backup and
             ownership could not be verified — pre-fix backup)"
00:56  I manually switch back to that account
00:57  cswap switches away again, 56 seconds later, same reason
```

The chain is: a dead access token makes `fetch_oauth_profile` return 401, so the identity oracle cannot resolve ownership, so `_classify_outgoing_credential` falls to `unresolved`, so the switch takes the conservative pre-fix backup — and the next tick repeats it. The account was healthy the whole time; its refresh token was never dead. `cswap recover` (#165/#166) addresses the manual side of this, and deliberately leaves auto/watch alone, so the automatic path still oscillates.

Note the divergence in the first line: the lineage mismatch is not exotic. Any cross-machine credential sync produces it — my own sync tooling writes the live keychain item and leaves cswap's backup to "resync on the next switch", and #164's export/import path reaches the same state from the other direction. Once cswap's backup and the live credential disagree, the only thing that can re-attribute them is the profile endpoint, which needs a live access token — which is exactly what nobody refreshed.

I am not proposing anything here; #164 already covers the export/import side thoroughly and #166 covers the manual recovery. Flagging the runtime piece in case it is useful context for either.

## 4. `network` / `timeout` never re-arms 429 recency

The Mac hit 8 consecutive `network` failures over ~40 minutes (its uplink, not yours). Throughout, `last429At` stayed untouched — 102h old on one account, `None` on the other — and the interval never moved to the post-429 floor. `test_unrelated_timeout_does_not_re_arm_recency` holding under real sustained failure rather than a fixture.

## 5. Unreported, will file separately

`claude_locks.py` targets `~/.claude.lock` with a 10s staleness. Claude Code 2.1.218 guards its OAuth refresh with **two** locks — a new `<config-home>/.oauth_refresh.lock` plus the legacy `~/.claude.lock` it still takes for compatibility — both at `stale: 60000, update: 5000`, with a dedicated `tengu_oauth_refresh_legacy_lock_contended` counter on the legacy path. Mutual exclusion holds today only because of that shim, and the 10s-vs-60s gap means cswap can take over a lock Claude Code still considers held if its toucher stalls. Opening that separately with the binary evidence rather than burying it here.

Samplers stay running; will report anything organic.

realiti4:
Thanks a lot, the report looks good. I didn’t want to hold up the fixes, so I also released v0.23.0.

I’ll be away from my computer for a couple of days, but I’ll dig into it properly when I get back.

PR #155, opened 2026-07-21
fix: reject directory destinations in export's atomic write (fixes #154)

### LINKED ISSUE #154: `--export` with a directory as destination corrupts the directory's permissions (chmod 0600, execute bit lost)
**Version:** 0.22.0

**Summary**

If the destination passed to `cswap --export` is an existing directory (instead of a file path), `_atomic_write_file()` in `claude_swap/transfer.py` silently succeeds in a confusing way and, as a side effect, strips the execute bit from the destination directory — leaving it as `0600` (`drw-------`) instead of `0700`/`0755`. Every file inside becomes inaccessible ("Permission denied" on stat, even though names are still listable), and — in a VMware guest — copying that folder to the host then fails with "Cannot get attributes of file on virtual machine."

**Root cause** (`transfer.py`, `_atomic_write_file`, ~lines 97-105):

```python
def _atomic_write_file(path: Path, content: str) -> None:
    temp_path = path.with_suffix(f".{os.getpid()}.tmp")
    temp_path.write_text(content, encoding="utf-8")
    if sys.platform != "win32":
        os.chmod(temp_path, 0o600)
    shutil.move(str(temp_path), str(path))
    if sys.platform != "win32":
        os.chmod(path, 0o600)
```

When `path` is a directory (e.g. `~/Desktop/Career`):

1. `path.with_suffix(f".{pid}.tmp")` builds a *sibling* path, e.g. `~/Desktop/Career.<pid>.tmp`.
2. `temp_path.write_text(...)` creates that sibling file.
3. `shutil.move(temp_path, path)` — since `path` is an **existing directory**, `shutil.move` moves the temp file *into* it instead of renaming it, giving `~/Desktop/Career/Career.<pid>.tmp`.
4. `os.chmod(path, 0o600)` still targets the original `path` — which is the **directory**, not the file that was just written. This removes the directory's execute bit (`0600` = `drw-------`), so nothing inside can be `stat()`-ed anymore.

**Reproduction**

```
mkdir -p /tmp/dest_dir
cswap --export /tmp/dest_dir
stat /tmp/dest_dir   # -> drw------- (0600), execute bit gone
ls -la /tmp/dest_dir # -> a stray dest_dir.<pid>.tmp file inside
```

Reproduced twice on a real machine with real user data: once with `~/Desktop` itself as the destination, once with a subfolder (`~/Desktop/Career`) — both times the folder's permissions broke and its contents became temporarily inaccessible until manually `chmod 700`-ed back.

**Suggested fix**

`_atomic_write_file` (and any other export/import path building on it) should reject a directory destination up front with a clear `TransferError`, e.g.:

```python
def _atomic_write_file(path: Path, content: str) -> None:
    if path.is_dir():
        raise TransferError(
            f"export destination must be a file path, not a directory: {path}"
        )
    ...
```

That turns a silent, destructive permission corruption into an immediate, actionable error — no directory ever gets `chmod`'d, and the user gets pointed at the fix (pass a file path, e.g. `cswap --export ~/Desktop/Career/export.json`).

Happy to send a PR with this guard if useful.

farhan6667:
Fixes #154.

`_atomic_write_file()` in `src/claude_swap/transfer.py` assumed its `path` argument is always a file. When `cswap --export` is given an existing directory as the destination instead:

1. `path.with_suffix(f".{pid}.tmp")` builds a sibling temp path next to the directory.
2. `shutil.move(temp_path, path)` — since `path` is an existing directory, `shutil.move` drops the temp file *into* it rather than renaming, so the write appears to "succeed."
3. `os.chmod(path, 0o600)` then runs against the original `path` — the directory itself — stripping its execute bit (`0600` = `drw-------`). Every file inside becomes unreadable via `stat`/`access` until the directory is manually `chmod`'d back.

This reproduced twice on a real machine (`~/Desktop` and a subfolder), corrupting real user data directories, and also broke VMware guest→host file copy ("Cannot get attributes of file on virtual machine") since the copy tooling stats every file recursively.

**Fix:** `_atomic_write_file` now raises `TransferError` immediately if `path.is_dir()`, before any write or chmod happens — turning silent, destructive permission corruption into an immediate, actionable error that points the user at the real fix (pass a file path, e.g. `cswap --export ~/Desktop/Career/export.json`).

No behavior changes for the normal case (file path destination).

### DIFF
--- a/src/claude_swap/transfer.py
+++ b/src/claude_swap/transfer.py
@@ -96,6 +96,10 @@ def _validate_imported_account(switcher: ClaudeAccountSwitcher, account: dict) -
 
 def _atomic_write_file(path: Path, content: str) -> None:
     """Write text atomically with 0600 perms — same pattern as switcher._write_json."""
+    if path.is_dir():
+        raise TransferError(
+            f"export destination must be a file path, not a directory: {path}"
+        )
     temp_path = path.with_suffix(f".{os.getpid()}.tmp")
     temp_path.write_text(content, encoding="utf-8")
     if sys.platform != "win32":


realiti4:
Thanks a lot for the fix, I've merged it.

PR #158, opened 2026-07-21
fix(windows): retry atomic file replace past transient sharing failures

bfishman:
# fix(windows): retry atomic file replace past transient sharing failures

## The error

On Windows, cswap intermittently fails a switch with:

```
WARNING  credentials.py Failed to write credentials file: [WinError 5] Access is denied:
  '...\credentials\tmpXXXX.tmp' -> '...\credentials\.creds-1-<email>.enc'
ERROR    switcher.py Switch failed: [WinError 5] Access is denied, attempting rollback
```

The same failure also hits the usage store, where it is swallowed rather than surfaced — writes are silently dropped, which makes the adaptive autoswitch scheduler behave erratically (poll intervals not backing off, escalations not refreshing candidates). The failures are load-correlated: the busier the machine, the more often they occur.

## Suspected cause

Every atomic write in cswap ends in `os.replace(tmp, target)`. On POSIX, `rename(2)` is genuinely atomic and cannot fail this way. On Windows it can: antivirus (Defender) and the search indexer open freshly-created files opportunistically to scan them, and a replace onto a file another process holds open fails with `ERROR_ACCESS_DENIED` (5) or `ERROR_SHARING_VIOLATION` (32). The handle is released within milliseconds, so the failure is transient — but unretried, it propagates as a hard write failure.

Measured directly on an affected Windows 11 machine — 4000 sequential tmpfile-then-replace cycles into a Defender-scanned directory:

```
1779/4000 os.replace failures (~44%); worst-case replace latency 33ms
```

This is invisible to Linux and macOS CI, which is why it has not surfaced before.

## The fix

`fsutil.replace_with_retry()` — `os.replace` retried with exponential backoff (2 ms doubling to a 250 ms cap, 10 attempts) on the three Windows error codes that mean "someone has this file open right now" rather than "this operation is invalid": `ERROR_ACCESS_DENIED` (5), `ERROR_SHARING_VIOLATION` (32), `ERROR_LOCK_VIOLATION` (33).

Deliberate constraints:

- **win32 only.** On POSIX the retry branch is unreachable, so behavior there is byte-for-byte unchanged.
- **Genuine errors are not retried.** A missing source or cross-device link raises on the first attempt rather than being retried into a delay. Covered by a test.
- **`fsutil` is a new dependency-free leaf module.** `settings`, `credentials`, `mappings` and `session` all need the helper, and they sit on both sides of the existing `paths → models → usage_store → settings` import cycle; putting it in `paths` triggers a circular import.

Applied to the seven single-file atomic writes: `credentials.py` (3), `settings.py`, `mappings.py`, `migrations.py`, `session.py`.

## Deliberately out of scope

The `os.replace` **directory** renames in `switcher.py` are untouched. They are plausibly vulnerable to the same locking, but no failures were observed there and a directory rename may warrant different handling — worth a follow-up rather than a speculative change here.

## Testing

Six new tests in `tests/test_paths.py` covering: retry-then-succeed for each of the three transient codes, give-up-and-raise after the attempt budget, no-retry-on-genuine-error, and no-retry-on-POSIX. Written failing first, then implemented. Because they monkeypatch `os.replace` and `sys.platform`, they run and assert on every platform, including a Linux/macOS CI.

Full suite on the affected Windows 11 machine (Python 3.14): **1428 passed, 34 skipped, 0 failed.**

Causal evidence the fix addresses the observed symptom: before the change, the autoswitch suite failed a varying 3-test subset on slower runs (e.g. 3 failures in a 74 s run). After, three consecutive runs at 54 s / 101 s / 111 s all passed 105/105 — including runs substantially slower, and therefore under more antivirus contention, than the ones that previously failed.

### DIFF
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -29,6 +29,7 @@ from typing import NamedTuple, Protocol
 
 from claude_swap import macos_keychain
 from claude_swap.exceptions import CredentialError, CredentialWriteError
+from claude_swap.fsutil import replace_with_retry
 from claude_swap.models import Platform
 from claude_swap.paths import (
     get_claude_config_home,
@@ -445,7 +446,7 @@ class CredentialStore:
             os.write(fd, json.dumps(data, indent=2).encode("utf-8"))
             os.close(fd)
             fd = -1
-            os.replace(tmp_path, str(path))
+            replace_with_retry(tmp_path, str(path))
             if sys.platform != "win32":
                 os.chmod(str(path), 0o600)
         except BaseException:
@@ -468,7 +469,7 @@ class CredentialStore:
             os.write(fd, credentials.encode("utf-8"))
             os.close(fd)
             fd = -1
-            os.replace(tmp_path, str(cred_file))
+            replace_with_retry(tmp_path, str(cred_file))
             if sys.platform != "win32":
                 os.chmod(str(cred_file), 0o600)
         except BaseException:
@@ -803,7 +804,7 @@ class CredentialStore:
             os.write(fd, encoded.encode("utf-8"))
             os.close(fd)
             fd = -1
-            os.replace(tmp_path, str(enc_file))
+            replace_with_retry(tmp_path, str(enc_file))
             if sys.platform != "win32":
                 os.chmod(str(enc_file), 0o600)
         except BaseException:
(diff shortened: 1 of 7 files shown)


realiti4:
Thanks for the fix, merged it!

PR #160, opened 2026-07-22
fix: fence long-running usage fetches

### LINKED ISSUE #157: Usage fetches can outlive the 10-second cross-process claim
### What happened?

Multiple local cswap surfaces share `cache/usage.json`, but the cross-process claim can expire while the request it protects is still running.

`UsageStore.reserve()` stamps `lastAttemptAt` for every selected slot before network I/O. `UsageEntry.claimed()` treats that stamp as live for only `CLAIM_TTL_S = 10` seconds. The bounded inactive-account path can exceed that:

```text
initial usage request       up to 5s
OAuth refresh after 401    up to 10s
usage retry                up to 5s
```

The collector also claims the full fetch set before starting its thread pool, then staggers starts by 250ms per account. Executor queueing and inventory staggering can therefore consume more of the claim before a particular request starts.

If `cswap watch`, `list`, the menu bar, or another collector arrives after the 10-second stamp expires, it can reserve the same slot while the first collector is still fetching. That duplicates usage traffic against the endpoint's constrained request budget and allows two outcomes to race into the shared row.

Expected behavior:

- the in-flight lease covers the bounded refresh + usage + retry path, inventory staggering, and executor queueing;
- a completed fetch releases the lease immediately, so the intentional 60-second urgent polling mode is not delayed by a long claim TTL;
- sentinel outcomes also release the operational lease without persisting the sentinel;
- a late outcome cannot overwrite a newer claim or a rebound account slot.

Proposed fix:

- persist a separate 90-second `claimUntil` lease and random `claimId` fencing token instead of deriving new in-flight state from `lastAttemptAt`;
- pass won claim IDs through the fetch cycle and accept an outcome only when both its slot identity and claim ID still match under the file lock;
- atomically commit each accepted successful measurement with its next poll plan, closing the completion-to-replan window;
- clear the lease for accepted success, failure, and sentinel outcomes;
- revoke an outstanding usage claim when credentials are refreshed;
- retain `lastAttemptAt` as attempt provenance and honor legacy rows that contain only its original 10-second claim.

The cache schema remains additive/backward-compatible. Long-running processes already executing pre-patch code should be restarted during upgrade: they cannot understand new additive lease fields until they reload the code.

I have a tested patch and will submit a focused PR.

_Update after filing:_ latest upstream `main` already contains PR #152's hour-scale `Retry-After` and multi-machine 429 fixes. This issue is limited to the independent local in-flight lease problem; no duplicate Retry-After change is proposed.

### cswap version and install method

`cswap 0.23.0b1`, installed with `uv` from source. Reproduced by code-path timing analysis on current upstream `main` (`c097d76`); the original field incident was on `1964097`.

### Operating system

macOS

### How do you run Claude Code?

CLI (terminal)

### Command output or logs

```shell
No credential or account identifiers are required to reproduce: the race is the 10-second claim lifetime versus the bounded 5s + 10s + 5s request path and pre-fetch staggering.
```

possibilities:
## Summary

- replace the 10-second `lastAttemptAt`-derived usage claim with a separate 90-second `claimUntil` lease
- fence each fetch with a random `claimId`, rejecting late results when either the claim or slot identity changed
- commit each accepted successful measurement and its next poll plan in one cache transaction
- release accepted success, failure, and sentinel leases immediately, so the 60-second urgent cadence is not delayed
- revoke an in-flight usage claim when credentials are refreshed
- retain the legacy 10-second interpretation for cache rows written without `claimUntil`

Closes #157.

## Why

The inactive usage path can spend up to 5 seconds on the initial request, 10 seconds refreshing after a 401, and another 5 seconds retrying usage. Inventory staggering and executor queueing add more time before some requests begin.

A second local surface could therefore reserve the same account after the existing 10-second claim expired while the first request was still running. Besides duplicating constrained provider traffic, the two outcomes could race into the shared row.

## Design notes

`UsageStore.reserve()` now returns the fencing IDs it won. The collector carries those IDs through the fetch and `record()` validates the stored identity and claim under the same file lock used for the outcome write.

Successful poll plans are calculated before the write and committed with the accepted measurement. Rejected records cannot install sentinels, poll plans, or dead-token state.

The cache change is additive and keeps `schemaVersion = 2`. Existing rows without the new field retain the original short claim behavior. Long-running processes executing pre-patch code should be restarted when upgrading because running old code cannot interpret additive fields.

## Validation

- `uv run pytest tests/test_usage_store.py tests/test_poll_policy.py tests/test_switcher.py` — **428 passed**
- focused lease, stale-writer, identity-rebind, credential-refresh, mixed-outcome, and atomic-plan regressions are included
- `git diff --check` — passed
- full local suite — **1460 passed, 3 skipped, 6 failed**; the same six credential move/swap tests fail unchanged on clean base `c097d76` on this host because its local keychain backend returns empty values

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -2858,28 +2858,29 @@ class ClaudeAccountSwitcher:
             for num in info_by_num
             if num not in sentinels and (fetch is None or num in fetch)
         ]
-        to_fetch = store.reserve(
+        claims = store.reserve(
             requested, identities, respect_plans=fetch is None
         )
 
-        if to_fetch:
+        if claims:
             pre = entries
             records = self._run_usage_fetches(
-                [info_by_num[num] for num in to_fetch]
+                [info_by_num[num] for num in claims]
             )
-            store.record(records, identities)
-            for num, record in records.items():
+            plans = self._plans_after_fetch(records, pre, info_by_num)
+            accepted = store.record(records, identities, claims, plans)
+            accepted_records = {
+                num: record for num, record in records.items() if num in accepted
+            }
+            for num, record in accepted_records.items():
                 if record.sentinel is not None:
                     sentinels[num] = record.sentinel
             entries = store.entries(identities, models)
-            self._persist_poll_plans(
-                records, pre, entries, info_by_num, identities
-            )
             # A fetch that just returned invalid_grant advances the strike to the
             # dead threshold. The pre-fetch quarantine scan above couldn't see it,
             # so surface "re-login needed" in *this* pass instead of leaving the
             # slot looking merely refresh-failed until the next refresh notices.
-            for num in to_fetch:
+            for num in accepted:
                 if entries[num].token_dead():
                     sentinels[num] = USAGE_RELOGIN_REQUIRED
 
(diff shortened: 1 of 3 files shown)


realiti4:
Hi, thanks a lot for addressing this and the PR. I think it is a good solution, looks good, merged it

PR #162, opened 2026-07-22
Add an opt-in vertical-meter watch layout (ui.watch_style)

hex:
cswap-meters-watch
## Summary

Add a `ui.watch_style` setting (`classic` or `meters`; default `classic`) that chooses the `cswap watch` layout. `classic` is the current horizontal-bar account list. `meters` is a compact grid of vertical gradient meters, one column per usage window. The default is unchanged, so existing users see the same watch screen unless they opt in.

## What's included

- A `MeterWatchScreen` beside the existing list view. Every account is a framed card of vertical 5h/7d (and per-model) meters, with the percentage in large digits and the reset time beneath. `s` arms selection, the arrows move a 2-D cursor over the grid, and Enter switches and keeps watching. A `watch_screen()` factory picks the screen from the setting.
- The meters read their colours from the active theme's `Palette`, so they follow `ui.theme` in dark and light. The maxed-window reds and the active-account green frame are fixed on purpose.
- `Ctrl+V` swaps the layout live from either watch screen and saves the choice.
- The grid is responsive. It fits the column count and bar height to the terminal, and falls back to a one-line-per-account view when even minimum-height cards will not fit.
- The preference lives in the `ui` section of `settings.json`, with `cswap config get/set/unset ui.watchStyle`.

## Auto-switch from the meters view

The second commit adds the auto-switcher to this screen, so I can arm it without leaving the display.

- `a` runs the engine in dry-run. The card it would move to next gets a breathing frame, and the cycle speeds up as the active account nears the threshold. `L` asks for the same confirmation the auto screen uses, then goes live. A real switch sweeps the green frame from the old card to the new one. The mode and the next target read out at the right end of the footer, so the grid keeps every row.
- The "next" comes from the engine, not the UI. `PollEvent` gains a `next_target` field (JSON `nextTarget`, only when set) that every tick fills from the same ranking and no-return bar the decision uses, so it stays right under `consume-first`, cooldown, quarantine and the API-key last resort. Existing event consumers are unaffected.
- Two engine changes that also apply to `cswap auto`: `stop()` now waits for a switch already under way to finish and refuses one that has not started yet, so a SIGTERM (or leaving live mode in the TUI) is never followed by one more switch. Events are no longer emitted while that lock is held.

## CPU while the card breathes

Two more commits, both perf. The first version of the animation pinned a core at 80% on a 257x90 terminal, and the fix for that lives in the auto-switch commit. What was left was Textual repainting the whole grid ten times a second because the grid was one Static.

- 854695f: `MetersGrid` is a grid container of one `MeterCard` per account. A breathing frame repaints one card, a handoff sweep two, a cursor move the old and the new card, none of them with a relayout. Compact mode is the same cards as one-line rows in a single column, hidden outside the cursor window instead of clipped. `meters_grid_text` no longer runs in production; the pilot tests keep it as the reference they compare the mounted cards against.
- 428ec21: the interval runs at 0.2 s while urgency is under 0.5 and at 0.1 s from there and during a sweep, with the phase step following the running period.

Headless with a predicted target, 257x90: 15% before the container, 6.7% after it at the full rate, 4.0% at rest. 120x40: 6.9% to 2.2% at rest. Idle stays under 1%.

## Compatibility

The default is `classic`, the existing watch view, so nothing changes for current users unless they set `ui.watchStyle` to `meters` or press `Ctrl+V`. The classic screen is untouched. The auto-switch keys exist only on the meters screen and do nothing until pressed.

## Testing

Tests cover the setting and its round-trip, both watch screens, the meter rendering in the dark and light palettes, the selection cursor following its account across a reorder, the live toggle, the next-target prediction under both strategies, the stop-mid-switch and deadlock cases, and the animation timer running only while there is something to animate. Full suite passes on Linux, macOS and Windows.

## Notes

This started as personal styling. I run `cswap watch` fullscreen on a small dedicated display and wanted a more glanceable, at-a-distance view, so I built the meter grid for that and have been using it this way. It is opt-in and leaves the default untouched, so I am opening it in case it is useful to anyone else, with no pressure to take it.

One open question: the active-account frame green and the maxed-bar reds are the same in both themes right now; I can add light-mode variants if you would prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,9 @@ Run `cswap` on its own (or `cswap tui`) for the full-screen dashboard: live usag
 
 <img src="assets/tui-watch.png" width="760" alt="cswap watch — live 5h/7d usage bars for every account, with reset times and the active account marked">
 
+The watch monitor defaults to the horizontal usage-bar cards above. For a compact, glanceable view — a responsive grid of vertical gradient meters — press **`Ctrl+V`** inside the watch screen to flip between the two live, or set the default up front with `cswap config set ui.watchStyle meters` (classic | meters).
+
+
 ### Refresh expired tokens
 
 If an account's token expires, log back into Claude Code with that account and re-run:
@@ -262,6 +265,7 @@ cswap config                              # list effective settings ("(default)"
 cswap config get autoswitch.threshold
 cswap config set autoswitch.threshold 80  # validated: rejects out-of-range values loudly
 cswap config set autoswitch.model Fable   # per-model switching (see "auto"); Fable,Opus for several
+cswap config set ui.watchStyle meters     # cswap watch layout: classic bars (default) or vertical meters
 cswap config unset autoswitch.threshold   # back to the default
 cswap config path                         # where settings.json lives
 ```
--- a/README.md
+++ b/README.md
@@ -161,6 +161,9 @@ Run `cswap` on its own (or `cswap tui`) for the full-screen dashboard: live usag
 
 <img src="assets/tui-watch.png" width="760" alt="cswap watch — live 5h/7d usage bars for every account, with reset times and the active account marked">
 
+The watch monitor defaults to the horizontal usage-bar cards above. For a compact, glanceable view — a responsive grid of vertical gradient meters — press **`Ctrl+V`** inside the watch screen to flip between the two live, or set the default up front with `cswap config set ui.watchStyle meters` (classic | meters).
+
+
 ### Refresh expired tokens
 
 If an account's token expires, log back into Claude Code with that account and re-run:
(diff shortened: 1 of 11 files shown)


hex:
Pushed a second commit on top, rebased on current main: the auto-switcher from inside the meters view. `a` runs it in dry-run and the card it would move to next gets a breathing frame, `L` goes live after the usual confirmation. The readout sits at the right end of the footer.



The engine side is small but not zero: `PollEvent` gets a `next_target` field (additive), and `stop()` now waits for a switch already in flight instead of returning while one is still running, which also covers the SIGTERM path in `cswap auto`. Description updated. CI green on all three jobs.

hex:
Two perf commits on top: 854695f makes the grid a container of per-account cards so a frame repaints one card, 428ec21 drops the breathing to 5 Hz while the switch is far off. 257x90 goes from 15% headless to 4.0% at rest. Description updated with the numbers.

PR #166, opened 2026-07-23
feat(auth): recover owner-held expired tokens

### LINKED ISSUE #165: Explicit recovery for owner-held expired OAuth access tokens
### What happened?

A managed account can remain temporarily unusable to automation when its OAuth access token expires while a live Claude Code process owns the credential.

This is the intentional safety behavior discussed in #62: `cswap list --json` reports `usageStatus: "token_expired"` rather than rotating an owner-held refresh token concurrently. That is the right behavior for a read-oriented inventory call.

The gap appears for consumers that admit accounts only from fresh `usageStatus: "ok"` evidence. Once the row becomes `token_expired`, they exclude that account from new launches and balancing. The account recovers only after an already-running owner happens to make another API request. In the observed case the account stayed excluded until one of two idle Claude sessions eventually issued a request; the next inventory observation then returned to `ok`.

### Proposed contract

Add an explicit mutation command, separate from `list`, `status`, watch, and autoswitch:

```text
cswap recover <num|email> --json
```

Narrow intended behavior:

- act only on a managed OAuth account whose access token is locally expired, whose refresh-token lineage is not quarantined, and whose actual credential-owning profile has an unambiguous live Claude owner;
- ask Claude Code in that exact profile to perform its own synchronized refresh through one bounded, noninteractive, no-tools, non-persistent minimal request;
- never directly rotate a live owner's refresh token;
- re-read and identity-check the credential after the request before reporting credential recovery; normal `list --json` collection remains the separate usage authority and retains its existing claim/fencing protocol;
- return a small versioned, PII-free JSON outcome suitable for automation;
- leave `relogin_required`, browser authorization, API-key/setup-token accounts, profile drift, and ambiguous ownership to the human;
- never run implicitly from read commands.

This would preserve #62's concurrency boundary while giving automation an intentional way to close the otherwise self-healing-but-unbounded credential gap. Consumers that route from usage should run their ordinary inventory command afterward and continue to trust only its result. It does spend one minimal Claude request when recovery is attempted; that should be explicit in the command documentation.

I have a focused implementation in progress and will link a PR here. I am happy to narrow the command or adjust the result vocabulary before review if another boundary fits the project better.

### LINKED ISSUE #187: bug: idle session expiry can falsely quarantine account backup
## Problem

Live reproduction:

- no live session process;

## Expected

possibilities:
## Summary

- add explicit `cswap recover <num|email> --json` for ordinary owner-held expired OAuth access tokens
- ask Claude Code in the exact live owner profile to perform its native synchronized refresh with one bounded, no-tools, non-persistent Haiku request
- re-read the same owner and require matching identity plus a non-expired OAuth credential before reporting recovery
- keep browser login, dead refresh-token lineages, API/setup tokens, profile drift, and ambiguous ownership human-only

Closes #165.

## Safety boundary

`list`, `status`, watch, auto, and the TUI remain read-oriented and never invoke recovery. The command does not directly rotate an owner-held refresh token, read a slot backup as owner truth, bootstrap a profile, or fetch/write usage outside the existing collector claim protocol.

Recovery is single-flight per slot and checks owner/profile/credential identity immediately before the request and again afterward. A manual `/login` after the final preflight is the residual external race; the post-run identity check detects drift. macOS Keychain failures fail closed instead of falling through to a potentially stale plaintext seed.

The canary runs with safe mode, no tools, no session persistence, one low-effort Haiku turn, sealed auth/provider environment, neutral cwd, ignored stdio, a hard deadline, and process-tree cleanup. It may be billable, so recovery is explicit and documented.

## JSON contract

```json
{"schemaVersion":1,"operation":"recover","accountNumber":2,"recoveryStatus":"recovered"}
```

`recoveryStatus` is one of `recovered`, `not_needed`, `retry_later`, or `human_required`. The payload contains no email, credential, path, PID, provider output, or exception text. `recovered` proves credential recovery only; callers continue to use normal `list --json` output as usage authority.

## Validation

- `uv run pytest tests/test_recovery.py tests/test_cli.py tests/test_session.py -q` — **275 passed**
- broad deterministic suite excluding the six known local macOS move/swap baseline failures — **1561 passed, 3 skipped, 6 deselected**
- full suite — **1561 passed, 3 skipped, 6 baseline failures**
- `git diff --check` — passed

No real account, credential, or Claude request is used by the tests.

## Follow-up

Also fixes #187: matching idle session profiles remain authoritative after access-token expiry, and stale backup `invalid_grant` can no longer falsely quarantine their newer lineage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,15 +163,21 @@ Run `cswap` on its own (or `cswap tui`) for the full-screen dashboard: live usag
 
 <img src="assets/tui-watch.png" width="760" alt="cswap watch — live 5h/7d usage bars for every account, with reset times and the active account marked">
 
-### Refresh expired tokens
+### Recover an expired token
 
-If an account's token expires, log back into Claude Code with that account and re-run:
+Recovery is explicit and automation-friendly:
 
 ```bash
-cswap add
+cswap recover 2 --json
 ```
 
-This will update the stored credentials without creating a duplicate.
+For an ordinary expired OAuth token, this asks the **live Claude Code profile that already owns the credential** to refresh it, then verifies that same profile still has the matching identity and a non-expired OAuth credential. It makes one minimal Haiku request, which may be billable. Recovery serializes commands for a slot and rereads the owner immediately before launch, but a manual `/login` or other external profile change after that final read can still race the canary; the mandatory post-run identity check detects the resulting drift. Recovery never runs implicitly from `list`, `status`, `watch`, `auto`, or the TUI, never restores a stale backup, and never starts browser login. `cswap list --json` performs the later usage observation through its normal collector protocol.
+
+The JSON `recoveryStatus` is one of `recovered`, `not_needed`, `retry_later`, or `human_required`. `human_required` is the safe limit for a missing or ambiguous owner, profile drift, setup tokens/API keys, missing or dead refresh tokens, and identity mismatches. In that case, log into Claude Code as the account yourself and re-run:
+
+```bash
+cswap add --slot 2
+```
 
 ### Other commands
 
@@ -182,6 +188,7 @@ cswap config                    # Show or edit settings (see Configuration below
 cswap list                      # Show all accounts with 5h/7d usage and reset times
 cswap list --token-status       # Add source-labelled OAuth token diagnostics
 cswap status                    # Show current account
+cswap recover 2 --json          # Explicitly recover an owner-held expired OAuth token
 cswap add --slot 3              # Add account to a specific slot (prompts before overwrite)
 cswap add --alias dev           # Add account and give it a short alias
 cswap remove 2                  # Remove an account
@@ -285,13 +292,14 @@ If an imported account is the one you're currently logged in as, activate the im
 
 ### JSON output for scripting
 
-Add `--json` to `list`, `status`, or `switch` to emit a single machine-readable JSON object on stdout (human-readable notices go to stderr). Useful for scripting auto-swap and quota tracking.
+Add `--json` to `list`, `status`, or `switch` to emit a single machine-readable JSON object on stdout (human-readable notices go to stderr). Explicit recovery is JSON-only. Useful for scripting auto-swap and quota tracking.
 
 ```bash
 cswap list --json                   # all accounts with usage/quota
 cswap status --json                 # current active account
 cswap switch --strategy best --json # switch, then report the result
 cswap switch 2 --json
+cswap recover 2 --json              # one PII-free recovery result
 ```
 
 <details>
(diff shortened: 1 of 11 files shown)


possibilities:
Live validation on Claude Code 2.1.218 exposed one missing process-launch input in the sealed canary environment: omitting `USER` and `LOGNAME` makes Claude's macOS Keychain account lookup report `Not logged in · Please run /login` even though the exact session-profile credential is readable and ordinarily refreshable. Commit `a0f66a4` preserves those two non-secret identity variables while continuing to scrub auth/provider/proxy overrides. Focused suite remains 275/275, the integrated stack is 703/703, and the live expired credential then returned `recovered` and produced a fresh normal usage observation.

possibilities:
Windows CI exposed one remaining portability defect in the test added by f7f13f3: `test_exact_argv_default_env_and_neutral_cwd` still unconditionally asserted the POSIX-only `USER`/`LOGNAME` canary environment. Production code already selects the Windows allowlist correctly. fb03f5a makes the assertion platform-specific (Windows verifies those variables are absent and uses `_WINDOWS_ENV_ALLOWLIST`; POSIX retains the Keychain assertions). Focused recovery/CLI/session suite passes locally: 278 passed. Published with an exact lease against f7f13f3; CI is rerunning.

possibilities:
Rebased onto current `main`, past #196. CI green on all three jobs.

Three notes from the rebase:

1. **The quarantine loop in `_collect_usage` needed a real merge, not a resolution.** #196's fingerprint-bound `_entry_token_dead` and strike-healing branch are kept intact; this PR's contribution now sits in a `_dead_lineage_sentinel` helper called from the dead branch. The two are complementary: #196 decides *whether* a stored lineage is dead, this decides *what quarantine that earns* — an idle slot whose session profile still holds a matching, expired-but-refreshable credential surfaces as recoverable rather than demanding a browser re-login.

2. **`scan_live_sessions`'s rename caught a live bug here** — thank you for the deliberate rename and the docstring explaining it. `recovery.py` was exactly the call site described: `bool(live_sessions_for(...))` would have read the new tuple as unconditionally truthy. It now uses `profile_is_quiescent`, which is also the correct semantics for this path: an unreadable session record means recovery declines to launch its canary rather than racing a live profile.

3. `README.md` follows `main`'s current structure (the Theme section is gone upstream).

Happy to split (1) out for separate review if you'd prefer to look at the interaction with #196 on its own.

possibilities:
Hi! Friendly bump on this one — I've just rebased onto current `main` (resolving against the profile-scoping changes from #237) and all three checks are green. If you have time to take a look I'd really appreciate it. Happy to make any changes you'd like.

Issue #117, opened 2026-07-09
Switch backs up the wrong account's credentials after a running Claude Code session writes its token back (slot poisoning: both slots become the same account)


gassao1998:
**Environment:** Windows 11 - claude-swap 0.17.1 (autoswitch via `cswap auto --once` every 5 min from a watchdog) - a long-running Claude Code session that stays up for days.

**Sequence**

1. Active = account 2; the long-lived Claude Code session holds account 2's OAuth in memory.
2. `cswap auto` switches to account 1 -> live credential files now hold account 1.
3. The running session later refreshes its OAuth and writes **account 2's** refreshed credentials back over the live files (Claude Code doesn't know about the switch).
4. Next switch: cswap logs `Backed up account 1` - but live is account 2's token, so **slot 1 now stores account 2's credentials and account 1's refresh token is lost.**

**Symptoms afterwards**

- `cswap --list` shows identical 5h/7d/per-model percentages *and* identical reset times for both accounts, climbing in lockstep while only one account is actually used
- `cswap --list --token-status` shows the same expiry minute for both
- Usage polling starts returning 429 for both accounts (the same org is now polled twice)
- Autoswitch flaps - with two identical accounts `--strategy best` has no stable winner (log below: 3 switches in 9 s)

**Log excerpt** - the `2 to 2` line is cswap itself noticing the external write-back:

```
2026-07-09 12:30:25 INFO - Switched from account 2 to 1
2026-07-09 13:06:04 INFO - Switched from account 2 to 2   <- live identity flipped back outside cswap
2026-07-09 13:06:11 INFO - Switched from account 2 to 1
2026-07-09 13:06:13 INFO - Switched from account 1 to 2   <- 3 switches in 9 s
```

**Suggestion**

Before backing up the outgoing account during a switch, verify the live credentials' identity actually matches the slot being backed up (e.g. compare `oauthAccount.accountUuid` in `~/.claude.json` with the slot's stored uuid). On mismatch, skip the backup and warn instead of overwriting - the overwrite silently destroys the slot's refresh token and turns two managed accounts into one, which also defeats autoswitch as a rate-limit escape hatch.

Related: #96 covered token rotation for session-mode profiles; this looks like the same root cause hitting the *main* profile via the backup-on-switch path (repro on 0.17.1; nothing in the 0.18.1 notes seems to cover this path).

[...]

Issue #62, opened 2026-06-21
Usage information unavailable without showing why.


Forne14:
Apologies if this is the wrong way to use this repo's issue tracker. 

I have an issue whereby sometimes one of my accounts usage information is unavailable to see. and the tool doesnt state why.

realiti4:
Hi! Access tokens usually expire after a couple of hours. Claude Code refreshes the expired ones with the refresh token when needed. Our tool refreshes expired access tokens the same way, so we can fetch usage data without relying on Claude Code to refresh them for us.

That said, our tool doesn't refresh the token for the active account, just to be safe and not interfere with Claude Code. We might be being too conservative here though, so I'll check whether it's actually fine. If it is, we can probably drop this guard. If not, I'll improve the message the tool shows so it explains why the usage is unavailable.

jparr721:
Separate thing I've hit. If you do say, `watch -n 1 cswap --list` it'll eventually rate limit you. Learned the hard way.

realiti4:
> Separate thing I've hit. If you do say, `watch -n 1 cswap --list` it'll eventually rate limit you. Learned the hard way.

That's weird `cswap --list` command has a 15-second cache specifically to prevent that. How long had you been running it before you got the rate limit warning? Maybe they have a broader rate limit as well.

realiti4:
I decided we should keep the current behavior but improve the warning instead. There's a really small chance, but when Claude Code is holding a lock on the credential file to refresh the token, we could end up refreshing at the same time. So I think leaving that to Claude Code for the active account is the right call.

Another improvement I've made: if there are no running instances, we now refresh expired tokens for the active account as well when list command is run, which should be safe. [#70](https://github.com/realiti4/claude-swap/pull/70)

jparr721:
To answer your question @realiti4 I ran it for maybe 20 minutes before it said usage unavailable. Note this was on v 0.10, I haven't done it in awhile

realiti4:
This behaviour is improved with #70, also releasing as [v0.15.0](https://github.com/realiti4/claude-swap/releases/tag/v0.15.0)

imloualvaro:
still same problem, even using latest build.

usage unavailable (http-429)

realiti4:
> still same problem, even using latest build.
> 
> usage unavailable (http-429)

Did this happen with `list` command or somewhere else like `watch` mode or `auto` mode?

How many accounts do you have? Have you been using the `list` command in a loop?

[...]

PR #167, opened 2026-07-23
fix(auth): refresh the active account's expired token under Claude Code's own lock protocol

### LINKED ISSUE #164: Shared accounts die simultaneously across machines: export/import + background refresh have no cross-machine token-rotation awareness
## Summary

Multiple **shared/pooled** Claude accounts (managed by more than one independent `cswap` installation) had their refresh tokens die within the same reporting window, on a machine that hadn't touched those accounts in 11-17 hours. Every *personally-owned* account on the same machine (never shared with another `cswap` install) stayed healthy. Root cause appears to be a structural gap: `cswap` has no concept of, or protection against, a second independent `cswap` installation managing the same underlying Anthropic account — which is our team's deliberate deployment model for shared department seats.

## Environment

- `cswap --version`: 0.22.0 (macOS)
- Deployment: one shared 9-seat `accounts.cswap` pool distributed to every developer's machine via a private channel; every machine runs its own `cswap auto` (LaunchAgent) against the *same* underlying Anthropic accounts for the shared seats, plus ad-hoc manual `cswap export` / `cswap import --force` hand-offs between teammates.

## What happened

1. On one machine ("Machine A"), 4 shared accounts (call them S1-S4) were last touched by *that machine's own* `cswap` 11-17 hours before the incident (confirmed via `claude-swap.log`: last `Backed up account N` / `Switched from … to N` entries for each).
2. Machine A's operator logged into two *other*, personally-owned accounts directly via Claude Code, then ran `cswap export backup.cswap` (default: all accounts) to hand those two logins to a teammate on Machine B.
3. Per `export_accounts` (`src/claude_swap/transfer.py`), **only the then-active account is freshened from the live vault** — every other account in the export, including S1-S4, is read from Machine A's own local backup store. Verified by diffing `sha256(refreshToken)` across several days of exports for S1-S4: every export has a *different* fingerprint, i.e. these tokens rotate frequently, so an 11-17h-old backup copy has a real chance of already being superseded.
4. The teammate on Machine B (who already had S1-S4 configured locally, from the same original pool distribution) ran the documented recipe for activating a received export — `cswap import backup.cswap --force` (required because the accounts already exist) then `cswap switch <n> --force` for the two accounts they actually wanted.
5. `--force` on `import_accounts` overwrites **every** account in the file unconditionally — there's no check comparing the incoming token's rotation-generation against what's already stored — so Machine B's own, independently-current copies of S1-S4 got silently regressed to Machine A's 11-17h-stale snapshot. Critically, `import_accounts` also unconditionally clears any dead-token quarantine flag on every slot it writes (even a forced overwrite of a healthy slot) — so this regression produces **no immediate symptom**; the slot reads as freshly healthy right after import.
6. A few hours later — not immediately — S1-S4 were reported `re-login needed — refresh token dead` (correctly detected by `cswap`'s own quarantine logic) on **both** machines, including Machine A, whose own copy had never been touched by the stale import at all. The delay between the stale overwrite (step 5) and the failure becoming visible is consistent with `cswap`'s adaptive polling cadence (infrequent checks for idle accounts) plus step 5's masking — the poisoned token sits quietly until something actually tries to use it again.

Point 6 (Machine A's own, never-touched-by-the-import copy also dying) is the key signal that this isn't just "Machine B clobbered itself" — presenting an already-superseded refresh token from a rotation lineage appears (consistent with standard OAuth refresh-token rotation/reuse-detection) to revoke the whole lineage, taking down whichever machine's copy was live too.

A second, independent way the same class of collision can happen with *no export ever involved*: `autoswitch.py::try_fetch_usage_for_account` opportunistically refreshes *inactive* accounts' tokens during background usage polling. Two machines' `cswap auto` pollers, both managing the same shared account, will each refresh it on their own schedule — same divergence, same eventual death, no export/import required.

## Why this matters for shared-seat teams

`cswap`'s hardening (fingerprint/"lineage" checks, "own-family" detection, discard-on-mid-refresh-race — all great, careful work) is entirely intra-machine: protecting one machine's store against races between `cswap` and a live Claude Code session on *that same machine*. None of it protects against a second, independently-live `cswap` installation holding the same account — which is exactly what happens whenever a team shares one Anthropic seat pool across multiple developers' laptops (our situation, and I'd guess not an unusual one — `cswap export`/`import` clearly anticipate handing accounts between machines, they just don't anticipate the *same* account being simultaneously live-managed on two).

## Suggested fixes

1. **`export_accounts`**: stamp each non-active account entry with its own local backup's last-refresh time (e.g. `"backedUpAt"` per account in the envelope), so an import has something to reason about.
2. **`import_accounts`**: when `--force` would overwrite an *existing, healthy* (non-quarantined) slot, compare the incoming refresh-token fingerprint/generation against what's already stored. If the existing token looks newer/different-lineage, refuse (or require an even stronger override) instead of silently regressing a live account. The README already flags "a stale export can carry an already-superseded token" as a known hazard — this would turn that caveat into actual enforcement.
3. **Docs**: explicitly call out in the "Backup & restore" section that running independent `cswap`/`cswap auto` against the *same* Anthropic account from more than one machine is unsupported and will eventually revoke that account's tokens. Nothing in the current README warns against this, and `export`/`import` read very naturally as "safe to hand an account between machines."

## Repro

I wasn't able to cleanly re-trigger this on demand (it needs two real machines and a live Anthropic OAuth exchange to actually observe the server-side revocation), but the mechanism above is derived directly from the source (`transfer.py::export_accounts` L182-195 read-non-active-from-backup behavior; `transfer.py::import_accounts` `--force` unconditional overwrite plus its unconditional `clear_dead_token` call around L510-516; `oauth.py::try_refresh_oauth_credentials` confirming rotation) against a real incident's logs and fingerprint-diffed (never raw-token) export snapshots. Happy to share the fingerprint table / log excerpts (with tokens redacted to sha256 prefixes) if useful — left out of this report to avoid dumping unnecessary detail.

Thanks for the tool overall — the amount of care already in the lineage/quarantine logic is obvious, this is very much an edge case outside the tool's evident target use case (one person, one machine, N of their own accounts) rather than a sloppy bug.

codeslake:
> Field-deployed (editable) on three machines with a behavior watch running throughout development; the soak has since shown the idle-expiry stall and the 56s bounce gone (Validation below). Grew out of the field measurements in #152's discussion.

## Problem

On an idle machine the active account's access token can expire with nobody refreshing it: cswap defers to the owner (#62), the idle owner never triggers its own refresh, and the daemon may not complete OAuth headless. The autoswitch engine then walks `USAGE_TOKEN_EXPIRED` → 30-min idle-hold → failover — and with the token dead, `fetch_oauth_profile` 401s, the outgoing classifier falls to `unresolved`, and a manual switch back to the account is bounced off it (measured: reverted 56s after the switch, repeatedly, on a healthy account).

The empty-token backups that finished the slot off have since been caught in the act during the live soak: on `invalid_grant`, **Claude Code itself wipes the live store's token fields** (observed directly on 2.1.181 — keychain item kept, `accessToken`/`refreshToken` emptied, no cswap write in the window), and each bounce's "pre-fix backup" then copies the wiped live credential over the slot's backup. The wipe is Claude Code's; the switch path is the propagation vector.

## Why touching an owner-held token is safe (verified in the 2.1.218 bundle)

Claude Code's refresh is built to *adopt* an external rotation, not collide with it: it takes **two** locks (`<config-home>/.oauth_refresh.lock` + legacy `~/.claude.lock`, both `stale:60000/update:5000`), re-reads the store under the lock and returns "refreshed" with no network call when the token already changed (`tengu_oauth_token_refresh_race_resolved`), and its 401 path adopts a changed store token before forcing re-auth. It holds both locks across its own token exchange.

cswap previously did neither half: `claude_locks.py` contended only the legacy lock with a 10s staleness (CC holds it legitimately up to 60s — a steal window), and the no-owner refresh consumed the grant **unlocked**, then could *discard* the rotated credential if an owner appeared — stranding a consumed generation as the live token.

**Reuse semantics, measured** (disposable lineage + live soak): re-presenting a consumed refresh token returns `invalid_grant` ("Refresh token not found or invalid") while the successor and sibling grants keep working — a rejection, not a family revocation. Rotation also invalidates the **predecessor access token** server-side immediately: machines holding a synced copy of a lineage started 401ing within ~2 minutes of a sibling machine's rotation, well before `expiresAt`. And the `invalid_grant` → wipe reaction is no longer bundle-inference — CC was observed emptying the live credential in the field (Problem above). "Persist the successor unconditionally" is therefore a hard invariant.

## Changes (implemented)

1. **`claude_locks.py`** (f705fca) — also acquire `.oauth_refresh.lock`, in CC's order (primary first; legacy contention rolls back the primary, so two waiters can't deadlock); credentials-lock staleness 10s → 60s; config lock unchanged at 10s; docstring corrected against the 2.1.218 bundle.
2. **`switcher.py`** (6c21286) — `_fetch_active_usage` now runs CC's own sequence: locks → re-read (adopt a fresher live credential without consuming) → provenance check (#117 lineage guard unchanged) → POST → **persist unconditionally** (backup first, then active store) → release. The discard-after-consume branch is gone. Owner-present is no longer a skip: a concurrently refreshing CC serializes on the locks and adopts the rotation. The owned+expired static sentinel is removed so the collect pass reaches the refresh; the one static remnant (active row inside a failure backoff — fetch unreachable that tick) still surfaces `token expired` so the Finding-2 idle-hold behavior is preserved. `USAGE_TOKEN_EXPIRED` now means: unattributable lineage, dead lineage, lock contention, or failed persist — the states that genuinely need the autoswitch ladder.
3. **Wipe-propagation guard** (035e409) — the outgoing classifier gains a `wiped` kind: an OAuth blob whose token fields are all empty (CC's `invalid_grant` reaction) is never written into any slot — config backup only, backup credential kept as the recovery path, user pointed at `cswap add`. The adopt branch likewise requires a full token pair before resyncing (a wiped blob can carry a future `expiresAt` and masquerade as non-expired). This closes the destruction chain end-to-end: CC may still wipe the live store, but the slot backup survives and `cswap switch --force` / #166's `recover` restores service.
4. **Rotation-before-collection** (ef27749, from review) — a CC rotation completing during normal use left the slot backup on the consumed predecessor; at the next expiry the recovery branch POSTed that dead grant and quarantined a healthy slot. Two-sided per the review: the fresh-token fast path resyncs the backup whenever the served credential's lineage differs from it (the adopt-branch resync extended to an already-completed rotation — same identity/full-pair/lock guards, best-effort), and the `live == creds` recovery arm discriminates by generation: backup newer → restore (as before); live newer → live's refresh token is the valid successor, POST it, never the backup's consumed grant. Regression tests cover both timelines plus the tolerated backup-write-failure drift.
5. **Review hardening** (dd44022 → 8fb1554, three external review rounds to convergence) — the under-lock identity guard runs unconditionally and compares `(email, organizationUuid)`; a live-store read *error* defers instead of masquerading as an empty store; a stranded live store (successor persisted to the backup but the live write failed) is **restored from the backup** on the next pass — attribution decides *how* to recover (live's grant vs the slot's own backup grant), never whether to give up; a locally-valid token the server 401s (revoked out-of-band — measured on synced lineages) triggers the locked refresh instead of hours of backoff, and every pre-consumption defer surfaces that 401 to the store; `no_refresh_token` joins the permanent-auth errors (quarantine, not an eternal retry loop); the expired-active sentinel overlay covers every fetch gate (backoff, a concurrent collector's claim, poll-plan gating) so a claim race can't reset the auto engine's idle-hold anchor; and the whole path honors `_fetch_account_usage`'s never-raises contract.

## Not claimed

- Does **not** fix #164's import/export mechanisms — it removes the steady-state *motivation* for syncing; provisioning, dead-lineage repair, and the ~4-week refresh-token expiry remain sync/login territory.
- Does **not** replace #166 — `cswap recover` stays the explicit manual tool; this covers the automatic path it leaves out.
- Does **not** stop Claude Code's own `invalid_grant` wipe of the live store — that's CC-side behavior. What this PR does guarantee (change 3) is that the wipe can no longer destroy the slot's backup, so recovery is a switch/recover away instead of a re-login.
- Bridge-connected sessions already run a proactive refresh timer; this matters mainly for machines without one.

## Validation

- `tests/test_claude_locks.py` — the two-lock protocol: order, rollback on legacy contention, 60s staleness (30s-old lock NOT stolen, 70s-old is), config lock unchanged.
- `tests/test_switcher.py::TestActiveAccountRefresh` — refresh under held locks (asserted during the POST), owner no longer blocks, adopt-without-consume on a fresher live read, provenance guard intact, invalid_grant → sentinel, lock timeout → clean deferral, backup survives a failing live write, non-expired path untouched.
- Wipe guard: a wiped live blob never reaches a token-bearing backup (switch path), never gets adopted (refresh path); byte-identical wiped live+backup stays quiet (damage predates the switch).
- Full suite: 1546 passed; the 11 remaining failures are the pre-existing local ANSI-assert baseline (also noted in #166), reproduced unchanged on a clean `main` checkout.
- **Live soak, day 1**: a lineage that had been synced across all three machines cascaded naturally — the machine whose CC rotated kept working; every synced sibling copy died at once (401 pre-`expiresAt`, then `invalid_grant` on refresh). The fixed build absorbed it with **zero oscillation**: one clean failover on one Mac, honest quarantine on the other ("re-login needed" / "no credentials", autoswitch holding on "no-comparison" instead of bouncing), strike-based backoff on Linux, and no backup destroyed anywhere. Incidentally a live demonstration of why per-machine independent grants beat cross-machine syncing — the steady state this PR makes viable.
- **Live soak, day 2** (independent grants everywhere, ten expiries across three machines): every expiry resolved without an error, a bounce, or a stall. Idle-slot backups rotated cleanly through three consecutive cycles of the same lineage; expiries under an *active* CC were rotated by CC and adopted by cswap with the slot backup resynced (the adopt-resync branch fired in the field exactly as designed); one useful negative: Claude Code's supervisor daemon runs its own proactive refresh timer while it holds a *valid* token, so cswap's consume branch is the machine's safety net for precisely the states that daemon can't handle — dead access token (`auth_required`, no timer armed), no daemon, or a headless OAuth it can't complete. That is the original incident's shape, and it is the case this PR fixes.
- Threshold autoswitch verified live along the way: two proactive 90%-threshold switches with the outgoing account backed up correctly, and the at-limit escape logic (a saturated candidate is refused below 100% active usage, accepted at it) fired in the field — the active account hit 100% on its 7-day window and two machines independently switched to a 90%-utilized candidate within a minute, no flap, both slot backups intact.

Refs: #62, #164, #165, #166 · field data: #152 (comments)


realiti4:
Thanks a lot for this. It closes a real gap that's been there for some time and thanks for digging into recent Claude Code and building on its own lock protocol. My agent also checked the lock-protocol claims against the 2.1.220 binary and they all hold.

Could we adjust a couple of things before merge though?

A rotation that completed before collection leaves the backup stale, and the recovery branch then consumes it. Sequence: CC rotates A→B during normal use → the fresh-token fast path serves usage off B but never resyncs the slot backup (still A, consumed) → at B's expiry the `live == creds` arm picks the backup and POSTs A's dead grant → `invalid_grant` → with `AUTH_DEAD_STRIKES = 1`, a healthy slot gets a false "re-login needed" quarantine while CC keeps working on B. Since the backup only resyncs at add/switch/adopt, and CC rotates roughly daily under use, this is probably the common state at exactly the expiry-on-idle moment the PR targets. I reproduced it on `8fb1554` with a two-phase characterization test (happy to share it).

What I think would fix it (open to a different approach if you see a cleaner one):

1. Resync the backup when a fresh live credential differs from the backup's lineage — basically your adopt-branch resync (identity re-check + full token pair, under the locks) extended to a rotation that already completed.
2. In the recovery branch, prefer the newer-`expiresAt` lineage for the POST — that's the local discriminator between "stranded consumed generation" (backup newer → restore, as you have it) and "CC rotated earlier" (live newer → live's refresh token is the valid one).
3. Regression tests for both timelines: rotation-before-collection carried through the next expiry, and the next pass after the tolerated "backup write failed, live write succeeded" case (same drift shape).

Also needs a rebase now that #160/#182 landed — it's mechanical: one conflict hunk in the collector (keep main's claims-dict `reserve` branches, key your sentinel loop on `claims`), and #182's new repair test patches `_active_cc_running`, which this PR deletes — that patch entry can just be dropped.

Everything else looks good to me — after this I'm happy to merge.

codeslake:
Thanks for the sharp catch — reproduced and fixed in ef27749, following your suggested split: the fresh-token fast path now resyncs the slot backup whenever the served credential's lineage differs from it (the adopt-branch resync extended to an already-completed rotation, same identity/full-pair/lock guards), and the recovery branch discriminates by `expiresAt` — backup newer → restore as before, live newer → live's refresh token is the valid successor, so the backup's consumed grant is never POSTed. Regression tests cover both timelines plus the tolerated backup-write-failure drift.

Rebased onto main (#160/#182) — the collector conflict resolved as you described, and #182's `_active_cc_running` patch entry dropped.

A follow-up commit (cf2e408) from our own review pass on the rebase: the locked refresh's POST is now bounded at 6s so a slow network can't starve a concurrent `cswap switch`'s 10s lock acquire, the claim-liveness predicate and the under-lock identity re-check are deduped into single helpers, and the warn-once set is keyed by (slot, email) so slot reuse warns afresh.

Happy to take your characterization test as well if you'd like it in the suite.

realiti4:
Thanks a lot for the quick adjustments, and again for the thoughtful work. Everything looks good, I've merged it.

codeslake:
Thanks @realiti4! If you have anything for me to look into, please let me know!

PR #168, opened 2026-07-23
feat(menubar): start at login via a launchd LaunchAgent

### LINKED ISSUE #141: Docs/feat: run the menu bar as a persistent background service (launchd LaunchAgent), not tied to a terminal
## Problem

`cswap menubar` runs in the foreground, so it's tied to the terminal that launched it — close the terminal (or log out) and the menu bar disappears. There's no documented way to run it as a persistent background app.

The README's [Menu bar section](https://github.com/realiti4/claude-swap#menu-bar-macos) covers install + `cswap menubar` but stops there. (`cswap auto` already gets cron/systemd guidance; the menu bar has no equivalent.)

## Suggestion

Document a **launchd LaunchAgent** — the native macOS way to keep a process running without a terminal. It survives logout/reboot, auto-restarts on crash, and needs no `.app` bundle.

`~/Library/LaunchAgents/com.cswap.menubar.plist`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>Label</key>
    <string>com.cswap.menubar</string>
    <key>ProgramArguments</key>
    <array>
        <string>/path/to/cswap</string>
        <string>menubar</string>
    </array>
    <key>RunAtLoad</key>
    <true/>
    <key>KeepAlive</key>
    <true/>
    <key>StandardOutPath</key>
    <string>/tmp/cswap-menubar.log</string>
    <key>StandardErrorPath</key>
    <string>/tmp/cswap-menubar.err</string>
</dict>
</plist>
```

Replace `/path/to/cswap` with the output of `command -v cswap`, then:

```bash
launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.cswap.menubar.plist   # start + enable at login
launchctl bootout   gui/$(id -u)/com.cswap.menubar                                # stop + disable
launchctl kickstart -k gui/$(id -u)/com.cswap.menubar                             # restart (e.g. after a cswap upgrade)
```

Tested on macOS (Darwin 25.5.0) with a `uv tool install`ed cswap — starts at login, no terminal, comes back after a crash.

## Optionally

Could go a step further and add a `cswap menubar --install-service` / `--uninstall-service` that writes/loads this plist (auto-filling the binary path), so users don't hand-edit XML. Happy to open a PR if that's wanted.

mrvnklm:
Closes #141.

`cswap menubar` blocks in the foreground, so the menu bar dies with its terminal and has to be relaunched by hand after every login. This adds a **Settings → Start at login** toggle, plus a headless `cswap autostart [on|off]` — setting up a second Mac over SSH shouldn't require hand-writing a plist.

```
cswap autostart          # on / off, and where the LaunchAgent lives
cswap autostart on       # takes effect at the next login
cswap autostart off
cswap autostart --json   # scriptable
```

Both paths share one implementation (`src/claude_swap/autostart.py`), so the GUI and the CLI can't drift.

### Design notes

- **Enabling only writes the plist**, deliberately not `launchctl bootstrap`. Both entry points are reached from a process that already *is* the menu bar (or a CLI call that shouldn't spawn a GUI), so loading the job now would start a second copy. launchd loads `~/Library/LaunchAgents` at login anyway.
- **Disabling only removes it**, and boots the job out *unless this process is that job* — toggling autostart off from inside the launchd-started app must not kill the app the user is still clicking in. Detected via `XPC_SERVICE_NAME`, which launchd sets to the job label.
- `KeepAlive` is scoped to `SuccessfulExit: false`: restart a crash, never resurrect a deliberate Quit from the menu.
- **Survives reinstalls.** A `uv tool upgrade` or a move to another Python relocates the console script, which would leave autostart silently dead; the menu bar re-points a stale plist on its next launch.
- **State lives in the plist, not `menubar_settings.json`** — launchd is the source of truth, and a settings copy could disagree with it after a manual edit.
- The agent gets an explicit `PATH` (tool bin dir, Homebrew, system) since it inherits nothing from a login shell — without it `claude` and `security` aren't found.

### Tests

31 new tests (`tests/test_autostart.py`, `TestAutostartCommand` in `tests/test_cli.py`). The plist builder and path helpers are pure, so nothing touches the real LaunchAgents directory or launchd; the single `launchctl` call is asserted through a patched `subprocess.run`.

Full suite: 1554 passed. The 6 failures in `test_swap_accounts.py` / `test_move_accounts.py` are pre-existing on `main` at c89e6f9 — I verified they fail identically without this branch applied.

### Verified on real hardware

Running on two Macs (macOS 15.2 and 26.5.1), one of them set up entirely over SSH via `cswap autostart on`. Worth noting for #141: the caveat at `menubar.py:882-889` about launchd blocking Keychain access did **not** reproduce on either machine — a `security find-generic-password -w` from inside a `gui/<uid>` LaunchAgent returned the credential normally. The one machine that did fail returned `rc=36` purely because its login keychain was locked, which is a different (and self-explanatory) condition.

### Relationship to #102

#102 also proposes a LaunchAgent (`--install-menubar`), bundled with an unrelated menu-detail change, and is currently conflicting. This differs in the parts that bite in practice: it sets `PATH`/`KeepAlive`/`ProcessType`, refuses to duplicate a running instance, won't boot itself out, self-heals after reinstalls, and exposes the toggle in the menu where users look for it. Happy to defer or merge ideas if you prefer that PR's direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ cswap alias                     # List all aliases
 cswap move 2 1                  # Assign an account to a slot (relocates to an empty slot, swaps if taken)
 cswap tui                       # Interactive dashboard (also: bare `cswap`)
 cswap watch                     # Dashboard, opened on the live watch page
+cswap autostart on              # Start the macOS menu bar at login (bare: status; `off` to remove)
 cswap upgrade                   # Upgrade claude-swap to the latest version
 cswap purge                     # Remove all claude-swap data
 ```
@@ -249,6 +250,16 @@ cswap menubar
 ```
 
 Shows every account's 5h / 7d / spend usage and switches with a click (specific / rotate / best / next-available), plus the TUI's add / disable-enable / remove / refresh actions. Enable *Settings → Auto-switch accounts* to run the same engine as [`cswap auto`](#automatic-switching) in the background; it shares the `autoswitch.*` settings, so the menu bar and CLI stay in sync. Off until you turn it on.
+
+`cswap menubar` runs in the foreground, so it stops when its terminal closes. *Settings → Start at login* keeps it around: it writes a launchd LaunchAgent (`~/Library/LaunchAgents/com.claude-swap.menubar.plist`) that starts the menu bar at every login and restarts it if it ever crashes — but never after you quit it from the menu. The same toggle is available headlessly, which is the easier route when setting up a second Mac over SSH:
+
+```bash
+cswap autostart          # on / off, and where the LaunchAgent lives
+cswap autostart on       # takes effect at the next login
+cswap autostart off
+```
+
+Reinstalling (`uv tool upgrade`, a move to another Python) relocates the executable the LaunchAgent points at; the menu bar re-points it on its next launch, so autostart survives updates.
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -198,6 +198,7 @@ cswap unclaimed                 # List stashed credential entries (slot + why th
 cswap unclaimed --purge ID      # Drop one (deletes its bytes; recover with /login + `cswap add`)
 cswap tui                       # Interactive dashboard (also: bare `cswap`)
 cswap watch                     # Dashboard, opened on the live watch page
+cswap autostart on              # Start the macOS menu bar at login (bare: status; `off` to remove)
 cswap upgrade                   # Upgrade claude-swap to the latest version
 cswap purge                     # Remove all claude-swap data
 ```
@@ -245,6 +246,16 @@ cswap menubar
 ```
 
 Shows every account's 5h / 7d / spend usage and switches with a click (specific / rotate / best / next-available), plus the TUI's add / disable-enable / remove / refresh actions. Enable *Settings → Auto-switch accounts* to run the same engine as [`cswap auto`](#automatic-switching) in the background; it shares the `autoswitch.*` settings, so the menu bar and CLI stay in sync. Off until you turn it on.
+
+`cswap menubar` runs in the foreground, so it stops when its terminal closes. *Settings → Start at login* keeps it around: it writes a launchd LaunchAgent (`~/Library/LaunchAgents/com.claude-swap.menubar.plist`) that starts the menu bar at every login and restarts it if it ever crashes — but never after you quit it from the menu. The same toggle is available headlessly, which is the easier route when setting up a second Mac over SSH:
+
+```bash
+cswap autostart          # on / off, and where the LaunchAgent lives
+cswap autostart on       # takes effect at the next login
+cswap autostart off
+```
+
+Reinstalling (`uv tool upgrade`, a move to another Python) relocates the executable the LaunchAgent points at; the menu bar re-points it on its next launch, so autostart survives updates.
 
 </details>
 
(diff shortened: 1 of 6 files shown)


mrvnklm:
Closing in favour of #258, now merged. The launchd half is the same job done the better way round — installing bootstraps the job instead of waiting for the next login — and this branch now conflicts with `main` in `README.md` and `cli.py`, so there is nothing here worth rebasing.

Two things have no equivalent on `main`. I would rather send them than let them go, and they are separate concerns, so separate PRs:

- **"Start at login" in the menu bar's Settings menu.** Not a wrapper over `--install-service`: `uninstall()` boots out unconditionally, which is right for a CLI-only path but would kill the process running the menu if it were called from the app itself. The guard is `XPC_SERVICE_NAME == LABEL` — launchd sets it in the job's own environment and a shell-started process does not have it — so it has to land with the toggle rather than after it.
- **A staleness check, so a service whose pinned program moved is repaired instead of staying dead.** Narrower here than it was on this branch: pinning the unresolved `~/.local/bin/cswap` already survives a plain `uv tool upgrade`, which is the case I wrote it for. What is left is an install that pinned a venv or checkout path, or a move to another Python — and the repair can only fire on a hand-started `cswap menubar`, since inside the launchd-started process `resolve_program()` returns the pinned path by construction. Worth a net, but a small one.

The rest I would leave. The `--json` status would be a widening of the documented v1 output surface rather than part of this. `ThrottleInterval: 10` I would not port at all — that is launchd's own default respawn interval, the same 10s @ardabayhan's two-agent measurement showed.

@ardabayhan — you flagged `is_stale()` as possibly deserving its own PR yourself; happy to hand it over instead if you would rather have it. Thanks for the quick turnaround on the three findings.

ardabayhan:
@mrvnklm both of these look like yours to me. You wrote them, and your narrowing of `is_stale()` here is sharper than anything I would have sent: I had not thought about the repair only being able to fire on a hand-started `cswap menubar`, since `resolve_program()` inside the launchd process returns the pinned path by construction.

That said, I would gladly take it on if you would rather hand it over. Truly either way — you know that code better than I do, so whichever you are more comfortable with is the right call.

On the `XPC_SERVICE_NAME` guard, I can confirm it from the other side: `uninstall()` boots out unconditionally, so a Settings toggle calling it from inside the menu process would be killing its own host. Agreed on `--json` and `ThrottleInterval` too.

One thing worth knowing before you touch the menu bar code: #310. On Python 3.14 rumps never draws the status item — no error, process healthy, menu bar simply empty. `uv tool install` picks 3.14 by default now, so a fresh menubar install today is silently dead; `--python 3.13` is the workaround. A minimal `rumps.App("TEST").run()` reproduces it with none of this project's code involved. Flagging it so you do not lose an afternoon to it like I did.

Thanks for the review — the KeepAlive one was a real bug.

mrvnklm:
Taking both then — thank you, and for confirming the bootout from the other side.

On #310: it would have caught me. Both my machines run the menu bar on 3.13.13, but only because they were installed before Homebrew moved `python3` to 3.14 — `uv python list` here now puts 3.14.5 first, so a reinstall on either would have picked it up and I would have been staring at an empty menu bar with a healthy process. Adding that as a second data point in case the "fresh install today is silently dead" framing needs one: it is also *reinstall* today, on a machine that was fine yesterday.

PR #169, opened 2026-07-23
feat(json): expose account capacity metadata

possibilities:
## Summary

Expose two additive, optional account-capacity fields on `list --json` rows:

- `subscriptionType`: admitted only for exact OAuth values `pro` and `max`
- `rateLimitMultiplier`: a numeric projection of the three known `rateLimitTier` values (`1`, `5`, or `20`)

Both fields come only from the credential already selected for that row. They are independent and omitted for unknown, malformed, API-key, or absent values.

## Why

Machine consumers can compare heterogeneous Claude subscriptions without inferring plan capacity from email, organization display text, or quota percentages. The existing additive schema contract makes omission the safe compatibility behavior.

## Safety

The credential object never enters the payload. Tests serialize the complete response and assert that synthetic access-token material is absent. No fetch, refresh, switching, or credential write behavior changes.

## Validation

- `uv run pytest tests/test_json_output.py tests/test_switcher.py tests/test_oauth.py -q` — **475 passed**
- upstream CI: **1,596 passed, 3 skipped** on Linux; **1,558 passed, 41 skipped** on Windows; macOS Keychain job green
- `git diff --check` — passed


kmizzi:
Real-world +1 from a machine consumer. We drive an internal cost/utilization dashboard off `cswap list --json` across 6 Claude seats, and this is exactly the gap we hit: a raw `sevenDay` of 90% means very different absolute headroom on a `default_claude_max_5x` vs a `20x` seat, so comparing utilization across heterogeneous seats — or judging whether a seat is over/under-provisioned — isn't possible from the percentages alone. `rateLimitMultiplier` makes the % normalizable, and `subscriptionType` lets us stop inferring the plan from the email/org string (we hardcode it in a side file today).

The additive-omission contract works well for us too — we treat missing fields as "unknown" and fall back, so this is a clean drop-in.

One small doc thought: since consumers will be tempted to compare raw `sevenDay` across seats, a one-line note that the percentage is relative to each account's own tier (hence the multiplier) might save people a footgun. Would love to see this land.

possibilities:
@kmizzi Thanks for the concrete dashboard example — it’s a helpful explanation of why these fields matter in practice.

I’ve incorporated the documentation suggestion. The JSON docs now state explicitly that percentages such as `sevenDay.pct` are relative to each account’s own tier, and that consumers comparing absolute capacity across heterogeneous seats should normalize with `rateLimitMultiplier` rather than treating equal percentages as equal quota.

For completeness, the pushed implementation provides:

- optional `subscriptionType` only for exact `pro` / `max` OAuth values;
- optional numeric `rateLimitMultiplier` mappings of `1`, `5`, and `20` for the three known tiers;
- independent omission of unknown, malformed, absent, and API-key values;
- row-local credential provenance, with full-payload assertions that access-token material never reaches JSON output.

During the follow-up review I also hardened the omission contract for syntactically valid but non-object credential JSON (`null`, arrays, strings, and numbers), which previously could raise while serializing a row instead of simply omitting metadata.

The branch is rebased on current `main` at `a4df2e2`; CI is green (Linux **1,596 passed, 3 skipped**; Windows **1,558 passed, 41 skipped**; macOS Keychain passed). Thanks again for the useful consumer-side feedback.

possibilities:
Rebased onto current `main` (post-#196). No conflicts; CI green on all three jobs.

Still applies cleanly — `json_output.py` has no `subscriptionType`/`rateLimitMultiplier` equivalent upstream, so the gap this closes is unchanged.

possibilities:
Hi! Friendly bump on this one — I've just rebased onto current `main` and all three checks are green. If you have time to take a look I'd really appreciate it. Happy to make any changes you'd like.

PR #170, opened 2026-07-23
feat(json): expose display-grade last-good usage

possibilities:
## Summary

Keep stale usage non-actionable while preserving it for human display:

- decision-trusted data continues to use `usage`, `usageFetchedAt`, and `usageAgeSeconds`
- when data is too old to drive a decision, `usageStatus` remains `unavailable` and `usage` remains `null`
- additive `lastGoodUsage`, `lastGoodFetchedAt`, and `lastGoodAgeSeconds` retain the display-only snapshot
- the same contract applies to list rows and the managed active row from `status --json`

This lets conservative automation fail closed without forcing UIs to discard the last useful context.

## Safety

Last-good fields never make an account actionable: `usageStatus` and `usage` remain authoritative. Fresh decision-grade rows do not emit duplicate last-good fields.

## Validation

- `uv run pytest tests/test_json_output.py tests/test_switcher.py tests/test_usage_store.py -q` — **429 passed**
- `git diff --check` — passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,7 +319,7 @@ cswap switch 2 --json
 
 Every payload carries a `schemaVersion` (currently `1`); on a handled error stdout is `{"schemaVersion":1,"error":{...}}` with a non-zero exit code. `--switch`/`--switch-to` report `{"switched": true|false, "from": …, "to": …, "reason": …}`.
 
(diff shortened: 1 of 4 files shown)


realiti4:
Merged it, thank you!

PR #171, opened 2026-07-23
feat(menubar): aligned account rows — columns, bars, binding-limit-first

mrvnklm:
The compact account row formats each account independently, so nothing lines up. Under a proportional menu font, comparing two accounts' 7d usage means reading every row — and the window labels (`5h`, `7d`, `Fable`, `$`) are repeated on every single line.

This adds three opt-in layouts under **Settings → Account rows**, sharing one column grid across all accounts. **Compact stays the default**, so no existing menu changes without the user asking.

| | |
|---|---|
| **Columns** | one column per usage window; the window names move into a dimmed header row |
| **Columns with bars** | the same grid with a drawn progress bar in front of each number |
| **Binding limit first** | leads with the window actually closest to blocking you, rest dimmed as context |

Plus **Show reset times**, an independent toggle adding a second dimmed line with each window's countdown, aligned under its own percentage.

Percentages turn orange past 80% and red past 95%. Only the constrained end is tinted, so a pool of idle accounts stays monochrome and colour means something.

### Why it's built this way

- **Alignment is AppKit's, not padding.** One shared `NSParagraphStyle` whose tab stops are measured from the widest cell per column. Every measurement is taken in the font the text is actually drawn in, and every gap is a multiple of that font's space width — so the grid follows the system font size rather than assuming one screen's metrics. Verified by rendering at 11pt, 13pt and 18pt.
- **Bars are drawn images, not block glyphs.** `▁ █ ▌` have no coverage in the menu font, fall back to whatever face does, and come out at inconsistent widths — and `▁` draws a hairline on the baseline, so an empty bar reads as an underscore. The bar is sized from the font's point size, so it scales too.
- **Percentages in the bars layout are padded with FIGURE SPACE.** Each cell is right-aligned on its tab stop, so a wider number ("37%" vs "0%") would push its bar left and leave the bars' left edges ragged.
- **Window columns share one width.** A `0%` column beside a `100%!` one otherwise collapses and breaks the vertical rhythm.
- **Long account names are truncated, domain first.** The name column is as wide as its widest entry, so one long address would push every number on every row to the right — in testing it doubled the menu width.
- **Sentinel usage rides in the name cell.** `re-login needed` dropped into a right-aligned percentage column runs backwards across the row; `4  d@x.de — re-login needed` reads correctly.
- **Spend is never tinted and carries no at-limit marker.** An exhausted budget caps cost, it doesn't block a request; flagging it like a maxed rate-limit window would misread.

### Testing

The layout lives in `build_attributed_rows()`, which takes `AppKit` as an argument the same way the existing glue takes `rumps`. That keeps `menubar.py` import-safe and unit-testable without pyobjc — and it let me render the rows to PNG for visual review without opening a menu, including cases this machine doesn't produce (three per-model windows, 60-character addresses, everything at 100%, sentinel and disabled rows side by side).

52 new tests. Full suite: 1557 passed. The 6 failures in `test_swap_accounts.py` / `test_move_accounts.py` are pre-existing on `main` at c89e6f9 — they fail identically without this branch applied.

All four layouts plus the reset toggle were also exercised in the running app on macOS 15.2.

Independent of #168 (both touch `menubar.py`, but different regions; happy to rebase whichever lands second).

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,14 @@ cswap menubar
 
 Shows every account's 5h / 7d / spend usage and switches with a click (specific / rotate / best / next-available), plus the TUI's add / disable-enable / remove / refresh actions. Enable *Settings → Auto-switch accounts* to run the same engine as [`cswap auto`](#automatic-switching) in the background; it shares the `autoswitch.*` settings, so the menu bar and CLI stay in sync. Off until you turn it on.
 
+*Settings → Account rows* picks how each account is drawn. **Compact** (the default) is one free-text line per account. The other three lay every account on one shared column grid, so the same window sits at the same place on every row and you can compare down a column instead of reading across:
+
+- **Columns** — one column per usage window, with the window names in a dimmed header instead of repeated on every row.
+- **Columns with bars** — the same grid with a small progress bar in front of each number.
+- **Binding limit first** — leads with the window that's actually closest to blocking you, and dims the rest as context.
+
+Percentages turn orange past 80% and red past 95%, so a pool of idle accounts stays monochrome and colour means something. Spend is never coloured — an exhausted budget caps cost, it doesn't block a request. *Show reset times* adds a second, dimmed line with each window's countdown, aligned under its own percentage.
+
 </details>
 
 ## Advanced
--- a/README.md
+++ b/README.md
@@ -258,6 +258,14 @@ cswap menubar --uninstall-service   # stop it and remove the plist
 
 The agent lives at `~/Library/LaunchAgents/com.cswap.menubar.plist` and logs to `~/Library/Logs/com.cswap.menubar.{log,err}`. It pins the `cswap` console script, whose path survives an upgrade — but the running process keeps the old build until it restarts, so after `cswap upgrade` either re-run `--install-service` or `launchctl kickstart -k gui/$(id -u)/com.cswap.menubar`.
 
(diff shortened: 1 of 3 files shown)


mrvnklm:
Rebased onto current `main` — the only conflict was one README section sitting next to the one #258 added; both kept.

Two fixes on top, both found while running this on a pool of eight accounts:

**One expired account was stretching the whole menu.** The sentinel notes carry a remedy as well as a state, and `re-login needed — refresh token dead; log in with Claude Code, then run: cswap add` renders 614pt wide against 314pt for every aligned account row. A menu is as wide as its widest item, so that single row nearly doubled it and left the usage columns marooned against the far edge with a lake of empty space in between. The menu now shows the state and `cswap list` keeps the full note, where one long line costs nothing and the remedy is what the reader came for. Cutting it takes the row to 201pt — narrower than the table, so it stops setting the width at all.

**The active-row checkmark floated between the lines.** `NSMenuItem.state` centres its image vertically over the whole item, and a grid row is two lines tall once reset countdowns are shown, so the mark sat in the gap between the name and the countdowns, aligned with neither. The active row is marked by weight inside the text instead. Weight rather than colour: an accent-coloured row reads as a link, and it would compete with the warning and critical tints on the percentages, which are the readings that matter. Compact rows keep the checkmark, where it is conventional and correctly placed.

Also dropped the repeated `resets in` caption. It printed once per account, saying the same thing each time, and it sat in the name column while its values sat under the window columns — so for every account whose 5h window was idle it pointed across an empty gap at nothing. The countdown now sits directly under the percentage it belongs to, and the header above already names the window.

`2225 passed, 3 skipped`. Happy to split the three layouts into separate PRs if that reviews better.

PR #172, opened 2026-07-23
fix(tui): observe external usage refreshes

possibilities:
## Summary

`cswap watch` currently drops every three-second tick while a fetch-enabled snapshot worker is running. If another process refreshes a session credential and publishes newer shared usage during that bounded fetch, the watch frame cannot observe it until its own worker returns. Transient `token expired` overlays can also disappear on a same-measurement pass or be repainted by a late worker.

This change:

- keeps one fetch-enabled worker and one strictly store-only repaint worker as independent single-flight lanes;
- uses the existing `fetch=set()` path for repainting, so provider cadence is unchanged;
- reconciles snapshots so a token-expired overlay survives until a newer measurement or superseding state arrives;
- rejects late per-account measurement regressions in the TUI-local snapshot source; and
- shows snapshot age and a long-running refresh indication in the watch title.

## Safety

The new concurrent lane is read-only and network-ineligible. No credential ownership, usage-store persistence, poll planning, routing, or provider request cadence changes. Network workers are never cancelled or duplicated.

This composes with the fetch-lease outcome fencing in #160: that PR protects the shared store from late writers, while this PR protects the watch frame and lets it observe newer shared state promptly.

## Validation

- `NO_COLOR=1 uv run pytest tests/test_tui.py -q` — 84 passed
- combined local integration stack — 840 passed

### DIFF
--- a/src/claude_swap/snapshot_source.py
+++ b/src/claude_swap/snapshot_source.py
@@ -15,8 +15,13 @@ from a background thread, never a UI event loop.
 
 from __future__ import annotations
 
-from claude_swap.models import AccountsSnapshot
+import threading
+from dataclasses import replace
+
+from claude_swap.json_output import USAGE_TOKEN_EXPIRED
+from claude_swap.models import AccountSnapshot, AccountsSnapshot
 from claude_swap.switcher import ClaudeAccountSwitcher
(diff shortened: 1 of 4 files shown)


realiti4:
Thanks for the fix and solution! It looks good, merged it.

PR #173, opened 2026-07-23
fix(cli): label token status by credential source

possibilities:
## Summary

`cswap list --token-status` currently formats the stored backup credential for every inactive account. Session-mode accounts intentionally let Claude Code own and rotate a separate profile credential, so the diagnostic can say `expired` while the credential that actually serves usage is fresh.

This change reports read-only OAuth state by credential source:

- active accounts: `active profile`
- inactive accounts with a matching session profile: `session profile`, then `stored backup`
- identity-drifted profiles: explicitly ignored rather than attributed to the slot

## Safety

The added token diagnostics are read-only. They do not refresh, copy, synchronize, persist, or route on credentials, and they do not change normal JSON, usage eligibility, or the ordinary list command's existing usage-fetch behavior. In particular, an expired backup is not overwritten with Claude Code's owner-held session credential.

This follows the credential ownership model established in #96 and makes that already-valid divergence visible instead of presenting the backup as the account's sole token state.

## Validation

- `NO_COLOR=1 uv run pytest tests/test_switcher.py tests/test_cli.py -q` — 442 passed
- live read-only validation showed a fresh session profile and expired stored backup as separate lines while ordinary usage remained healthy

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,10 @@ Not sure which one? `cswap list` is the dashboard — every account's 5-hour and
 
 ```bash
 cswap list
+cswap list --token-status      # show labeled token diagnostics per credential source
 ```
+
+`--token-status` only adds diagnostics: active rows label the active profile, inactive rows distinguish the session profile from the stored backup, and known-drifted session logins are ignored. It never synchronizes credential sources or changes which credential the normal usage path selects.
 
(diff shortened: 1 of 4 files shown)


realiti4:
Thanks for the fix! It looks good, I've merged it.

PR #178, opened 2026-07-23
feat(autoswitch): burn-rate-aware consume-first (skip stranded soonest)

kmizzi:
Follows up the merged consume-first strategy (#93) with the burn-rate-aware refinement discussed there — thanks @realiti4 for the green light and @deathemperor for the foundation. Builds cleanly on current `main` (incl. the two-phase freshness fix `cf4d63b`).

## The problem: soonest ≠ most-salvageable

`consume-first` parks on the account whose weekly window resets soonest. But weekly quota can only be drained **through** the 5-hour window, so an account can reset soonest yet still hold more perishable quota than you could realistically consume before it resets. Parking there to "protect" that quota loses the same quota anyway — the stranded-tail case from the #93 review.

## The refinement: a burn-rate reachability gate

Measure the **active** account's burn rate — only the active account burns quota, so its own two most recent successful samples are the workload throughput — as `Δ7d_pct / Δt`. For a proactive (below-threshold) move onto a sooner-resetting candidate, project:

```
reachable = min(remaining_pct, burn_rate * hours_to_reset)
```

If `reachable < MIN_REACHABLE_PCT` (default 5%), the switch would spend a cooldown to realize almost nothing, so the engine **holds** and lets the next-soonest, *drainable* account be considered instead. Net effect: a soonest-but-stranded account loses to a slightly-later account you can actually finish; and when the imminent window can't be meaningfully drained at all, it stays put (new `reset-too-imminent` hold reason, distinct from `already-consuming-soonest`).

### A gate, not a re-sort — and why

The original sketch was "invert the key: rank by reachable." I implemented it instead as a **gate on the unchanged soonest-first ordering**, because ranking purely by reachable reorders the *common* all-drainable case by remaining quota and loses the urgency-first property that is the whole point of consume-first. The gate only changes behavior when the soonest account is genuinely unreachable — everywhere else the merged strategy is byte-for-byte unchanged.

### Graceful degradation

With no measured rate yet — fewer than two samples, a flat/negative delta (idle time or a just-reset window), or samples straddling an implausible interval — the gate is **inert** and the strategy is exactly today's soonest-first. So early behavior, and the two-account case @Vesely raised, are unchanged; the adaptive part only sharpens target choice once there are multiple alternatives and a real rate.

## Storage

The burn rate needs a prior weekly sample, which the store didn't retain. Per your note that the store schema is best extended here: an **additive** `prev` field (`{sevenDayPct, fetchedAt}`) is rolled forward on each successful `record()` — no other schema change, absent on old rows (→ inert), and only written when both the outgoing and incoming samples carry a 7d utilization. Surfaced on `UsageEntry` as `prev_seven_day_pct` / `prev_fetched_at`.

## Notes

- Samples are sparse for idle candidates, but the rate is read only from the **active** account (polled frequently, and the only one burning), so it's well-sampled in practice; anything outside `[BURN_RATE_MIN_DT_S, BURN_RATE_MAX_DT_S]` is treated as unmeasured.
- Computed inside the existing pure `_rank_candidates`, so the two-phase commit re-derives `reachable` on the fresh phase-2 data for free — the switch-time re-verify covers it.
- All three tuning knobs (`MIN_REACHABLE_PCT`, the two dt bounds) are module constants, easy to tune.

## Tests

New engine tests: stranded-soonest holds (`reset-too-imminent`), later-but-drainable beats soonest-but-stranded, no-rate and idle-active both degrade to soonest-first, drainable-soonest still switches. Plus store tests for the `prev` rollover (two-sample requirement, failure/window-less no-ops). Full suite green under `uv run pytest` (the 6 pre-existing `test_swap_accounts`/`test_move_accounts` fs-permission failures are unrelated and fail on clean `main` too).

Happy to adjust the gate-vs-rank call, the threshold, or fold in a conservative prior if you'd rather it engage before two samples exist.


realiti4:
Hi, sorry for the late reply — I was mostly away from the keyboard, and there was a queue of auth fixes that needed to land first. Thanks a lot for the PR.

The direction is exactly what we discussed in #93, and the gate-not-a-resort call is right. Since main has moved a lot under this branch, I reviewed the current head now so you can fold fixes into the rebase in one pass. No blockers — the design degrades safely to plain soonest-first everywhere — but three things worth fixing; could you check these on your end too?

1. **Cross-plan units in `reachable`.** `burn_rate` is % of the *active* account's weekly quota per hour, but `remaining` is % of the *candidate's* quota — they're only the same unit when both accounts are on the same plan tier. A Max-20x active burning 2%/h against a Pro candidate resetting in 2h with 90% left projects `reachable ≈ 4%` and holds forever, even though that workload would drain a big chunk of the smaller Pro window. The usage payload has no plan info, so documenting the assumption is fine for now (if #169's `subscriptionType` metadata lands, that's the natural way to gate it properly later).

2. **`prev` survives the weekly reset.** The rollover keeps a pre-reset sample; `delta <= 0` catches the big drop, but a *low* pre-reset pct with fast post-reset burn gives a positive delta over an inflated dt → understated rate → wrongful `reset-too-imminent` hold. Invalidating `prev` when it's older than the active account's last weekly reset (the reset timestamp is already available in the engine) closes it.

3. **Two arms of the gate are unpinned by tests** — I mutation-tested this: deleting the `remaining` arm of the `min()`, computing `remaining` from the active instead of the candidate, and dropping the `trigger == "consume-first"` scope check all pass the full suite. Worth adding: a case where `remaining` is the binding arm, the `reachable == 5.0` boundary, an out-of-band dt, a negative delta (reset straddle), and an above-threshold tick confirming the gate stays out of escape paths.

For the rebase itself, two seams to know about: #160 rewrote `record()` into an atomic claims API — the `prev` roll belongs after the sentinel early-return, just before the `row["lastGood"]` write; and #202 (merged today) added an `all_above` escape before the strategy branches in `_rank_candidates`, so your gate lands in the `elif consume_first:` arm and the rank tuple grew by one at both call sites.

One optional thought, take or leave: right after a switch the gate justified, the rate is unmeasured for a couple of samples — if the workload happens to idle there, the next tick can switch onto the very account the gate just rejected. Self-limiting, but carrying the last measured rate briefly would close it.

Once it's rebased I'll re-check and merge quickly. Thanks again.

mrvnklm:
A measurement from a 8-account fleet that bears on `_active_burn_rate`'s resolution, in case it is useful for the rebase.

Every `pct` value the usage endpoint gives me is a whole number. On my own store right now, all 27 window percentages across 8 accounts are integers (`0.0, 1.0, 2.0, 6.0, 8.0, 9.0, 13.0, 14.0, 24.0, 34.0, 39.0, 62.0, 63.0, 65.0, 70.0, 89.0, 90.0, 98.0`), and `build_usage_result` passes them through unrounded (`oauth.py`: `d7_entry = {"pct": d7["utilization"]}`), so the integers are what the API sends, not a rounding on our side.

Since `prev` is the previous *fetch*, `dt` in `_active_burn_rate` is the active account's poll gap — 180s at the movement floor, 600s for a calmer account, well inside `BURN_RATE_MAX_DT_S`. With an integer 7d pct and dt = 180s, `delta` can only be 0 (→ `delta <= 0` → `None`, gate inert) or exactly 1 → `1 / (180/3600)` = 20 %/h. So `burn_rate` is quantised to multiples of ~20 %/h at the floor and ~6 %/h at 600s, with nothing in between. On a weekly window that granularity is coarse enough that `reachable = burn_rate * hours_to_reset` mostly resolves to "0" or "much more than `remaining`", and the gate ends up rejecting roughly the candidates whose reset is inside the first quantum.

This is a resolution problem rather than a design problem — the gate does the right thing when the rate is right. It also compounds the `prev`-across-reset case in the review above, since both are ways of getting a wrong `dt`/`delta` pair. If it helps, the direction that fixes it is a *longer* baseline rather than a shorter one: keeping `prev` from the oldest sample still inside `BURN_RATE_MAX_DT_S` rather than the immediately preceding fetch would put dt in the tens of minutes, where one whole point is a sane quantum. That also happens to make the reset-straddle guard cheaper, since there are fewer, longer intervals to invalidate.

Happy to re-run any of this against a branch if that is useful.

PR #181, opened 2026-07-24
Fix OverflowError in failure backoff on huge consecutive-failure counts

### LINKED ISSUE #179: OverflowError in _failure_backoff_s: backoff shift overflows float after ~1024 consecutive failures (usage stops updating)
## Summary

`cswap auto` (and any usage refresh) dies with `OverflowError: int too large to convert to float` once an account accumulates a large number of consecutive usage-fetch failures. Because the failing account's error aborts the whole evaluation tick, **no** account's usage ever updates — the tool silently stops refreshing.

## Traceback

```
File ".../claude_swap/autoswitch.py", line 706, in _tick_inner
    entries, usage, headroom = self._collect_scheduled_usage(...)
File ".../claude_swap/switcher.py", line 2828, in _collect_usage_entries
    store.record(records, identities)
File ".../claude_swap/usage_store.py", line 588, in apply
    row["backoffUntil"] = now + _failure_backoff_s(failures, rec.retry_after_s)
File ".../claude_swap/usage_store.py", line 349, in _failure_backoff_s
    BACKOFF_BASE_S * (2 ** max(0, consecutive_failures - 1)), BACKOFF_CAP_S
OverflowError: int too large to convert to float
```

## Root cause

`_failure_backoff_s` (`src/claude_swap/usage_store.py:347`):

```python
def _failure_backoff_s(consecutive_failures: int, retry_after_s: float | None) -> float:
    computed = min(
        BACKOFF_BASE_S * (2 ** max(0, consecutive_failures - 1)), BACKOFF_CAP_S
    )
```

The `min(..., BACKOFF_CAP_S)` caps the *result*, but `2 ** (consecutive_failures - 1)` is evaluated **first** as an arbitrary-precision `int`. With `BACKOFF_BASE_S = 30.0` (a `float`), the multiplication forces that int to `float`, which raises `OverflowError` once the int exceeds the IEEE-754 double max (~1.8e308) — i.e. when `consecutive_failures` reaches ~1024:

```python
>>> 30.0 * (2 ** (1024 - 1))
OverflowError: int too large to convert to float
```

An account whose usage endpoint consistently fails (e.g. a personal-plan account that returns HTTP 403/429 on the usage API) increments `consecutiveFailures` on every poll with no upper bound, so the counter eventually crosses the threshold. In my case one account reached `consecutiveFailures: 1024`, and from then on every `tick()` raised the `OverflowError`, was swallowed by the `except Exception` in `tick()` as a transient error (`... (will retry)`), and every account's usage went stale.

## Reproduction

```python
from claude_swap.usage_store import _failure_backoff_s
_failure_backoff_s(1024, None)   # OverflowError: int too large to convert to float
```

Or in the wild: add an account whose usage API returns 403/429, run `cswap auto` for long enough that its `consecutiveFailures` climbs past ~1024, then any `cswap auto --once` fails and `cswap list` stops updating.

## Suggested fix

Cap the exponent before the shift. The backoff result already saturates at `BACKOFF_CAP_S` once `2 ** n >= BACKOFF_CAP_S / BACKOFF_BASE_S` (here `n >= 5`), so bounding the exponent is fully behavior-preserving and simply prevents the giant int:

```python
computed = min(
    BACKOFF_BASE_S * (2 ** min(max(0, consecutive_failures - 1), _MAX_BACKOFF_SHIFT)),
    BACKOFF_CAP_S,
)
```

PR attached. Happy to adjust the approach if you'd prefer clamping `consecutive_failures` at the source instead.

## Environment

- claude-swap 0.23.0 (latest)
- Python 3.14.6 (macOS, arm64) — but the bug is Python-version-independent

adminfriso:
Fixes #179.

### Problem

`_failure_backoff_s` (`src/claude_swap/usage_store.py`) builds the exponential term **before** the cap:

```python
computed = min(
    BACKOFF_BASE_S * (2 ** max(0, consecutive_failures - 1)), BACKOFF_CAP_S
)
```

`2 ** (consecutive_failures - 1)` is an unbounded Python `int`. With `BACKOFF_BASE_S = 30.0` (a `float`), the multiply coerces that int to `float`, which raises `OverflowError: int too large to convert to float` once it exceeds the IEEE-754 double max — i.e. when `consecutive_failures` reaches ~1024.

An account whose usage endpoint keeps failing (e.g. a personal-plan account returning HTTP 403/429 on the usage API) increments `consecutiveFailures` every poll with no ceiling, so it eventually trips this. When it does, the exception aborts the whole `tick()` (swallowed as transient, `... (will retry)`) and **every** account's usage stops refreshing.

### Fix

Bound the shift with a `BACKOFF_MAX_SHIFT = 20` constant. The backoff already saturates at `BACKOFF_CAP_S` by shift 5 (`30 · 2^5 = 960 > 600`), so capping the exponent is fully behaviour-preserving for every reachable value — it only keeps the intermediate int finite.

### Verification

- New regression test `test_backoff_does_not_overflow_on_huge_failure_count` (n = 1024 / 10k / 10M → `BACKOFF_CAP_S`); it raises `OverflowError` on `main`.
- Existing backoff suite unchanged: `pytest tests/test_usage_store.py -k backoff` → 13 passed.

### DIFF
--- a/src/claude_swap/usage_store.py
+++ b/src/claude_swap/usage_store.py
@@ -78,6 +78,11 @@ RATE_LIMIT_TRUST_MAX_AGE_S = 7200.0
 # Failure backoff when the server sent no Retry-After: 30s · 2^(n-1), capped.
 BACKOFF_BASE_S = 30.0
 BACKOFF_CAP_S = 600.0
+# Bound the shift so 2**(n-1) can't grow into an int that overflows float once
+# a persistently-failing account's failure count climbs unbounded. The result
+# already saturates at BACKOFF_CAP_S by shift 5 (30 · 2^5 > 600), so any cap
+# above that is behaviour-preserving; 20 leaves generous headroom.
+BACKOFF_MAX_SHIFT = 20
 
 # The usage endpoint enforces a per-access-token request budget on
 # non-first-party User-Agents (proven 2026-07-11: an idle token, polling
@@ -346,7 +351,8 @@ def _rate_limited_trust_ok(
 
 def _failure_backoff_s(consecutive_failures: int, retry_after_s: float | None) -> float:
     computed = min(
-        BACKOFF_BASE_S * (2 ** max(0, consecutive_failures - 1)), BACKOFF_CAP_S
+        BACKOFF_BASE_S * (2 ** min(max(0, consecutive_failures - 1), BACKOFF_MAX_SHIFT)),
+        BACKOFF_CAP_S,
     )
     if retry_after_s is None:
         return computed
--- a/tests/test_usage_store.py
+++ b/tests/test_usage_store.py
@@ -317,6 +317,12 @@ class TestBackoff:
     def test_backoff_cap(self):
         assert usage_store._failure_backoff_s(50, None) == BACKOFF_CAP_S
 
+    def test_backoff_does_not_overflow_on_huge_failure_count(self):
+        # A persistently-failing account grows consecutive_failures unbounded;
+        # 2**(n-1) must not build an int that overflows float on multiply.
+        for n in (1024, 10_000, 10_000_000):
+            assert usage_store._failure_backoff_s(n, None) == BACKOFF_CAP_S
+
     def test_retry_after_is_the_floor(self, store, clock):
         store.record(
             {"1": FetchRecord(error="http-429", retry_after_s=90.0)}, IDENT


realiti4:
Thanks a lot for addressing this.

This was already merged through #184, which applies the same clamp and adds an end-to-end test through `record()`. It also covers why the crash becomes permanent: after the failed write, the counter never advances past 1024.

I did take your `BACKOFF_MAX_SHIFT` idea though and followed up on main by extracting the constant, with credit to you in the commit 14ed6ef7ceb75475cba8129f328752ed5362e284.

Thanks again, appreciate the report and the PR.

PR #182, opened 2026-07-24
fix(usage): keep exhausted accounts polling

### LINKED ISSUE #180: bug: exhausted accounts age unavailable before quota recovery
## Problem

An account that reaches a relevant 100% window is currently parked until the latest reported reset:

```python
if headroom <= 0:
    next_poll = limiting_reset_ts(...)
```

That can be days away. Meanwhile decision trust for ordinary scheduler staleness is capped at one hour. The result is a contradictory state:

- human `cswap list`/watch still shows the stale last-good 100% measurement;
- `cswap list --json` changes to `usageStatus: "unavailable"` after the trust ceiling;
- routing removes the account as unknown rather than retaining a measured exhausted state;
- no normal collector fetches it again until the distant stored plan is due, so an early provider-side quota grant/reset is not observed.

I reproduced this live with two healthy OAuth credentials: their measurements were 10h and 15h old, `nextPollAt` was parked at their weekly resets, and both JSON rows were unavailable even though token diagnostics were fresh.

## Expected

Exhausted accounts should keep a bounded, provider-safe slow polling cadence. That preserves decision-grade status and detects quota recovery promptly enough to return the account to balancing. A nearer reported reset may pull a poll forward, but a distant reset should not suppress all polling.

Existing persisted reset-parked plans should self-heal after upgrading rather than remaining unavailable until their old deadline.

possibilities:
## Summary

Keep exhausted accounts on a bounded slow usage-poll cadence instead of parking them until a distant reported reset.

- poll accounts with binding usage at 100% every ten minutes (subject to wider 429/AIMD pacing)
- pull that poll forward when a reported reset is nearer
- cap the auto engine's all-exhausted sleep at the same cadence
- detect and repair persisted reset-parked plans written by existing releases

## Why

A reset-parked account can go unfetched for days, while ordinary scheduler trust expires after one hour. It then changes from measured exhausted to `usageStatus: "unavailable"`, drops out of routing as unknown, and cannot observe an early provider-side quota grant.

The bounded cadence remains below the measured usage-endpoint budget (6 requests/hour/token), keeps decision trust continuous between planned polls, and re-admits the account as soon as a fresh response reports headroom.

Fixes #180.

## Safety

- existing per-token fetch claims and stale/plan eligibility still gate fetches
- failure backoff and wider post-429 AIMD intervals win over the exhausted cadence
- explicit fetch semantics are unchanged
- reset metadata can only pull a poll earlier
- legacy-plan repair applies only to structurally impossible deadlines beyond the stored interval plus jitter/reset slack envelope, rechecked under the reservation lock

## Validation

- focused poll policy + usage store + autoswitch + switcher: 575 passed
- upstream CI: 1,583 passed, 3 skipped on Linux; 1,545 passed, 41 skipped on Windows; macOS Keychain contract/real-wrapper job green
- live upgrade repaired two 10h/15h reset-parked rows: both became freshly measured `ok`, received ~10-minute plans, and Keeper now reports them as exhausted rather than unavailable


realiti4 (changes requested):
Hey, this looks really good — thanks for the PR.

Could we adjust a couple of things before merging?

1. **Could we rebase** #160 landed and rewrote the exact reserve block this PR restructures, so the branch now conflicts in `_collect_usage_entries`.

   It should be mechanical: `reserve()` now returns a claims dict (`slot → fencing id`) that feeds into `record(records, identities, claims, plans)`, so the two list-typed `to_fetch.extend(...)` calls would become claims-dict updates.

2. **A guard for non-future resets.** `limiting_reset_ts` doesn’t filter past timestamps, unlike its sibling `earliest_future_reset_ts`, and both new consumers trust it:

   * In `plan_after_fetch`, this:

     ```python
     next_poll = min(next_poll, reset_ts + RESET_SLACK_S)
     ```

     with a past `resets_at` at 100%—from a boundary race or one of the stale server shapes we’ve seen—produces a permanently due plan, causing a refetch on every engine tick until 429s contain it.

     Current `main` happens to be immune because `reset_ts > next_poll` silently ignores past resets. After the rebase, it gets a bit worse: the released claim doesn’t gate, so the scheduled and escalation passes can double-fetch the same row within one tick.

   * `_earliest_recovery` turns the same past timestamp into a past `_sleep_until_ts`, collapsing the blocked sleep to the 60-second tick interval.

   Only applying the pull-forward when `reset_ts > now`, and skipping non-future resets in `_earliest_recovery`, should close both. A small test for each would be great.

Two tiny nits while you’re in there:

* The `EXHAUSTED_INTERVAL_S` comment says “Six-minute polling,” but 600 seconds is ten minutes. Six is the per-hour count, as the PR body has it.
* The two README lines about exhausted accounts—“sleeps until the first one becomes usable” and “exhausted ones left alone until they reset”—could use a matching touch-up.

One observation, explicitly not asking for changes: in all-exhausted mode, the escalation refresh (`respect_plans=False`) can refetch a candidate whose post-429 plan is wider than the 600-second wake. So the “AIMD intervals win” property holds in the planner, but not through escalation.

The worst case is still your budgeted six requests per hour per account, and plan-bypassing escalation predates this PR. Just flagging it in case you think it’s worth a follow-up.

possibilities:
I apologize. I accidently clicked "re-request review". Please ignore.

possibilities:
And I'll likely push again in a minute. Having the agent look over it again along with your comments (first push was automated).

possibilities:
Thanks for the careful review — these were good catches. I’ve updated the PR at `e6dd4a7`:

- Rebased onto current `main` after #160 and kept the two reservation paths as one fenced claims map feeding `record(..., claims, plans)`.
- Guarded both reset consumers: `plan_after_fetch` only pulls forward for `reset_ts > now`, and `_earliest_recovery` treats past/equal reset timestamps as unprovable instead of installing a past sleep target. Both past and boundary-equal cases have regressions.
- Corrected the cadence wording to “ten-minute polling (six requests/hour)” and updated both stale README descriptions.
- I also took the optional AIMD observation: all-exhausted escalation now preserves a decision-trusted exhausted account’s wider post-429 plan rather than bypassing it at the bounded wake cadence. A regression covers the no-reset five-minute fallback case.

While rechecking the legacy-plan path, I tightened two adjacent edges as well:

- reset-parked-plan recognition now happens inside the same store lock that awards the fetch claim, so a concurrent valid replan cannot be bypassed;
- recognition is structural (deadline beyond interval + jitter/slack), not dependent on the current scoped-model selection, so deselecting a formerly exhausted model cannot strand the account on its old distant reset plan.

Validation is green: focused policy/store/autoswitch/switcher suite **575 passed**; Linux CI **1,583 passed, 3 skipped**; Windows CI **1,545 passed, 41 skipped**; macOS Keychain job passed. The PR body’s safety/validation notes are updated too. Ready for another look when convenient.

possibilities:
Happy to send her in for any additional changes, lemme know :)

realiti4:
Thanks for the quick adjustments! Looks good, I've merged it.

PR #183, opened 2026-07-24
fix(theme): consume OSC 11 replies through a DA1 boundary

### LINKED ISSUE #177: OSC 11 reply can leak into shell input after cswap list with ui.theme=auto
## Environment

- claude-swap 0.23.0
- Linux, zsh
- `TERM=xterm-256color`
- `ui.theme=auto` (the default)

## Reproduction

1. Upgrade to 0.23.0.
2. Keep the default automatic theme:
   ```console
   $ cswap config get ui.theme
   auto
   ```
3. Run `cswap list` in a terminal that replies to OSC 11 after the probe timeout.
4. After the command returns, the shell prompt receives text resembling:
   ```text
   11;rgb:1e1d/1e1d/1e1d
   ```

Setting `ui.theme` to `dark` or `light` prevents the problem.

## Expected behavior

The OSC 11 response is fully consumed by cswap and never becomes shell input or visible text.

## Actual behavior

A late terminal response can arrive after cswap restores the tty and exits. The response is then handled by zsh as keyboard input, with the control bytes stripped or interpreted, leaving the visible `11;rgb:...` payload at the prompt.

## Likely cause

`_query_terminal_background()` in `appearance.py` sends `ESC ] 11 ; ? BEL`, waits `_TIMEOUT_S = 0.15`, and then restores the original tty mode. If the response arrives after that deadline, it remains outside the read window and reaches the shell.

There is a `drain_stdin()` helper intended for late OSC replies, but it appears to have no call sites in v0.23.0. Its own docstring also notes that a reply arriving after both the detection deadline and the drain can still leak, so an immediate one-shot drain may not completely close the race.

## Workaround

```console
$ cswap config set ui.theme dark
```

A robust fix may need to ensure outstanding OSC replies cannot outlive the process read window, or avoid probing on short-lived plain CLI commands unless explicitly requested.

Yuanpeng-Li:
## Summary

- send a DA1 query immediately after OSC 11 and keep reading until the ordered DA1 response arrives
- extend the safety timeout from 150 ms to 1 second for SSH latency while letting unsupported terminals return promptly through DA1
- skip probing on `TERM=dumb` and the Linux console, and require complete BEL/ST-terminated OSC 11 frames
- add regression coverage for a response beyond the old 150 ms window, split responses, DA1-only terminals, query echo, and unsupported environments

## Why

With the default `ui.theme=auto`, a terminal that answered OSC 11 after the 150 ms read window could deliver that response after cswap restored the tty. The shell then received the response as input, leaving text such as `11;rgb:...` at the prompt.

DA1 acts as an ordered response boundary: terminals process OSC 11 before the following DA1 query, so consuming the complete DA1 response also consumes any preceding colour response before the tty is restored. This is the same feature-detection approach documented by [terminal-colorsaurus](https://docs.rs/terminal-colorsaurus/latest/terminal_colorsaurus/feature_detection/).

A clarification to #177: `drain_stdin()` is already called on the TUI path before Textual starts; the plain CLI path has no equivalent drain. This change avoids adding an unconditional exit-time `TCIFLUSH`, which could discard real typeahead or stdin consumed by interactive commands.

## Tests

- `uv run pytest` — 1539 passed, 3 skipped (Linux, Python 3.14)
- `uv run --python 3.12 --isolated pytest tests/test_appearance.py tests/test_cli.py -q` — 157 passed

Fixes #177

### DIFF
--- a/src/claude_swap/appearance.py
+++ b/src/claude_swap/appearance.py
@@ -1,14 +1,16 @@
 """Terminal appearance detection and theme resolution.
 
 Determines whether the terminal has a light or dark background by querying it
-with OSC 11 (``ESC ] 11 ; ? BEL``) and reading the ``rgb:…`` reply, then
-classifying by perceived luminance. Cross-cutting: both the CLI printer and the
-TUI resolve their theme through here.
+with OSC 11 (``ESC ] 11 ; ? BEL``) followed by DA1 (``ESC [ c``), reading
+through the ordered DA1 reply, and classifying the preceding ``rgb:…`` reply by
+perceived luminance. Cross-cutting: both the CLI printer and the TUI resolve
+their theme through here.
 
 The query MUST happen while this process owns the terminal in cooked mode —
 before Textual's input driver starts — or the reply is reissued as keystrokes.
 Everything fails safe to ``None`` (→ resolved ``dark``): a terminal that doesn't
-answer, a pipe, Windows, or a parse failure never blocks and never errors.
+answer, a pipe, Windows, or a parse failure never blocks indefinitely and never
+errors.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks a lot for addressing this! It looks good, I've merged it.

PR #190, opened 2026-07-27
fix(add): capture the CLAUDE_CONFIG_DIR profile's own credential

keepitmello:
On macOS, `cswap add` can file a slot with one account's email and a different
account's token. It reports success, and nothing you can check afterwards
disagrees. This makes `add_account` read the credential from the same profile it
read the identity from.

## What I wanted

Two accounts at the same time: one terminal on account B, everything else on
account A, sharing the same settings, skills, and transcripts. That is what
session mode is for. The README opens with:

> Easily switch between multiple Claude accounts without logging out

and describes session mode as:

> every other terminal and the VS Code extension stay on your default account,
> so two accounts can work in parallel

On macOS, adding the second account broke both of those promises.

## What happened

I registered account B and launched it with `cswap run`. Every session on my
default profile logged out, mid-work.

Once I was back in, nothing looked wrong. `cswap list` showed both accounts, and
`claude auth status` in the new session reported account B. Neither one checks
the token, though — both print stored metadata. So I resolved each slot's access
token against `GET /api/oauth/profile` directly. Slot 1 came back as account A.
So did slot 2.

Slot 2 had never captured account B. Both "accounts" were drawing on one
account's 5h and 7d budget. Two accounts, one quota.

The repo already has that oracle: `oauth.fetch_oauth_profile`, the "whose token
is this" call `switcher.py` uses when resolving an identity. It just isn't on
the `add` path.

The logout follows from the same root. `cswap run` refreshes the profile's
credential while bootstrapping it (`session.py`, `_bootstrap`), and an OAuth
refresh rotates the token family. Slot 2's token belonged to the default
profile, so that rotation invalidated the copy every other terminal was still
using.

## Why

`add_account` fills a slot from two reads. The credential read has two backends,
and only one of them follows the env var:

| read | looks at | follows `CLAUDE_CONFIG_DIR` |
| --- | --- | --- |
| identity — `.claude.json` | `paths.get_global_config_path` | yes |
| credential — file backend | `paths.get_claude_config_home` | yes |
| credential — macOS Keychain | `CLAUDE_CODE_KEYCHAIN_SERVICE` | **no** |

That service name is a constant, so it always points at the default profile. On
Linux and Windows the credential comes from the file backend, which follows the
env var, so both reads land in the same profile. On macOS they don't.

Nothing in the suite captures a credential with `CLAUDE_CONFIG_DIR` set, which
is why CI stays green. The macOS job runs only the Keychain wrapper tests
(`test_macos_keychain.py`, `test_macos_keychain_contract.py`), and the new tests
below fake the platform, so they fail on any runner.

## The fix

Read the credential from the same profile the identity came from.

`_read_capture_credentials` does that, and `add_account`'s two capture points
call it. It tries three sources, all inside the profile `CLAUDE_CONFIG_DIR`
names:

1. The profile's hashed Keychain entry, then its `.credentials.json`
   (`session.read_config_dir_credentials`).
2. The active store — but only if the env var names the default profile. Then
   they are the same profile, and someone exporting
   `CLAUDE_CONFIG_DIR=~/.claude` may only have the unsuffixed item.
3. The profile's own `primaryApiKey`, so an API-key login still gets
   `_reject_live_api_key_capture`'s message instead of a generic read error.

Writing is untouched. cswap deliberately never writes claude's hashed Keychain
entry — `session.py`'s module docstring gives the reason, and this fix doesn't
need it. Reading that entry is already done (`read_session_credentials`);
writing it isn't.

Two details the diff turns on:

**Hash the raw string, not a `Path`.** Claude hashes the exported value exactly
as written (`session.py`: "Hash exactly the string we export"). `Path()` strips
a trailing slash and a leading `./`. So `export CLAUDE_CONFIG_DIR=/path/to/p/`
would look up a name claude never wrote, miss, and fall back to the old seed
file — the same wrong-token bug, silently. `keychain_service_name` now takes
`str | Path`, and `read_session_credentials` calls the new function, so existing
callers don't change.

**Resolve paths when comparing them, never when hashing them.** `_same_directory`
resolves both sides so a `$HOME` reached through a symlink still counts as the
default profile. Same pattern as `paths.migrate_legacy_backup_dir` — `resolve()`,
and a literal compare if that raises `OSError`.

## Out of scope

The write path has the same split, and it is worth recording. With
`CLAUDE_CONFIG_DIR` set, `cswap switch` writes `<config_dir>/.credentials.json`
on Linux, but the default profile's Keychain item on macOS — and then
`_refresh_stale_credentials_file` rewrites `<config_dir>/.credentials.json` as
well. So on macOS a `cswap switch` inside a session shell writes account B
twice: into the default profile's Keychain, and into the session profile's seed
file. The session profile's own hashed entry still holds account A, and claude
reads that first. Closing this means writing claude's hashed entry, which the
docstring above rules out, so it needs its own discussion. Happy to open an
issue and follow up.

One smaller gap in this change. The active store retries a busy Keychain read
(`_ACTIVE_READ_ATTEMPTS`) before falling back to the plaintext file; the profile
read tries once, matching `read_session_credentials`'s existing behavior. So a
Keychain that is locked right at capture time can still capture an old seed
token, silently. Both paths end at the same file, so the window is narrower than
before — but it is still there. Fixing it means a strict mode on
`read_config_dir_credentials` for the capture call only, which I left out to
keep this to one change.

## Scope

```
src/claude_swap/paths.py      +11  -1   get_default_claude_config_home
src/claude_swap/session.py    +20  -5   read_config_dir_credentials
src/claude_swap/switcher.py   +56  -2   _read_capture_credentials, _same_directory
tests/test_session.py        +224  -1   TestCaptureCredentials
```

No new configuration, no new dependencies.

Tests sit in `test_session.py` next to `TestReadSessionCredentials`, since this
is `CLAUDE_CONFIG_DIR` profile resolution rather than the account bookkeeping
`test_switcher.py`'s `add_account` tests cover.

## Validation

All runs use `env -u CLAUDE_CONFIG_DIR` — an exported value changes what the
suite exercises.

- New tests against unmodified `main` (`db55bc2`): **6 failed, 12 passed**. The
  six failures are the macOS cases, which is the bug. The twelve that pass are
  the Linux cases and the guards, pinning behavior that already works.
- Full suite on this branch: **1549 passed, 6 failed, 3 skipped**. Those six
  also fail on unmodified `main` (`test_move_accounts` 2, `test_swap_accounts`
  4) and are unrelated.
- `uv run python -m compileall -q src`, `git diff --check`.

Most of these started as a failing test: a trailing slash on the env var, an env
var naming the default profile, a `$HOME` reached through a symlink (POSIX-only,
skipped on Windows), a machine-wide API key answering for an OAuth profile, and
a credential-less profile falling back to a different one. The rest pin guards —
the env var unset or empty, and the in-place refresh branch reading the same
source as the first capture.

### DIFF
--- a/src/claude_swap/paths.py
+++ b/src/claude_swap/paths.py
@@ -53,6 +53,16 @@ def get_global_config_path() -> Path:
     return base / ".claude.json"
 
 
+def get_default_claude_config_home() -> Path:
+    """Return the *default* profile's config home, ignoring ``CLAUDE_CONFIG_DIR``.
+
+    ``_read_capture_credentials`` has to tell an env var that names the default
+    profile from one that names another, since only the former's credential is
+    the active store's.
+    """
+    return Path.home() / ".claude"
+
+
 def get_default_global_config_path() -> Path:
     """Return the global config path of the *default* profile.
 
@@ -61,7 +71,7 @@ def get_default_global_config_path() -> Path:
     (session sharing) must not source from another session when invoked from
     inside one.
     """
-    legacy = Path.home() / ".claude" / ".config.json"
+    legacy = get_default_claude_config_home() / ".config.json"
     if legacy.exists():
         return legacy
     return Path.home() / ".claude.json"
--- a/src/claude_swap/session.py
+++ b/src/claude_swap/session.py
@@ -161,14 +161,16 @@ def session_dir_for(backup_dir: Path, account_num: str, email: str) -> Path:
     return backup_dir / "sessions" / f"{account_num}-{slugify_email(email)}"
 
 
-def keychain_service_name(session_dir: Path) -> str:
+def keychain_service_name(config_dir: Path | str) -> str:
     """Keychain service name Claude Code derives for this config dir.
 
     Claude hashes the raw ``CLAUDE_CONFIG_DIR`` env var value, NFC-normalized
     and unresolved (claude src ``envUtils.ts``/``macOsKeychainHelpers.ts``).
     Hash exactly the string we export — never a resolved/realpath variant.
+    Accepts a ``str`` so a caller holding the raw env value can hash it without
+    a ``Path`` round-trip, which would drop a trailing slash or ``./``.
     """
-    normalized = unicodedata.normalize("NFC", str(session_dir))
+    normalized = unicodedata.normalize("NFC", str(config_dir))
     digest = hashlib.sha256(normalized.encode("utf-8")).hexdigest()[:8]
     return f"Claude Code-credentials-{digest}"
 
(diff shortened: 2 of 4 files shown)


realiti4:
Thanks for the fix and addressing this, looks good merged it.

realiti4:
Quick follow-up on what landed since the merge: 253877c makes the capture fail closed when the profile keychain is unreadable instead of silently reading the stale seed — the gap you flagged — and #205 extends the resolver to `CLAUDE_SECURESTORAGE_CONFIG_DIR`, which 2.1.220 consults ahead of `CLAUDE_CONFIG_DIR` for secure storage. 

On the switch write-path split: your offer to open the issue would be very welcome. I'm planning to take the implementation myself (I think it can stay inside the "never write claude's hashed entry" rule: delete the profile's hashed entry and reseed, the same mechanism seeding already uses), but if you'd rather send a PR for it that's of course appreciated too.

mysubb01:
Thanks for the merge, and for closing the locked-keychain gap in 253877c — fail-closed there is exactly what that path needed.

Opened the write-path issue as #206. Your delete-and-reseed plan sounds right to me (it stays read-only with respect to the hashed entry's format, which was the whole constraint), so I noted it in the issue and left the implementation to you. Happy to review or pick it up later if it falls off your plate.

PR #195, opened 2026-07-28
fix(transfer): close world-readable window on exported credentials

780Farva:
## Summary

`cswap --export <path>` writes a JSON envelope containing each account's live OAuth `accessToken`/`refreshToken` (or raw API key). `_atomic_write_file()` in `transfer.py` built the temp file with `Path.write_text()` and only `chmod`'d it to `0600` afterward — the file sat at the umask-derived default mode (typically `0o644`, world-readable) for the window between creation and that `chmod` call.

Every other credential-bearing writer in this codebase (`credentials.py`, `settings.py`, `migrations.py`) avoids this by using `tempfile.mkstemp()`, which is `0600` from the instant the fd is opened — no window at all. `transfer.py` was the one place that didn't follow that pattern, and unlike the internal backup dirs (`credentials_dir`/`configs_dir`, chmod'd `0700`), the export destination is an arbitrary user-chosen path, so directory permissions don't compensate.

Practical impact: on a shared/multi-user host with a default umask, exporting an account for a cross-machine move (`cswap --export ~/backup.json`) briefly leaves a world-readable file containing a live refresh token in the destination directory — enough for a co-resident local user polling the directory to steal it.

## Fix

Swap the `write_text()` + `chmod()` sequence for the same `tempfile.mkstemp()` + `os.write` + `os.replace` pattern already used elsewhere, so the temp file is `0600` from creation with no exposure window. Also drops the now-unused `shutil` import.

## Test plan

- [x] Added `TestFilePermissions.test_export_temp_file_never_created_world_readable`: spies on `tempfile.mkstemp` and asserts the temp file's mode is `0600` immediately at creation under a permissive `umask(0o022)` — this fails against the old `write_text`-based implementation (the spy is never invoked, since `mkstemp` wasn't used).
- [x] `uv run pytest tests/test_transfer.py -q` → 77 passed
- [x] `uv run pytest -q` (full suite) → 1631 passed, 3 skipped

### DIFF
--- a/src/claude_swap/transfer.py
+++ b/src/claude_swap/transfer.py
@@ -9,8 +9,8 @@ from __future__ import annotations
 
 import json
 import os
-import shutil
 import sys
+import tempfile
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks! Merged it.

Issue #189, opened 2026-07-27
When logging into another account all other accounts always need re-login


Tropaion:
### What happened?

I'm not sure what I'm doing wrong or if it is an bug, but when I login to another account to add it (browser then claude code login) all other accounts get "re-login needed - refresh token dead" so it is impossible to add two or more accounts at the same time.

### cswap version and install method

cswap 0.23.0 uv

### Operating system

Windows

### How do you run Claude Code?

VS Code extension

### Command output or logs

```shell

```

Axiara:
I sometimes notice this issue too. More specifically, if i login and add multiple account back to back, its fine for like a day, then the inactive accounts auth tokens die the next day. Im wondering if there is a mechinisum missing to keep the auths refreshed periodically to keep them from going stale?

Ryrahul:
Facing the same issue

Jimmy89:
I have the exact same issue constantly. Simply logging in another account instantly provided me this error. I run Linux (Fedora 43, Gnome), claude installed in default folders. No customization, jails or anything. Using Teams Pro accounts

Update: with the latest 0.24.0 version I do not seem to experience this behavior :)

firstintent:
I met this one too. This is a recurring issue, so switching between 2 accounts is not possible. This is ubuntu2204.

albusdemens:
@firstintent On Ubuntu 24.04, I found that the solution was to only use `/login`, in Claude Code. So no `\logout` was the way to go.

realiti4:
README is updated to advise against using `/logout` between account logins via Claude Code.

#196 should also help with this case hopefully, also released in [v0.25.0](https://github.com/realiti4/claude-swap/releases/tag/v0.25.0)

Please let me know if the issue continues, so we can reopen it.

Issue #124, opened 2026-07-11
Noob hit a bug


binary64:
Hi, thanks for your add, I installed it, set up 3 accounts, but then did `cswap list` but it shows my first 2 as logged out. What is the trick? Or what caused my problem? I did /logout+/login after 1st one, but for the 3rd one I just did /login.

I am doing the full oauth login, not api key.

Do I need to use --full?

realiti4:
Hi, no, `--full` is only for exports. `/login` then `cswap add`, and repeat with `/login` on claude cli then `cswap add`. No `/logout` needed, maybe you added logged out account to cswap?

binary64:
I tried latest version 0.20.0 but same issue. I login, add, and cswap shows it, but then i login to second and cswap add, but then cswap tui shows the first account as `re-login needed.

All my accounts are the same type - Max20.

realiti4:
Which platform are you on? You do your logins via claude cli right then add? Are you using claude desktop too at the same time? If so can you try again while claude desktop is totally closed.

binary64:
Hi, I'm using Ubuntu 22 desktop. Yes, I'm only using claude cli to `/login`. I never use claude desktop. So roughly here is the sequence of my actions:

1. `claude` - choose dark mode - choose login to plan - it auths fine - i don't say anything
2. `/exit`
3. `cswap add`
4. `claude`
5. `/login` - it auths to 2nd account fine - i don't say anything
6. `/exit`
7. `cswap add`
8. `cswap` <-- shows 1st account as un-authed

realiti4:
Could you share your `claude --version` output and whether Claude is installed via `npm` or the native installer?

Since `cswap purge` deletes the log, first attach your current `~/.local/share/claude-swap/claude-swap.log` file, if it exists. Then run `cswap purge` for a clean slate and repeat the flow with debug logging enabled:

1. `/login` to account 1
2. Run `cswap add`
3. Run `cswap list --debug`
4. `/login` to account 2
5. Run `cswap add`
6. Run `cswap list --debug`

That second `cswap list --debug` is the important moment. If account 1 changes to `re-login needed` there, we’ve caught the issue in the act.

Afterward, please attach the fresh `claude-swap.log`. With `--debug`, it records the server’s exact rejection.

One note: the log does contain your account email addresses on the add and switch lines, so feel free to search and replace them before attaching it. No tokens or other secrets are ever written to the log.

binary64:
```
[...]

PR #196, opened 2026-07-29
Never consume a stale refresh token; never quarantine on one that was

### LINKED ISSUE #187: bug: idle session expiry can falsely quarantine account backup

codeslake:
> Field-deployed (editable) on two machines with behavior monitors and a ~6h soak. Grew out of #167's rotation-before-collection finding and a live false-quarantine incident on a locked-keychain laptop. Companion TUI QoL is split out as #199.

## Problem

A refresh token is one-time-use. Whenever cswap acts on a **stale copy** of a credential — a keychain-fallback file read, a backup that missed a resync, a copy superseded by a sibling's rotation — POSTing that copy's refresh token returns `invalid_grant`, and with `AUTH_DEAD_STRIKES = 1` a single such POST quarantines a **live** account behind "re-login needed" (#187, and part of #189). The user then re-logs-in an account that never died.

The invariant this PR enforces: **`invalid_grant` is evidence about the POSTed bytes, never a verdict on the slot.** Consume only bytes whose freshness is provable; when a consume fails anyway, verify before condemning.

## Changes (landing in order)

1. **Degraded-read provenance** (535274f) — on macOS Claude Code rotates keychain-only, so after a failed Keychain read the plaintext fallback (and the slot backup) can hold a superseded generation that agrees with itself. `ActiveCredentials` now carries `degraded`; a degraded read may be adopted/served but its refresh token is never consumed — the pass surfaces `keychain unavailable` and CC's own next-use refresh heals it. Backup reads gain an unreadable-vs-absent verdict: an unreadable-keychain slot shows "keychain unavailable" (display) and "retry from a GUI terminal" (switch) instead of prescribing a re-add that would burn the stored grant.
2. **Single consume gate** (16a4b18) — `consume_backup_grant` is now the only place a backup refresh token is POSTed: locked re-read (a caller's snapshot may be superseded) → unlocked bounded POST → reacquire and compare-and-swap on the refresh-token fingerprint. A dead `cswap run` profile holding a strictly newer lineage supersedes the backup (#96's shape) and is resynced-then-consumed instead. On a CAS conflict the successor is stashed as an unclaimed credential — a consumed generation is never discarded (#167's invariant) — and the racing writer's newer lineage is adopted. Routed from the collector's inactive refresh (new `refresh_via` seam in `try_fetch_usage_for_account`), autoswitch's freshen, and the session bootstrap (moved before the bootstrap lock: the gate's FileLock is non-reentrant and network never runs under locks).
3. **Strike hygiene** (166412a) — the permanent `invalid_grant` verdict now requires the RFC 6749 §5.2 parse (top-level `error` member); substring hits inside other envelopes stay transient, `invalid_client` becomes a distinct systemic kind (our client credential — never a dead-token strike), and `no_refresh_token` demands a structurally complete OAuth dict (a torn read is transient). Before condemning, the active path re-reads the live store: a moved lineage means our POST consumed a superseded copy — recorded transient, the next pass consumes the successor. Strikes stamp the consumed generation's fingerprint (additive `struckFingerprint` field, #178 convention); `token_dead(stored_fp=…)` unbinds when the stored credential no longer matches, so every credential-writing path (add, import, switch persist, gate CAS) heals a stale strike automatically — generalizing #138's clear-on-import and mirroring autoswitch's fingerprint-released quarantine.
4. **Store-resolution parity** (021b8c9) — when `CLAUDE_SECURESTORAGE_CONFIG_DIR` is set (honored by CC ≥2.1.220, unmirrored by cswap) the consume gate refuses (transient, nothing strikes) rather than consume against a store CC left behind; `switch_to` refuses when `CLAUDE_CONFIG_DIR` points inside a `cswap run` session profile (that shell's live store is the profile). First landable increment — full resolution mirroring can follow upstream if wanted.
5. **Review hardening** (07a8f90, 702d26c) — 14 root-cause gaps in 1–4, each reproduced against the branch before being accepted; all closed:
   - a **per-slot consume lock** held across re-read → POST → CAS (the slot FileLock never covers the network call), so two gates — freshen vs collector, or two concurrent `cswap run`s — can never POST the same one-time-use grant; a loser adopts the winner's fresh successor instead of burning another generation;
   - **exception containment after the grant is consumed**: persist/stash I/O failures (disk full) no longer raise through the never-raises collect pass or discard the successor — last resort is the stash, and the stash is no longer write-only: the next gate pass **adopts** a stranded successor (keyed by the generation it superseded), completing the failed persist without consuming anything;
   - **`consumed_fp` on failure outcomes too**, so a strike binds to the bytes the gate actually POSTed exactly when it substituted a locked re-read or session profile for the caller's snapshot;
   - **active-slot strike parity**: the heal/quarantine verdict considers both stored sources (live + backup) — previously a strike bound to the backup lineage was mis-healed by the live fingerprint every pass, keeping a genuinely dead backup out of quarantine forever;
   - CAS treats an **emptied slot** (remove-account mid-POST) as a conflict (stash, never resurrect); the profile-precedence step honors **`STALE_MARKER`** (a presumed-stale profile can't clobber a deliberate re-add); an **unreadable** (vs absent) backup re-read defers instead of falling back to the snapshot; the M4 refusal also covers the active path; degraded reads no longer feed the rotation resync's write; `store-unmirrored`/`invalid_client` keep their identity on the 401-retry path and render their remedy on every surface (CLI, TUI, autoswitch); import's #136 auto-heal is fingerprint-bound like the collectors; and a fresh strike can no longer inherit a stale `struckFingerprint` from an earlier healed one.

## Why not just raise AUTH_DEAD_STRIKES

#105/#106 set the threshold at 1 deliberately: "one server-confirmed invalid_grant is definitive". That holds when the POSTed bytes are provably the current generation — and this PR's consume gate is what makes them provable. Raising the threshold would slow honest quarantines without fixing the false ones.

## Boundaries kept

- #136: import performs no token validation, no network — unchanged; fingerprint-voiding generalizes #138's clear-on-import without replacing it.
- #62/#95/#167: the active account refreshes only under Claude Code's own lock protocol; this PR narrows what may be consumed, never widens it.
- #165/#166: dead lineages and owner-held expired tokens stay explicit/human recovery territory. #166 covers the session-profile recovery case with an explicit command; this PR is the automatic never-consume-stale layer underneath — the two compose (division of labor: #166 = recovery verbs, this = consumption hygiene).
- #121: the sibling probe follows the advisory-oracle discipline (bounded, never under locks, failure degrades to today's behavior).

## Validation

Per-mechanism TDD; each guard mutation-checked by line number. Full suite on
this head: **2054 passed, 3 skipped**. CI green on every push, including
post-rebase onto the current `main` (0.25.0b1).

**Live soak, two machines (linux + macOS), ~6h**: 8 autoswitch/threshold
switch cycles across 3 OAuth accounts; an active-token expiry end to end
(expiry → "token expired" sentinel → next-pass refresh through the gate →
healthy 8h token, no strike); 2 rotation-outside-collection resyncs; both
machines' engines LIVE against the same 3 accounts. Zero `invalid_grant`,
zero strikes, zero quarantines, empty unclaimed stash throughout.

## Keychain-state correctness

Every rule below is one answer to "is this slot's credential readable", and
each was wrong in a way that produced confident, actionable, false advice.
Each is reproduced against the branch and mutation-checked by line number.

- **`_perform_switch` read the raw Keychain flag** while two sibling sites
  excluded a file mode cswap itself pinned. After one deliberate file-mode
  write, switching to a genuinely empty slot raised "the backup is in the
  Keychain but unreadable — do not re-add", the one remedy that cannot work.
  The three sites now share a `_keychain_unreadable` predicate (41313b9).
- **A failed persist returns credentials AND an error.** The session
  bootstrap's pre-lock consume branched on `not outcome.credentials`, but the
  gate's failed-persist path returns the successor together with
  `error="transient"` while the backup still holds the spent generation — so
  the warning never fired and `_bootstrap` seeded the profile from a consumed
  grant (a8f7e6c).
- **A secure-store miss captured the other profile's key.** With
  `CLAUDE_SECURESTORAGE_CONFIG_DIR` and `CLAUDE_CONFIG_DIR` diverged, a miss
  fell through to a tail reading `primaryApiKey` via
  `get_global_config_path()`, which follows `CLAUDE_CONFIG_DIR` — exactly the
  cross-profile capture the branch above it refuses. Confirmed against the
  Claude Code binary 2.1.220 (b12bdd1).
- **One transient timeout became permanent.** `_keychain_op_failed` recorded
  failures and nothing cleared it, so a later write that pinned file mode
  (zeroing the re-probe cooldown) latched it for the process: `degraded=True`,
  "keychain unavailable" on every usage pass, `cswap add` refused (5928119).
- **A success on one item answered for another.** `_kc_call` clears the flag
  on any success, and backup reads bypass `_use_keychain`, so one idle slot's
  readable backup erased the verdict the ACTIVE read had recorded. Measured,
  same fixture, only slot order changed:

      [2,3]  degraded=True   sentinel='keychain unavailable'  POSTed: []
      [3,2]  degraded=False  error='invalid_grant'            POSTed: ['rt-SPENT']

  `_active_read_failed` now records the active read's own outcome (1383f54).
- **The idle-slot sentinel asked a process-wide flag** rather than the slot's
  own read. With every Keychain read denied and a live backup on slot 2, both
  slots reported "no credentials" — advice to re-add a slot that has one. It
  now asks `_read_account_credentials_ex` (50eeee8).
- **A verified clear outranked later failures.** `_pin_file_mode` zeroes the
  re-probe deadline, but `_kc_call` re-arms it on any failure with no pin
  check and backup reads reach it without `_use_keychain` — so the active read
  IS reachable after a pin. Measured:

      t0  write falls back, delete SUCCEEDS   verdict=True
      t1  a backup read fails                 deadline re-armed
      t2  cooldown lapses, ACTIVE read fails  degraded=True   (correct)
      t3  the next read                       degraded=False  (wrong)

  A verified clear now settles the flags at the pin instead of overriding
  them (2b99678).

## A corrupt session file cannot escape the scan

`_post_backup_write` runs after the credential store write has completed, and
its first call is `list_sessions`. Anything that escapes there escapes past a
write that already happened: at the post-POST site the grant is spent while
the successor is neither persisted nor stashed and the result is reported as
`transient`; on `cswap add` the credential is stored and the roster is not.

The scan caught `json.JSONDecodeError`, `KeyError`, `TypeError` and `OSError`.
Three faults a file's own content can induce fell outside that set, all of
them in `list_sessions` and its sibling `list_ide_instances`:

```
undecodable bytes          UnicodeDecodeError   (a ValueError, not an OSError)
pid >= 2**31               OverflowError        (os.kill, via is_pid_alive)
deeply nested JSON         RecursionError
a top-level JSON array     AttributeError       (data.get on a list)
```

The last two are the same input on different machines: the nesting depth that
raises `RecursionError` on one parses on another and then reaches `.get`, so
both outcomes have to be inert. `clear_session_stale` had the matching gap on
the other side — its legacy child unlink was guarded while the sibling unlink,
which is where every marker is written today, was not, so a denied
`sessions/` directory made `remove_account` destroy the credentials and then
raise before the roster write.

Widening that tuple keeps a bad file from taking out a listing, which is what
a SCAN needs. It is the wrong answer for the GUARDS reading the same
directory: `list_sessions` hands back a shorter list, and "no live sessions"
and "no readable records" are the same empty list. The callers gating
destruction — `_bootstrap` (deletes the profile's Keychain entry, overwrites
`.credentials.json`), the stale-marker invalidation, the history merge,
`_ensure_no_live_session`, and `--purge` — read the second as the first, so
one unreadable record opened every one of them underneath a running claude.

`scan_sessions` now returns the count of records it could not read alongside
the sessions, and `list_sessions` is the scan-facing view that drops it.
`profile_is_quiescent` is false whenever anything was unreadable; the two
sites that report PIDs refuse with their own message carrying the count,
since they cannot name a PID they never read. `live_sessions_for` was renamed
to `scan_live_sessions` so a call site left on the old name is a `NameError`
rather than a tuple read as unconditionally truthy.

Mutation-checked in three directions — dropping the count in `scan_sessions`,
dropping `and unreadable == 0` from the predicate, dropping the refusal in
`_ensure_no_live_session` — each kills at least one test. A test that only
asserts the exception tuple survives all three.

The same invariant holds two more places in this file, both with a destructive
handler already wired to the negative answer.

`_is_session_valid` caught `(OSError, TimeoutExpired)` around `claude auth
status` and returned False; the post-bootstrap caller reads False as "the
profile is invalid" and runs `_cleanup_failed_session`, which deletes the
profile's Keychain entry AND `rmtree`s the directory before telling the user
to re-add the account. So `claude` missing from PATH, or the probe exceeding
its 10s timeout on a loaded machine, destroyed a profile that had just been
built. The comment above that probe already records this happening on Windows
through `FileNotFoundError` — the PATHEXT cause was fixed and the collapse
left in place. It is now `_session_validity` → `"valid"` / `"invalid"` /
`"unknown"`, where `"unknown"` covers every way the PROBE fails; the
destructive caller refuses on it and leaves the profile alone, naming what
could not be verified. `_is_session_valid` remains as the reuse-shaped
boolean, where `"unknown"` answering False was already the safe result.

`_delete_session_profile` owned two artifacts and asked about one. The stale
marker is deliberately a SIBLING of the profile directory, and `purge` removes
profile DIRS (`iterdir()` filtered by `is_dir()`), leaving the dot-file beside
them by design — so the marker outlives the directory exactly when the
early-out on a missing directory skips clearing it, and the next profile for
that slot+email re-bootstraps on a flag nothing set for it. The unlink no
longer sits under the existence check.

Mutation-checked, three more directions, each killing the test that names it:
collapsing `"unknown"` back to `"invalid"`, dropping the caller's `"unknown"`
branch, and restoring the early-out. The probe tests carry a control asserting
that a probe which RUNS and reports not-logged-in still cleans up — without it
they pass against a version that never cleans up at all.

Fixes #187. **Addresses** #189 — narrowed from "Fixes" on review: current Claude Code's `/logout` POSTs the refresh token to `${TOKEN_URL}/revoke` before wiping locally, so the logout→login account-add flow genuinely revokes the prior account's grant server-side. No client code can prevent that slice; the remedy is documentation (add accounts via `/login` without `/logout`), which realiti4 is taking. The **#124 claim is dropped** pending that docs change. Addresses the remaining enforcement gap from #164 (closed at #167's merge). Builds on #167's lock protocol and persist-successor invariants.

Refs: #96 #105 #117 #142 #160 #161 · field data: #152, #167 review thread


codeslake:
Rebased onto `bae0889` (merged `upstream/main` in, as before). Two things worth flagging, because #225 and this PR fix the same defect from opposite ends.

## #225 and this PR collide in `_is_session_valid`

`1d72a0b` makes `_is_session_valid` return `True` on `TimeoutExpired`; this PR had replaced that function with a tri-state `_session_validity` whose `"unknown"` answered `False`. The merge conflicts on that except clause, and `test_probe_timeout_leans_valid` lands beside our tests and disagrees with our code. Measured on the conflict resolved to our side, before changing anything:

```
test_probe_timeout_leans_valid    FAILED   assert False
test_probe_oserror_stays_invalid  PASSED
```

**That failure was real and #225 was right about it.** `_is_session_valid` is the *reuse* check. With `"unknown"` answering `False`, a machine loaded enough that claude's cold start exceeds the 10s probe stopped reusing a good profile: it fell through to bootstrap and then raised `SessionError`. This PR closed #224's deletion and opened a hard failure on the same input.

**And #225 is not a superset of this PR either.** It leaves `OSError` failing closed, so `claude` missing from PATH still reaches `_cleanup_failed_session` and still deletes the profile — the Windows PATHEXT incident the comment above that probe records.

Resolved by splitting the probe-failure verdict, because reuse wants opposite answers for the two and deletion wants the same answer for both:

| verdict | meaning | reuse | delete |
| :-- | :-- | :-- | :-- |
| `"unknown"` | the probe ran and did not answer (timeout, unparseable output) | leans **valid** — #225's contract | never |
| `"unreachable"` | `claude` could not be spawned at all | leans **invalid**, so you get the bootstrap path and its "check that `claude` is on PATH" error | never |
| `"invalid"` | the probe answered and the profile is bad | no | yes — the only verdict that does |

Both of #225's tests pass unmodified. Suite 2056 passed, 3 skipped. Mutation-checked, each direction killing exactly one named test:

```
OSError back to "unknown"            -> test_probe_oserror_stays_invalid
boolean back to `== "valid"`         -> test_probe_timeout_leans_valid
caller stops refusing "unreachable"  -> test_an_unverifiable_probe_does_not_destroy_the_profile
```

## CI cannot see this PR's other half, because it pins 3.12

`requires-python = ">=3.12"`, and all three CI jobs pin `python-version: '3.12'`. That gap hides a live one:

```python
# main today
enc_present = enc_file.exists()   # 3.12 RAISES on an unsearchable dir; 3.13+ returns False
```

On 3.13+ the `except OSError` arm below it never fires, so an unsearchable `credentials/` reads as `("", False)` — byte-identical to a genuinely absent backup, which is the exact confusion this PR exists to remove. Measured on Python 3.14.3, same tree, `test_unsearchable_credentials_dir_is_not_absent`:

```
CI (3.12)      2157 passed, 3 skipped
local (3.14.3) FAILED  ('', False) is byte-identical to a genuinely ABSENT backup
```

This PR already fixes it — `stat()` raises on every version, so the guard works on all of them — but the class is worth a CI matrix (`3.12` + latest) regardless of this PR's fate. A guard that silently stops guarding on the interpreter most contributors run is the failure mode a version pin creates.

codeslake:
Merged `ba6f3f9` in. Short follow-up to the comment above, because that one was written against `bae0889` and you superseded it the same day — so "their two tests pass unmodified" was a statement about a tip that no longer exists. Re-run against `ba6f3f9`, it holds, but only after a real fix.

**`test_stale_marker_plus_probe_timeout_still_rebootstraps` failed on the merged tree, and it was right.** It died on this PR's own error:

```
SessionError: Session profile for Account-2 could not be verified:
`claude auth status` did not run or did not answer.
```

Refusing to *delete* on an unverifiable probe is this PR's point and is unchanged. Refusing to *proceed* was not thought through: after a good bootstrap from the backup credentials, a probe timeout must not fail the launch. Your test is the only thing that said so.

Fixed by lifting your fallback into `_artifacts_say_usable(session_dir, email, org_uuid)` and calling it from both places that ask the same question — the reuse check and the post-bootstrap check — rather than keeping a copy in each:

| site | `"unknown"` (probe ran, no answer) | `"unreachable"` (`claude` would not spawn) |
| :-- | :-- | :-- |
| `_is_session_valid` (reuse) | your artifact fallback | `False` — no file on disk says whether `claude` runs |
| post-bootstrap | proceed | stop, profile left in place |
| `_cleanup_failed_session` | never | never |

`_session_validity` itself is unchanged: the verdict layer answers what the *probe* established, and a timeout established nothing. Putting the artifact fallback there would make "the probe timed out" and "the profile is bad" one value again, which is the conflation the tri-state exists to undo.

2063 passed, 3 skipped, with your 145 new test lines unmodified. Mutation-checked, each killing a named test:

```
drop the post-bootstrap promotion -> test_stale_marker_plus_probe_timeout_still_rebootstraps
promote "unreachable" as well     -> test_an_unverifiable_probe_does_not_destroy_the_profile
reuse arm back to blanket True    -> test_stale_marker_plus_probe_timeout_still_rebootstraps (+3)
```

One thing from the earlier comment still stands and is untouched by `ba6f3f9`: the `except OSError` arm still returns `False`, so `claude` missing from PATH still reaches `_cleanup_failed_session` and still deletes the profile — the Windows PATHEXT case the comment above that probe records. That half is what this PR is still for.

realiti4:
Hi, thanks a lot for the PR! This will be a really good improvement. Implementation and tests look solid. Sorry for the late reply.

Could you check these as well please before merge, thanks:

---

Full review done on 5881f4d: trial merge (trivial — you're already on main), full suite locally (2063 passed, 3 skipped), line-level pass of all production changes, plus three deep audits — lock ordering, interaction with f073962's oracle attribution, and overlap with #205. The design is right and no structural change is being asked for. But there are three paths where the PR's own invariant — never act on a spent/stale grant — doesn't hold yet, so: **request changes** on those, all three surgical.

## 1. Session bootstrap still seeds a spent grant (session.py:654-671)

When the gate consumes the grant but the persist fails (including `consume-gate-persist-lock-failed` — a 10s slot-lock timeout under multi-surface contention, so not an exotic trigger), it returns the successor with `error="transient"`. a8f7e6c made the warning fire — but the code then continues into `_bootstrap`, which re-reads the backup, and in exactly this state the backup still holds the generation whose grant was just spent. The profile is seeded with a dead refresh token; claude's first refresh gets `invalid_grant`.

`test_a_failed_persist_warns_rather_than_seeding_a_spent_grant` asserts the warning and that the backup is unchanged, but never asserts what the profile was seeded with — so it pins the half that was fixed and not the half its name promises.

The demoted shape is distinguishable at the call site (`error == "transient"` **with** credentials present; genuine network-transient returns `credentials=None`). Suggested fix: raise a retry-shaped `SessionError` there instead of proceeding — the next run's gate pass adopts the stashed successor without consuming anything and bootstraps normally, so the recovery machinery you already built does the rest. Seeding the profile directly from `outcome.credentials` is a possible alternative (and in the `consume-gate-unpersisted` corner it's the only thing that preserves the successor at all), but it's only safe if the profile→backup reconciliation is guaranteed afterwards — your call whether that's worth the extra path.

## 2. An unreadable stash manifest reads as "nothing stashed" (credentials.py:1449-1459)

`_read_stash_manifest` collapses every failure to `{}`, and `_adopt_stashed_successor` iterates manifest rows only. So with a stash pending (backup = spent generation) and the manifest momentarily unreadable, the gate is blind to the successor and POSTs the spent generation — the exact POST the PR exists to prevent. Note the failure is correlated, not independent: the stash exists *because* storage I/O already failed once, so manifest trouble is most likely in precisely the runs that need it.

Ask: the absent-vs-unreadable `_ex` verdict pattern you applied everywhere else, applied to the manifest read, deferring as `stash-unreadable` (the kind, message, and rendering already exist). Scoped firmly to *unreadable* → defer. Corrupt is a judgment call: the entry **bytes** survive as orphan files (`cswap unclaimed` lists them), but the slot/consumedFp mapping is gone, so nothing can adopt them safely — `{}` gets the slot moving again at the cost of one self-healing strike, while failing closed (at least when orphan entry files exist) avoids POSTing a spent grant at the cost of freezing refreshes until a human intervenes. Your call which side of that you want.

## 3. `refresh_input = current or snapshot` (switcher.py:2137)

A slot removed between the caller's read and the locked re-read comes back absent (not unreadable), and the gate falls back to POSTing the caller's snapshot — spending a grant and then stashing the successor of an account the user just deleted, which contradicts the never-resurrect rule your own CAS branch enforces twenty lines later. All three production callers source their snapshot from the backup store, so an absent re-read really does mean removed; deferring there breaks nothing.

## Fixes-claims and one discovery you should know about

Verified against the installed 2.1.227 binary: current Claude Code's `/logout` is no longer local. `performLogout` now POSTs the refresh token to `${TOKEN_URL}/revoke` (best-effort, then wipes locally). So the logout→login account-add flow genuinely revokes the prior account's stored grant server-side — that slice of #189/#226 is real death no client code can prevent, and the fix for it is README territory (add accounts via `/login` without `/logout`; #226's reporter confirmed that procedure works). I'd narrow `Fixes #189` to `Addresses #189` and drop "likely fixes #124" until the docs change lands — I'll take the README update.

## What checked out

For completeness, the audits came back clean: lock ordering is globally consistent (autoswitch-state → consume → slot → CC's locks, matching CC's own internal order — no reversible pair anywhere), the consume lock is a kernel flock so a killed process can't leak it, and f073962's attribution layer survives byte-identical with the gate layering under it rather than around it. The b12bdd1 `primaryApiKey` fix also closes a residual leak in #205's capture path — good catch. The #225/ba6f3f9 reconciliation with the tri-state verdict is better than either side was alone, and I agree on the CI matrix (3.12 + latest); I'll set that up.

Two wording nits, batch them with the fixes: the gate docstring's "the single place a backup refresh token is POSTed" overclaims — `_fetch_active_usage` is a second site, correctly serialized under the same consume lock, and its own comment at that site says so; and the "not mirrored by cswap" warning is overbroad since #205 — the capture path does mirror now, it's consume/switch that don't. The switch path having no `CLAUDE_SECURESTORAGE_CONFIG_DIR` guard is a real gap but out of scope here; noting it for a follow-up.


codeslake:
Thanks — this is the first review any of these have had, and all three findings were real. Each one reproduced here before I touched anything, each fix is RED-first, and each guard is mutation-checked. `5a154cc`.

Suite **2068 passed, 3 skipped**, on both **3.12** (what CI pins) and **3.14**. CI green on all three jobs.

## 1. Session bootstrap seeded a spent grant — fixed by refusing

Reproduced first. With the gate stubbed to the demoted-persist shape, `setup_session` returned normally and the log says it all:

```
Could not refresh the token for Account-2; continuing with the stored credentials.
INFO: Bootstrapped session profile for account 2 at .../sessions/2-account2_example.com
```

Your reading of the test was exact: it asserted the warning and the untouched backup, and never what landed in the profile. It now asserts that, and is renamed to what it checks (`test_a_failed_persist_does_not_seed_the_profile_from_a_spent_grant`).

I took the retry-shaped `SessionError`, not the seed-from-`outcome.credentials` alternative — the recovery machinery already exists and the next run's gate adopts the stash without consuming anything, so the extra reconciliation path buys nothing here.

**The cost, stated plainly:** in the `consume-gate-unpersisted` corner you flagged, the successor is lost, because it survived only in that return value. I think that is the right trade: reaching that corner means both the persist *and* the stash write failed, and the existing log there already tells the operator the successor "survives only for this pass. Fix the storage failure". Seeding a profile from a disk that cannot write is not a recovery. Say the word if you want the other side of it.

The shape is distinguishable exactly as you said — `error != None` **with** credentials — and a genuine network-transient (`credentials=None`) still warns and proceeds, which `test_refresh_failure_uses_stored_creds` pins unchanged.

## 2. Unreadable stash manifest — fixed, and it hid a second bug

`_read_stash_manifest_ex` now carries the absent-vs-unreadable verdict, and `_adopt_stashed_successor` defers through the `CredentialReadError` path you already built, surfacing as `stash-unreadable`. Reproduced first: with a stash pending and the manifest read denied, the gate POSTed `rt-old` — the spent generation.

**Corrupt stays `{}`.** Not for symmetry, for a deadlock: the repair is `_write_stash_manifest` renaming the bad file aside, and that runs **only on a manifest WRITE** — which deferring is precisely what prevents. Fail closed there and the slot never refreshes again without a hand `cswap unclaimed --purge`. Unreadable self-clears; corrupt does not. `test_a_corrupt_manifest_still_posts_rather_than_deadlocking` pins the line so nobody "fixes" it later.

**The second bug, found reviewing my own diff:** splitting on `OSError` alone lets undecodable bytes escape the reader entirely — `read_text` decodes inside the read and `UnicodeDecodeError` is a `ValueError`. The single `except Exception` I replaced had been swallowing them. The reader takes **bytes** and decodes on the parse side, so each failure class lands in exactly one verdict.

That also removed an interpreter dependency, measured here on the same unsearchable dir:

```
python 3.12: exists() RAISED PermissionError   read_bytes RAISED PermissionError
python 3.14: exists() -> False                 read_bytes RAISED PermissionError
```

Same trap as the `enc_file.exists()` one this PR exists to close — a manifest verdict that changed with the interpreter.

## 3. `refresh_input = current or snapshot` — fixed, with one thing for you to rule on

Deferred on an absent re-read. Reproduced first: the gate spent a grant for a slot with no stored credential and logged `consume-gate-slot-removed` while stashing the successor. That stash is permanently unadoptable, incidentally — `credential_fingerprint("")` is `None`, so `_adopt_stashed_successor` returns before the scan. It was one orphan file per pass, forever.

**What I need your call on.** The `or snapshot` fallback fed one more thing I did not expect, and my guard now returns before it: the session-profile precedence block below. With the backup absent, that block could resync a newer session profile into the empty slot and POST it.

For the case you named, that is moot — `_delete_account_files` deletes the session profile too, so after a real `cswap remove` there is no profile to promote. What my guard does remove is the case where the backup vanished **without** the profile vanishing (a wiped keychain, a half-finished delete): the gate used to silently heal the slot from the profile, and now it defers.

No test covered that path either way, which is why I am raising it rather than claiming it. I left it deferring, because it matches the verdict your CAS branch already reaches for an empty store and the documented remedy for an empty slot is a re-add. If you would rather keep the heal, the signal is available — a present profile is evidence the account was not removed — and I will add it.

## Wording nits, both done

- The gate docstring no longer claims to be "the single place". It names `_fetch_active_usage`'s recovery branch as the second call site under the same consume lock, and says what actually is single: the serialization. Fixed in the matching `autoswitch.py` comment and the `TestConsumeGate` docstring, which repeated it.
- The `store-unmirrored` warnings said "not mirrored by cswap"; since #205 the capture path *does* mirror it. Both warnings and both comments now say capture mirrors, consume and switch do not.

## Fixes-claims

Description updated: `Fixes #187` stands, #189 narrowed to **Addresses**, and the #124 claim **dropped** pending your README change. Your `/logout` finding is the useful half of that — worth recording that `performLogout` POSTs to `${TOKEN_URL}/revoke` before wiping locally, so no client code can save that flow.

Agreed on the CI matrix, and thanks for taking both that and the docs.

## Mutation checks

Each guard removed in turn; each kills exactly the test that names it, and the restored tree is green.

```
refresh_input = current or snapshot -> test_a_removed_slot_defers_instead_of_posting_the_snapshot
ignore manifest_unreadable          -> test_an_unreadable_stash_manifest_defers_instead_of_posting
warn-and-proceed on demoted persist -> test_a_failed_persist_does_not_seed_the_profile_from_a_spent_grant
corrupt treated as unreadable       -> test_a_corrupt_manifest_still_posts_rather_than_deadlocking
decode inside the read              -> test_a_manifest_of_invalid_utf8_is_corrupt_not_unreadable
```

Two controls guard the new tests from passing on a gate that simply never POSTs: `test_a_readable_empty_manifest_still_posts` and `test_a_corrupt_manifest_still_posts_rather_than_deadlocking`.

realiti4:
Thanks for the fast adjustments. About the question, yes I think deferring is the right call.

The review pass found two residuals in `5a154cc` (plus a test-premise nit). Could you check these as well please? Thanks again.

---

Verified independently on `5a154cc`: suite 2068 passed / 3 skipped locally, CI green, and findings 1 and 3 are properly closed — the renamed test asserts the profile side this time, and the removed-slot defer is the right policy. The corrupt-vs-unreadable split with the bytes-based reader is a good catch on your own diff. Wording fixes all read correctly. Three residuals:

## 1. Your question: keep the defer

Keep it as implemented. The half-state (backup gone, profile surviving) can be a removal observed midway, and healing from the profile is the resurrect class this PR forbids everywhere else — it's the same verdict your CAS branch reaches for a slot emptied mid-POST, and a half-finished delete shouldn't be undone by a background poll. The wiped-keychain case has the documented remedy (re-add). One ask since this is now defined behavior on a previously untested path: pin the profile-present variant with a test, so the defer doesn't silently regress into a heal later.

## 2. `consume-gate-unpersisted` is misreported by the new SessionError

The new session branch says "Nothing was lost — the successor is stashed. Please retry" — but the demoted shape it catches doesn't distinguish a stashed successor from the `consume-gate-unpersisted` corner, where both the persist *and* the stash failed. There the message is wrong on both counts: the successor was lost (it survived only in the return value the raise discards), and retrying POSTs the spent predecessor. You stated that cost plainly in your comment — accepting it is fine; the message claiming a stash that didn't happen is the part that shouldn't survive.

Smallest fix: carry the stash disposition on the outcome (a field or just `stashed: bool`) and vary the message — stashed → "retry, the next run adopts it"; unpersisted → "fix the storage failure; if the slot strikes, re-login and `cswap add`", which is what the gate's own ERROR log already says at that site. An honest two-sided message without the field would also do.

## 3. The manifest fix doesn't cover the mutator, and the corrupt rationale has a flawed premise

**`_mutate_stash_manifest` still reads through the plain wrapper**, which discards the unreadable verdict. So a stash writer hitting an unreadable-but-valid manifest gets `{}`, and `_write_stash_manifest` then renames the good manifest aside and writes a fresh one containing only its own row — automatically orphaning every previously mapped successor. Same invariant, same correlated-failure setting as the adopt-scan fix, one read site missed. Fix is small and matches the mutator's own contract ("a failed STASH must be loud"): read through `_ex`, raise on `unreadable=True`.

**The corrupt-stays-`{}` control test proves less than it says.** `test_a_corrupt_manifest_still_posts_rather_than_deadlocking` stashes a successor of `_OLD` — so `_OLD`'s grant is spent by premise — and then mocks the POST of `_OLD` to *succeed*. In reality that POST returns `invalid_grant` and the gate returns before any manifest write, so the corrupt manifest is not renamed and there is no "self-heal at the cost of one POST": the actual cost is a guaranteed dead POST, a strike, quarantine until re-add, and reuse-detection risk against the very successor sitting orphaned on disk. The deadlock argument still has force — fail-closed does need an exit — but `cswap unclaimed --purge` plus the `stash-unreadable` guidance is that exit. Suggested policy split: corrupt manifest with **no** orphan entry files → set aside and proceed (nothing to protect); corrupt **with** orphan entries → fail closed with operator guidance, because a pending successor may exist. And either way, the control test's mock should return `invalid_grant` so it exercises the world its premise describes.

This is an approve once these land.

codeslake:
All three landed in `027bc4c`. The test-premise nit was the one that mattered — it took a rationale I had argued for down with it.

Suite **2074 passed, 3 skipped**, on **3.12** and **3.14**. CI green on all three jobs.

## 3. The corrupt rationale was built on a world that cannot happen

You are right, and the defect was in my reasoning before it was in the code. The old control test stashed a successor of `_OLD` — so `_OLD`'s grant is spent *by premise* — and then mocked the POST of `_OLD` to **succeed**. Nothing in that scenario can occur. With the mock returning `invalid_grant`, the gate returns at

```python
if result.error is not None or not result.credentials:
    return dataclasses.replace(result, consumed_fp=consumed_fp)
```

before any manifest write, so nothing is ever set aside. "Self-heal at the cost of one POST" was an artifact of the mock. The real cost is a guaranteed dead POST, a strike, and quarantine at `AUTH_DEAD_STRIKES = 1`, with the successor orphaned on disk.

The deadlock argument survives, but the exit is the operator's, not a POST. Policy is now split on whether anything is actually at risk:

| manifest | entry files on disk | verdict |
| :-- | :-- | :-- |
| corrupt | none | proceed — `{}` is not a guess, there is provably nothing to lose, and this is what lets the set-aside happen at all |
| corrupt | some | fail closed — a pending successor may exist |
| unreadable | either | fail closed — transient, self-clearing, costs a pass |

That needed three states, so `_read_stash_manifest_ex` returns `ok`/`unreadable`/`corrupt` rather than a bool. A bool cannot carry it: the two failure modes need different answers, and one of them is caller-dependent.

**I did not want to assert the exit in a sentence**, since a message naming a remedy that does not work is worse than no message. `test_the_purge_exit_the_fail_closed_message_names_actually_works` walks it: under a corrupt manifest it confirms `cswap unclaimed` still lists the orphan (the glob in `_list_unclaimed_credentials` does not need rows), purges it, and confirms the slot then moves. That is also why the mutator's refusal is scoped to `unreadable` and not to corrupt — scoping it wider would have blocked the exit I was pointing at.

## 3b. The mutator

Fixed as asked: reads through `_ex`, raises on `unreadable`. Reproduced first — the healthy manifest really was renamed aside and replaced by a single-row file. The test asserts the bytes are unchanged *and* that no `.corrupt-*` sibling was created, because "did not lose rows" and "did not set a good file aside" are separate failures.

## 2. `consume-gate-unpersisted`

`RefreshOutcome` now carries `stashed`, set at the demotion return from `stashed_reason != "consume-gate-unpersisted"` — that reason is assigned precisely *when* the stash write raised, so it is the one demoting reason that parked nothing.

The two corners now get opposite advice, and the second no longer claims a stash:

- stashed → "The successor is stashed — please retry, and the next run adopts it automatically."
- not stashed → "…could neither be stored nor stashed, so the backup holds a spent grant and the successor is gone. Fix the storage failure first; retrying before that spends nothing but earns a strike."

I used the `cswap --add-account --slot N` phrasing the sibling errors in that file already use, rather than the `cswap add` form from the gate's log.

## 1. Defer kept, and pinned

Kept as implemented. `test_an_absent_slot_defers_even_when_a_newer_profile_exists` pins the profile-present variant — absent slot, surviving profile on a newer generation, no live session — and asserts both that nothing was POSTed and that the slot's stored credential was not resurrected. Removing the guard kills that test *and* the removed-slot one.

## Mutation checks

Each new guard removed in turn; each kills exactly the test that names it, restored tree green.

```
mutator ignores unreadable      -> test_a_stash_write_refuses_an_unreadable_manifest
drop corrupt+orphans condition  -> test_a_corrupt_manifest_with_orphan_entries_fails_closed
session ignores outcome.stashed -> test_an_unpersisted_successor_is_not_reported_as_stashed
unlistable dir reads as empty   -> test_an_unlistable_dir_counts_as_entries_at_risk
remove the absent-slot guard    -> test_an_absent_slot_defers_even_when_a_newer_profile_exists
                                   + test_a_removed_slot_defers_instead_of_posting_the_snapshot
```

The fourth is one I added unprompted: `_stash_entry_files_exist` is what lets a corrupt manifest proceed, so a glob that *fails* must not read as "nothing on disk" — same empty-means-safe conflation as the rest of this PR, one level down. It answers True when it cannot list.

Thanks for both passes. The premise catch in particular is the kind that survives a green suite indefinitely.

realiti4:
Thanks a lot for the PR and adjustments again! It looks good, I've merged it

PR #197, opened 2026-07-29
fix(usage): clear the Retry-After deadline instead of retrying on it

codeslake:
Follow-up observation on #152, from that PR's own post-merge data.

## What #152 fixed, and what it left

#152 raised `RETRY_AFTER_FLOOR_CAP_S` from 900 to 3600 so we stop re-probing *inside* a block that was going to last the full window anyway. That was right, and this PR keeps it.

What it left is the block **boundary**. We honor `Retry-After` exactly, so the retry lands on the deadline itself, and the server is not reliably ready there.

## Evidence

Parsed one machine's entire `claude-swap.log`: 511 `http-429` lines, grouped into 39 blocks by shared wall-clock deadline (`ts + retry_after`, 5s tolerance).

**The penalty is a fixed hour, not capacity aging out.** The opening `Retry-After` of a block is exactly 3600 in **37 of 39** blocks (outliers 3356 and 1794). Under a rolling window the opener should be scattered below 3600.

**Retrying at the deadline costs another hour.** Re-derived from the full log during review (41 blocks, 40 opening at exactly 3600): of the lapses followed by another 429, **20 of 35 re-blocked within 900s of their own deadline**, band +2s..+887s, next lapse +1004s:

```
+2.0s  +2.6s  +3.2s  +71.2s  +79.3s  +101.2s  +336.9s  +345.7s  +713.9s  +715.5s
```

Each earned a fresh full-hour penalty. The next lapse after those is **+3852.7s**, so the distribution is bimodal with an empty band between ~716s and ~3853s.

2026-07-28 was an 11th instance: `backoffUntil` 22:30:16, next 429 at 22:30:19, i.e. **deadline + 3s**. At that moment cswap had made *one* request in the trailing hour on that row (`_row_eligible` hard-gates on `backoffUntil` before any caller-mode branch). This is not a request-volume problem, and no local rate counter would have refused that request.

## The change

Wait `Retry-After + RETRY_AFTER_MARGIN_S` (900) so the retry lands in the
empty band, and confine the trust bound so it can never shorten that wait.

```python
if trust_expires_in_s is not None and trust_expires_in_s < retry_after_s:
    asked = min(retry_after_s, asked)
return max(asked, computed)
```

**Absolute, not proportional.** An earlier revision used
`Retry-After × 1.25`. The margin has to land past a FIXED deadline, and
`Retry-After` is a countdown to it, so what the server reports depends on when
we ask. The budget is account-scoped, so a second machine polling into a block
another one opened sees only the remainder — 35 of 72 observed 429s were
mid-block. A fractional margin decays toward zero as the deadline nears: a
1800s remainder landed +450s and a 900s remainder +225s, both inside the
measured +2s..+716s band. An absolute margin does not decay. It applies only
above `BACKOFF_CAP_S`, so short blocks (measured accurate on 2026-07-06:
`Retry-After: 300` meant 300s) are not inflated.

**The trust bound must never shorten the wait.** A 429 does not move the
account's real windows, so `last_good` stays a valid lower bound until its
soonest reset (or the age ceiling). When that trust expires BEFORE the
server's deadline, waiting the extra 900s buys blindness and no freshness, so
the ask is trimmed back to the deadline. When it survives to the deadline the
margin is worth its price. Measured, blind seconds under the re-block model:

    scoped reset   base    unconfined trim   confined (here)
    +3700          4042         4095             800
    +4000          3885         4095             500
    +4400          3675         4095             100

The floor is never crossed: exhaustive sweep over 11,664 (ask, trust, failure
count) combinations, plus a differential sweep against `trust=None` across
~52M points — the trim's output is either a no-op or exactly `retry_after_s`,
never below.

`RETRY_AFTER_FLOOR_CAP_S` goes 3600 → 4500 so a normal full-window block is
not clipped back onto the bare deadline it is meant to clear. A pathological
header is still bounded.

## Also: the documented budget scope was wrong

`poll_policy`'s docstring said the budget is **per-access-token**. On 2026-07-28 a freshly minted token was blocked 135s after issue, which a per-token counter cannot produce. The scope is the **account/org**.

Consequences worth having written down:
- Re-authenticating does not clear a block.
- Two machines holding *different* tokens for one account still compete for one budget, which is exactly what #152's AIMD converges.

Nothing in this PR depends on which scope is right: the wait comes from the server's own deadline, so it is scope-agnostic either way. Comment-only change, plus the log wording (`per-token usage budget` → `usage-endpoint budget`) so the message does not imply a re-login would help.

## Defects found and fixed during review

Each reproduced against this branch first; each fix mutation-checked by line
number.

- **The trust bound made waits SHORTER than the server asked.** The first
  revision trimmed unconditionally, granting 8 fetches per block against
  base's 1. Floored at the ask.
- **The floor still landed inside the re-block band.** Trimming whenever a
  trust bound existed put waits at +100..+800 past the deadline — the exact
  band. Confined to `trust < retry_after`.
- **The scoped-window bound never saw the configured models.** `entries()`
  honours per-model windows when it bounds trust, so a scoped window that ENDS
  the trust must be visible to the backoff too; `models` defaulted to `()` and
  the margin overshot exactly the window that had already killed the data.
- **Three lines had no test.** Mutating `<` to `<=`, dropping the `min()` that
  carries `RETRY_AFTER_FLOOR_CAP_S` (an 86400s ask then produced an 86400s
  wait), and cutting the `models=models` call-site wiring each left the whole
  suite green. All three are now pinned.

## Known limit

The margin is bought once per re-block chain, not per block. `trust_left`'s
age half is anchored to `fetchedAt`, which a failure never advances, so each
wait is deducted from the next block's budget: measured over an 8-block chain,
7 of 8 head retries land exactly on the deadline (base: 8 of 8). Head still
wins decisively on the episode model — mean blind 0s vs base's 1890s, recovery
+4500 vs +7524, 1.00 blocks vs 2.09 over 200 seeds — so this is the fix
under-delivering on chains, not a regression.

## Tests

- `test_hour_scale_margin_clears_the_measured_re_block_band` — the hour-scale
  block that produced the evidence clears the whole 900s band.
- `test_margin_survives_a_mid_block_observation` — a remainder does not decay
  the margin.
- `test_the_trust_bound_never_shortens_the_wait_below_the_ask`.
- `test_the_trim_never_lands_inside_the_re_block_band`.
- `test_the_trim_is_confined_to_trust_expiring_BEFORE_the_deadline` — the
  `<` boundary, at `trust == ask`.
- `test_the_trim_never_lifts_the_floor_cap`.
- `test_the_collector_hands_the_trust_bound_its_configured_models` — drives
  `_collect_usage_entries`, the only path production reaches.

Full suite on this head: **1731 passed, 3 skipped** in 6.2s.

## Corrections made while building this, each reproduced first

- **The "pre-existing, not a regression" claim was arm-specific.** Its proof
  quoted a table byte-identical to upstream — but every row was the
  non-429 arm, where both trees take `TRUST_MAX_AGE_S` (3600) and the table is
  identical by construction. The arm this PR changes does regress: PR cap 4500
  vs upstream 3600 gives `blind = age_at_fail − (ceiling − park)`, onset 900s
  earlier and +900s flat thereafter. Comment corrected; the pinning test now
  covers the 429 arm (it had used `http-500` on all nine cases).

- **A comment claimed "exactly 1 test dies without the PARK BOUND".** Measured:
  **9**, including the test the comment sits inside.

- **The `EngineHarness` isolation guard was not load-bearing.** It scoped on
  `$XDG_DATA_HOME`, which `get_backup_root()` reads only on Linux/WSL, so the
  test could not fail off Linux. Replacing it with `Path.home()` was also wrong
  — XDG takes *precedence* over `Path.home()` there, so with XDG ambient two
  harnesses collapsed onto one store. Both patches are now kept: removing the
  XDG one kills 1 test, removing `Path.home()` kills 2, removing both kills 5.

- **`Retry-After: 0` was parked 300s on non-429 responses.** The saturated-budget
  edge was measured on the usage endpoint's 429s only; a Cloudflare 503 saying
  "retry now" took the same 300s floor. Now gated on `rate_limited`. Blast
  radius measured with six controls — exactly one cell moves (non-429 + ask 0:
  300s → 30s); the 429 edge, the no-header case, non-zero asks on both arms, and
  the exponential curve at 5 failures are all unchanged.

- **Lapse figures re-derived three times, independently.** The published numbers
  depend on which prior-block deadline the method means. Under the recorded
  reading (the block's *opening* observation) it is **20 of 35** with 3 negative
  gaps; under a last-observation reading, 21 of 36 with 2. The two differ on
  exactly one gap. `RETRY_AFTER_MARGIN_S` = 900 clears the +887s band edge by
  13s under either.

Mutation coverage: 16/16 production mutations killed, positive control
included; every figure in the `RETRY_AFTER_MARGIN_S` comment re-derived from
the log and reproducing at every clustering tolerance from 5s to 300s.

## Coordination

This touches the same two files as #196, but not the same places: #196 edits
`FetchRecord`/`UsageEntry`/`entries`/`record`, while this only edits the module
constants and `_failure_backoff_s`. Same for the tests.

Rebased onto the current `main` (`3d9f596`). The xdist parallelisation, the
real-Keychain serialisation hook and the `uv sync --locked` CI gate that this
branch used to carry all landed upstream via #209, so those commits are gone
from the range — `pyproject.toml`, `tests/conftest.py`, `.github/workflows/ci.yml`
and `uv.lock` are now byte-identical to `main` here.


realiti4:
Thanks for the PR, it looks good and I've merged it!

PR #199, opened 2026-07-29
TUI: unattended auto-switch — resume LIVE, one engine per machine, and reachable empty slots

codeslake:
The auto-switch view is usable without babysitting it, and the account list is reachable end to end. Five things, all TUI-facing.

> **Reading the diff:** this branch has #196 and #210 merged into it, because sections 2 and 3 build directly on their credential-store work and the three conflict in `switcher.py` otherwise. The five sections below are what is *new here*; the rest of the diff belongs to those PRs and lands with them. Reviewing this alone against `main` will show their changes too.

## 1. Resume LIVE instead of re-asking — `cswap tui --auto`

Going live took a keypress and a modal confirmation *every* launch, so a TUI that restarted (deploy, reboot, tmux respawn) sat in dry-run watching an account burn past its threshold. `cswap tui --auto` now opens the auto view with the engine LIVE.

**The flag is the only thing that does it.** An earlier draft of this branch persisted a confirmed go-live in a setting, and this section used to be written around that. The setting is gone: `start_live` is set at exactly one place, from the flag, and nothing reads it from disk. The code says why in its own comment — *"reaching the auto view from the menu watches without switching — opening a view must never begin switching accounts."* A bare `cswap tui` lands on the dashboard; see 4.

## 2. Only one LIVE engine per machine

Two LIVE engines race. `_perform`'s state lock serializes the *write*, but its cooldown recheck is consulted only for the `proactive` and `consume-first` triggers — `at-limit` and `failover` skip it by design, since an exhausted account cannot wait out a cooldown. So both engines decide independently and the second switches away from what the first just chose.

Measured with two TUIs on one machine: the one making the switches was not the one being watched.

A LIVE engine now holds `flock` on `<backup_dir>/.auto-live.lock` for its lifetime. A second one **demotes itself to dry-run and says so once** — not a refusal, because a second TUI must still show its dashboard, and an error dialog would not tell you which instance owns the engine. `flock` rides the open file description, so a killed holder frees it with no stale-pid sweep. Explicit dry-run engines never take it: a watching TUI must not lock out the one doing the work.

The warning is emitted from `tick()`, not `run_loop()` — `cswap auto --once` (cron) never enters the loop, and a cron tick that silently demoted itself is the same invisible failure in a smaller shape.

`stop()` only *asks* the loop to exit, and a tick already in flight runs to
completion (Textual's `run_worker` is not exclusive). Every `stop()` caller
builds the successor immediately, so releasing the lock there handed it to a
successor that could switch while the predecessor was still inside `_perform`.
`_perform` re-checks `_stop` under the state lock.

That covered the switch but not the mutations before it:

    _freshen_target consults _stop? False   # POSTs a one-time refresh grant
    _quarantine     consults _stop? False   # writes quarantine state
    _perform        consults _stop? True

so a departing engine could consume a refresh grant, or quarantine a slot, for
accounts its successor already owned. The tick now stops where dry-run does,
for the same stated reason: freshening is a mutation.

**A demotion writes nothing.** Going live is per-launch, so a
demoted TUI persisting `false` does not decline to record consent — it revokes
the LIVE holder's. Measured: TUI#1 goes live (`true`), TUI#2 presses `l`, is
demoted, persists `false`, and the next launch starts dry-run for a user who
never asked. The refusal is about the lock, not about consent.

## 3. Reach a slot that has no stored login

`switch_to` refused an empty slot with *"has no stored credentials. Re-add with: `cswap --add-account --slot N`"* — advice that cannot be followed. There is nothing to add until a login exists, and a login lands in whichever slot is **active**. So the one slot you needed to reach was the one slot you could not. This bites right after an import, where the account list arrives and credentials deliberately do not.

It now lands there, **logged out**. The current login is backed up first (that step already ran before the refusal), then the live credential is cleared and `activeAccountNumber` moves. Clearing rather than leaving it is the point: a slot that reads #4 while #2's token is still serving lies about whose quota is burning. `/login` then writes to the slot you are on, and `cswap add` captures it. The result carries `needsLogin: true`.

macOS is checked **first**: a locked or non-GUI Keychain reports "no credentials" exactly like a genuinely empty slot, and on macOS the live credential lives in that same Keychain — logging out on that guess would destroy a working login with nothing on disk to restore from. That path still refuses.

The check is `_keychain_blind()`, and it PROBES when the capability cache is unset rather than reading `None` as "fine" — `--force` skips the live-identity prefetch entirely (measured: zero keychain calls before this guard) and an `.enc`-satisfied read short-circuits before the keychain, so a locked keychain was otherwise indistinguishable from a probed one. It answers through #196's `_keychain_unreadable`, which excludes a file mode cswap pinned deliberately: nothing failed there, so an empty read means the slot really is empty and the switch must land rather than refuse.

**Two credential defects the landing path had to close.** An UNREADABLE live
credential (`value is None`, not `""`) was deleted without ever being stashed;
and a FAILED clear still popped `oauthAccount`, so the landed slot wore the
active mark over the departed account's live token. Both now refuse.

**The post-clear check was blind under a pinned file mode.** Its two witnesses
both come from `_read_active_credentials`, which never asks the Keychain there:
`_use_keychain()` is False and `_keychain_unreadable` is False by construction
— nothing failed, we chose the file. Measured, reads succeeding throughout:

    after the pin     use_kc=False  unreadable=False  file_ours=True
    post-clear read   value=''      unavailable=False -> guard passed
    Keychain residual still present: True

The switch then completed and marked the landing slot active while another
account's token was what Claude Code would use — the exact invariant the guard
exists to enforce. `delete_password` returns only on rc 0 or rc 44 and raises
otherwise, so a return is proof; `_delete_active_keychain_entry` and
`_clear_oauth_credential` now report it and the guard reads it.

## 4. A bare `cswap tui` never enters the auto view

An earlier draft of this branch persisted the go-live, and a plain `cswap tui` then opened straight into the auto view with the engine LIVE. That made `--auto` meaningless — both spellings did the same thing — and the persisted flag was reachable from a single confirmation. Opening a dashboard is recoverable; auto-switching accounts because of a checkbox ticked once is not. The setting was dropped rather than defended.

## 5. Reset countdowns on Next-best, from one reading

Next-best candidates now show `5h(⟳1h38m):24% · 7d(⟳4d8h):100%`, so you can see *when* a spent peer comes back without leaving the view. Both surfaces render a window through the same `data.window_chip_label`, so the auto view and the dashboard can no longer disagree about the same number.

## 6. The dashboard mini line shows what the card shows

The one-line minis iterated 5h/7d plus scoped windows at or over their cap,
and everything else fell through to `usage unknown` — while `usage_rows`, in
the same file, rendered those accounts in full one screen over.

Two shapes reached the fallback. An extra-usage (pay-as-you-go) account has
neither a 5h nor a 7d window, only `spend`. An account whose only window is a
per-model limit *below* its cap has nothing the maxed loop counts.

Spend is now a row of its own rather than a last resort, so a budget cannot
hide behind a window that reads healthy, and it dims with the rest once the
measurement is older than `STALE_OK_S`. Both surfaces take the row from
`usage_rows`, so the auto view and the dashboard cannot spell the same amount
two ways.

```
window + spend   ' 1  user1@example.com  [personal]   5h(⟳10d6h):25% · 7d(⟳13d4h):10% · $$ 51% · $10.29 / $20.00'
scoped only      ' 1  user1@example.com  [personal]   Fable 99%'
```

Display only: spend never enters the ranking. The line is longer, and the
accounts panel wraps rather than truncates (`height: auto`), which is
pre-existing behaviour — the same account wraps identically at width 80 with
the spend row removed.

---

**Behaviour change, deliberate:** issue #41 established that `switch_to N` must keep failing on a broken slot. Section 3 changes that for the *empty* case. The two tests encoding the old contract were rewritten rather than deleted.

**Scope:** the cloud pin is NOT in this PR. This branch carried a whole copy
of `pin.py` (absent from `upstream/main`, so it was an addition rather than an
edit) left over from before the pin was split out; it had drifted from the
generation #210 carries, so keeping both meant losing one. Removed here in
full — module, tests, the `cswap pin` subcommand, and every TUI/CLI surface
that rendered or set it. #210 owns all of it.

**Testing:** full suite on this head: **2303 passed, 3 skipped** (identical parallel and serial, so no order dependence). Coverage for
the LIVE lock (including a cross-process check — the two-TUI shape unit tests
cannot reach), the stopped-engine race on both the switch and the mutation
paths, the consent write, empty-slot switching, the locked-Keychain refusal,
the pinned-file-mode blind guard, and the opening-screen rule.

Every guard mutation-checked by line number with the landing asserted:
always-acquire, never-release, dry-run-also-locks, skip-the-logout,
drop-the-Keychain-guard, the delete reporting success on failure, the guard
dropping `residual_gone`, the freshen `_stop` gate, the demotion check, and
restoring the old opening-screen condition each fail exactly the test that
names them.

The blind-guard test is pinned on the no-failure path (the managed-key path
pins after a write that SUCCEEDED) rather than through a failed write. Both
reach the same state, but only the former survives beside #196, which records
a separate `_keychain_op_failed` — a test written on the failure path passes
in the merged tree for a reason that is not this guard. Caught by the
integration rebuild: green on either branch alone, red merged.

Live-fired on Linux and macOS: with a TUI LIVE, `cswap auto --once` emits the
demotion warning and does not switch; with it stopped, the same command
acquires cleanly.

## 7. Also in this diff, and not covered above

Four changes that ship here and had no paragraph. Naming them so a reviewer is
not surprised by them in the file.

- **A running engine now moves off a disabled active account.** `cswap disable`
  gains a trigger it did not have, which is a semantic change to that command
  rather than to the TUI. It is the only item here that also touched the README.
- **`autoswitch.decisionLog`** — an opt-in bool, *"Record why each tick switched
  or did not, to its own log file"*. A new setting and a new log file.
- **A `SIGINT` handler in `cswap auto`**, so a Ctrl-C during a tick ends the run
  rather than leaving the loop mid-decision.
- **`switch_off_at_limit_account` / `current_at_limit`.** This one deserves the
  maintainer's judgement rather than a defence: it has **one in-tree caller**,
  and its docstring is about an out-of-tree proxy that reads a 429 on
  `/v1/messages` and uses the symbol's presence as a capability probe. It is
  generic in form, but it is separable from everything else in this PR and
  would review faster on its own branch. Say the word and I will split it out.

## Round 6

Round 5's own fixes were the subject, and two of them had defects.

**The LIVE-lock demotion was reporting the wrong cause.** Round 5 stopped
`__init__` raising on an unwritable `backup_dir` by swallowing the `OSError`
and taking the CONTENTION branch. That is the right outcome and the wrong
sentence: a writable dir with an unwritable `.auto-live.lock` — a root run, a
tight umask — then announced that another engine held the lock. There is no
other engine, nothing was logged, and `_retry_live_promotion` re-raised the
same errno every tick, so every tick decided to switch and did not.
Reproduced: `PROBE-C EVENT: ConfigWarningEvent | another LIVE auto-switch
engine is already running`, with `active still: 1` after two ticks.

The errno is recorded now and said, and a cause that CHANGES is announced
again rather than the first one being repeated forever.

**`cswap auto`'s banner read `args.dry_run`, not `engine.dry_run`.** A demoted
run announced itself as switching. The identical one-line fix was already in
`tui/autoview.py` with a comment saying why, and was missed here.

**`_keychain_blind()`'s probe had no test** — deleting its two lines left the
suite green, on the path whose stated stake is clearing a working login to
reach a slot that was never empty. Two cases now: the probe must fire on an
unset cache, and must NOT fire on a filled one.

**The at-limit docstring named the wrong lock.** Round 5 replaced "the LIVE
lock protects this" with "the state lock serializes the write" — and
`.autoswitch_state.lock` is the engine's own; this file never opens it.
`switch()` takes `FileLock(self.lock_file)`, `claude_credentials_lock()` and
`claude_config_lock()`, and every one of them serializes the WRITE, not the
DECISION. The conclusion was right, the cited mechanism was not, which is the
shape the round-5 fix existed to remove.

## Round 7

**A rotation is not a login, and five compares said it was.** The server
re-mints `refreshTokenExpiresAt` on every refresh of one lineage, with
sub-second jitter of arbitrary sign (427 ms measured between the Keychain
item and the plaintext file on one host). Five places ordered logins by
comparing that stamp exactly: the two-store read in `credentials.py`, the
healthy-slot and dead-slot arms of the reader-side adopt, the switch-time
foreign adopt, and the quarantine's "newer login" release. On the measured
sign the two-store read served the file's superseded generation over the
Keychain's current one, which produced an `invalid_grant` strike on a
healthy account, a quarantine, and a profile-401 loop while Claude Code ran
fine on the Keychain generation; on the other sign the same compares
overwrote a healthy backup with a spent generation, adopted a foreign
credential, or refused a dead slot its own cure. One predicate now decides
all five: `credentials.newer_login(a, b)` is true only when `a` is later by
more than 5 s (`_LINEAGE_STAMP_JITTER_MS`); within that window the stamps
name one login, and for the two-store read the later `expiresAt` picks the
newer generation of it. Each site is pinned at ±427 ms with an 8 s control,
and the boundary is exact at 5 000 / 5 001 ms.

**The docstring above `_refresh_expiry` still states the disproven premise**
("it ORDERS LOGINS and never two generations of one"). Left as is in this
round; the predicate beside it is the contract now.

https://claude.ai/code/session_01QFPGPPSYmx8NNGqbEsSpNc


codeslake:
Heads-up on this one now that #196 has landed, plus **one design question that decides how I rebase it** — I'd rather ask than pick.

## Where this PR stands after #196

#196 and this branch were developed in parallel and this branch carries a **stale copy of #196's lineage** — 86 of #196's 96 commit subjects also appear here. Measured:

```
merge-base(196, 199)          ba6f3f9   (not a #196 commit)
consume_backup_grant at base  0 files   <- the base LACKS the gate
                     in 199   3 files
                     in 196   3 files
```

So both branches added it independently, and merging the post-#196 `main` into this branch conflicts in **61 hunks** — almost all of it git reconciling old-#196 against final-#196, not this PR's own work. Note `git cherry main <this>` reports all 278 commits as new: #196 landed squashed, so patch-id matching can't drop the duplicates and neither rebase nor `merge-base --is-ancestor` will tell you what's already in.

I'm rebuilding it as **`main` + only this PR's own work** rather than resolving 61 hunks by hand. That will shrink the diff a lot and make it reviewable.

## The question: `_entry_token_dead` on an UNREADABLE second source

Both versions have a written rationale and they protect complementary risks.

**`main` today** returns `True` — hold the strike:

> the caller's `elif` spends that answer on `clear_dead_token`, which zeroes authDeadStrikes AND struckFingerprint in the PERSISTED store. One momentary lock would un-quarantine a genuinely dead account permanently.

**This PR** widens the return to `bool | None` and refuses to answer:

> guessing `True` condemns an already-healed slot; guessing `False` lets the healed-strike-clear branch permanently erase a strike that may still hold. `None` refuses to guess.

They're both right about their own failure. `True` never erases a quarantine on a transient Keychain lock; `None` never condemns a slot a re-login already healed. The tri-state covers both, but only if the callers handle `None` — which is the part this PR adds, and it changes behaviour you reviewed and merged deliberately in #196.

**That's why I'm asking instead of resolving it in a merge commit.** Slipping it in as a conflict resolution would override a merged decision somewhere no review would look.

Measured, running this PR's `tests/test_degraded_active_read.py` (473 lines, doesn't exist on `main`) against `main`'s implementation:

```
14 failed, 4 passed
  11  TypeError      signature drift only — the extra param. Mechanical.
   3  AssertionError real behaviour main does not have:
        - post-fetch call site does not pass _active_read_degraded
        - an already-healed active slot IS condemned on a degraded read
        - the re-login warning does not return once the keychain answers
```

I also checked whether `main` reaches the same protection another way — it doesn't: `_active_read_degraded` exists (`switcher.py:2656`) but its two uses guard `_resync_rotated_backup` and the active-usage defer, and none of the three `_entry_token_dead` call sites (4638, 4711, 4776) pass the flag.

**Which do you want?**

1. **Keep `main`'s binary `True`** — I drop the tri-state, rebase the rest, and rewrite these tests to assert `main`'s contract. Smallest diff, and the healed-slot case stays a known cost.
2. **Take the tri-state** — I port `bool | None` plus the caller handling, as its own commit with the rationale in the message, so it's reviewable on its own rather than buried in a rebase.

Happy either way — 2 is more code and closes the healed-slot case; 1 is less and keeps what you already reviewed. One note if you pick 2: `test_post_fetch_call_site_passes_the_flag` asserts that an *argument is passed*, which pins implementation shape rather than behaviour, so I'd re-express it as a behavioural test regardless.

No rush — this PR isn't going anywhere until you've got time, and nothing here is blocking you.

codeslake:
Correction to the comment above, on the framing — the question stands, my premise for it did not.

I wrote that the binary `True` is behaviour **you** reviewed and chose in #196, and that changing it would override a merged decision of yours. That is wrong, and I should have run one command before saying it.

Every commit that has ever touched `_entry_token_dead` is mine:

```
$ git log --all -S'_entry_token_dead' --format='%h %s' -- src/claude_swap/switcher.py
0ccda15 Never consume a stale refresh token... (#196)
65b0d60 refactor(switcher): _entry_token_dead takes what it uses, not a 7-tuple map
bab239d fix(import): the heal must ask the same question the collectors ask
07a8f90 fix(gate): close all 14 ultra-review findings on the consume path
```

and the `unreadable → True` branch specifically came in **with** #196 (`f6e6f7d`, "three sites read 'unreadable' as 'absent', each destructively") — it was not in `main` beforehand. The rationale comment I quoted at you as yours is also mine.

So this is not "your decision vs my proposal". It is **two versions of my own code**, developed in parallel on two branches, one of which happened to land. You merged #196 as a whole; you never adjudicated this branch, and I framed it as though you had — which loads the question in a way that could push you toward defending a choice you never made.

The actual question is smaller and I'd rather you answer that one:

> Slot's second credential source can't be read (Keychain locked). Is the account dead? `main` says yes, hold the quarantine. #199 says unknown, refuse to guess.

Trade-off unchanged and still real: `True` never erases a quarantine on a momentary lock; `None` never condemns a slot a re-login already healed, but costs the caller changes to handle a third answer. My measurements in the previous comment stand — those were run, not inferred.

Still happy either way, and there's genuinely no rush.

codeslake:
Two updates on this branch since my last comment, and **the question above is still open** — I've repeated it at the bottom so it isn't buried.

## 1. New commit: a disabled ACTIVE account was a place the engine sat forever

`dd2f957`. Reported behaviour: pick a disabled account by hand in the TUI, turn auto on, and the engine stays on it indefinitely.

**Cause.** `_tick_inner` asked how much headroom the active slot had, and never asked whether that slot was *allowed to be* the active slot. `is_account_disabled` was consulted when choosing a destination and nowhere on the departure path — so a disabled account was excluded as a candidate but perfectly acceptable as a place to remain.

**The fix is a new trigger**, `disabled-active`, and it departs on the first tick. Not after `unhealthy_ticks` (default 3): that counter exists for *transient* unreadability, where waiting is what stops a momentary blip from causing a switch. "Disabled" is neither transient nor a guess — it is a flag we wrote to our own `sequence.json`, so there is nothing to confirm and three ticks of confirming it is three ticks of sitting on an account the user removed from rotation.

Three smaller holes came out of the same read, each on the departure side:

- the api-key gate suppressed departure from a disabled active when `include_api_key_accounts` is off — so a disabled api-key slot was doubly stuck;
- the escalate leg had no disabled term, so the engine could decline to look past a full account it was not allowed to be on;
- the no-return bar treated the departure as "this account is spent", which meant **re-enabling the account did not bring it back**. Released specifically for a `disabled-active` departure.

Ships with 5 test classes (`tests/test_autoswitch.py`, +338). Each has a labelled CONTROL plus an instrument guard asserting the setup actually happened, because three of these assertions pass trivially if the first tick silently no-ops. CI green on `dd2f957`.

One behavioural note for the docs: "a disabled account is still a valid explicit `cswap switch` target" now holds **only while no engine is running**. With auto live, an explicit switch to a disabled account is undone on the next tick, by design. `README.md` and the `models.py` docstring say so.

## 2. This branch was force-pushed — history rewritten, trees untouched

Author/committer identity and message metadata were normalized across the branch. If you have a local copy, it will not fast-forward; re-fetch rather than merge.

Measured, not inferred — the rewrite changed no content at all:

```
pre-rewrite tip  d12ccc2  tree 6bfd849710cf
post-rewrite     dd2f957^ tree 6bfd849710cf   <- identical
CONTROL: dd2f957 itself does change the tree (5 files, +425/-9),
         so the comparison above is not vacuous
298 commits rewritten (1c3122e..dd2f957)
```

`upstream/main` was not touched and is still an ancestor. #216 got the same treatment (5 commits, tree `b90e13f015c3` identical either side).

## Still open, unchanged

> A slot's second credential source can't be read (Keychain locked). Is the account dead? `main` says yes, hold the quarantine. This PR says unknown, refuse to guess.

Option 1 is smaller and keeps `main`'s binary `True`; option 2 ports the tri-state plus the caller handling as its own reviewable commit. The branch currently carries option 2 because the merges had to resolve *something* — say the word and I'll strip it. Still genuinely no rush, and nothing here blocks you.

codeslake:
Two commits on this branch since my last comment. The open design question above is unchanged and still yours to answer — this is a status note, not a nudge.

**`64679ded` — leave an account measured at-limit off the poll**

A session sat in "Session limit reached · Retrying in 1h" while its *active* account had 83% headroom. Probed off `/api/oauth/usage` at the time: 5h 20%, 7d 56%, scoped 0%, and `seven_day_opus`/`seven_day_sonnet` both `null` — there is no hidden per-model window. The 429 the client was backing off on belonged to a *previous* account, and Claude Code arms that wait client-side from `error.rateLimits.resetsAt`, so once armed nothing outside the process can clear it.

`switch_off_at_limit_account(switcher)` gives an out-of-band observer a way to say "this account is spent" without a percentage. Two things it had to work around, both measured:

- **Not an engine tick.** A LIVE engine holds `.auto-live.lock` for its whole lifetime, so a second engine demotes itself to dry-run and switches nothing. The first version of this was `cswap auto --once --at-limit` and it would have no-opped on every real machine while passing every test that did not hold the lock first. `at-limit` does not need a tick anyway — it already skips the cooldown, the no-return bar and hysteresis.
- **`strategy="best"` alone refuses to move.** A rate-limited account's pct is frozen at `last_good`, which is a usage *lower* bound and therefore a headroom *upper* bound: right for a destination, wrong for the account being left. The slot known to be dead then outranks every healthy candidate and `best` returns `stay` — 20% beating 56% in the shape above. `current_at_limit` forces `current_headroom = 0.0`, so any candidate with real headroom wins and a field of exhausted ones still returns `exhausted`.

Mutation-checked: deleting the override, moving it below the `is None` guard, and dropping the flag from the seam each kill a distinct test.

**`3d6a2913` — gate the decision log live instead of binding it**

Follow-up to the `autoswitch.decisionLog` work. The bind hook existed only because the sink read a snapshot, so it needed calling at construction *and* from `_retry_live_promotion` 1,900 lines apart, plus a matching clear in `stop()`. Reading `settings.decision_log and not self.dry_run` at the emit makes all three unrepresentable rather than fixed. Tests fold with it — asserting the private `_decisions` was testing the memo, not the behaviour, so one test now walks demoted → stopped → promoted against the file itself. Every mutation still dies: the setting gate, the dry-run gate, `getLogger` instead of `Logger`, dropping `event.ts`, and snapshotting the gate at construction. −44 lines.

**Suite: 2268 passed, 3 skipped, on `3d6a2913`.**

codeslake:
Retriggering CI: this fork PR's last commit (19ad7014) never received a workflow run (0 check-suites) — closing/reopening to re-fire the pull_request event without changing the head.

Issue #192, opened 2026-07-28
bug: shared session items link to ~/.claude/<item> unresolved, so Claude Code's one-hop atomic write silently detaches dotfiles-managed settings


charlie0228:
### What happened?

When an item under `~/.claude` is itself a symlink (a common dotfiles setup — mine has `~/.claude/settings.json` → a git-tracked file in my config repo), `cswap run` shares it into the session profile by linking to the **unresolved** path:

```
<profile>/settings.json  →  ~/.claude/settings.json  →  <dotfiles-repo>/claude-code/settings.json
```

That link-to-a-link is what breaks. Claude Code's atomic settings write (temp file + `rename`) resolves the chain **only one hop**, so the rename lands on the *middle* link and replaces it with a regular file. `rename()` overwrites the directory entry and never follows a symlink, so the chain is silently severed — the write itself succeeds, the content is correct, and nothing warns.

Consequences once that happens:

- the dotfiles repo silently stops receiving any further settings changes
- the next dotfiles sync run overwrites the local file from the repo, **discarding** everything changed in the interim (in my case `model` and `effortLevel` had diverged before I noticed)
- it is invisible: no error, and the CLI keeps working normally

This is the Claude Code side of the bug — I've filed the detail under anthropics/claude-code#78162 (that issue covers the same one-hop resolution, originally reported as an EROFS/EACCES failure when the intermediate directory is read-only; when it's writable you get this silent variant instead). Upstream shows no movement in ~2 weeks, and cswap can sidestep it entirely.

**Expected:** shared items link to their fully resolved target, so the chain is only ever one hop deep and Claude Code's one-hop writer lands on the real file.

**Suggested fix** — resolve once in `SessionManager`'s share loop (`session.py`, the `for name in active_items:` block) and use it for both the link comparison and the link creation:

```python
link_target = src.resolve() if src.is_symlink() else src
```

then `dest.readlink() != link_target` and both `dest.symlink_to(link_target)` call sites. I deliberately left `src` itself untouched so `_prepare_history_share(src, ...)`, `src.is_dir()` and `shutil.copy2(src, dest)` keep their current behaviour. I've been running this locally and the profile link now points straight at the repo file; the middle link has stopped being clobbered.

Happy to open a PR — flagging that #163 also touches session sharing, so I'd rather rebase onto whatever lands first than create a conflict.

### cswap version and install method

cswap 0.22.0, installed with uv

### Operating system

macOS

### How do you run Claude Code?

[...]

PR #201, opened 2026-07-30
fix(settings): atomic_write_json writes through a symlink, never over it

codeslake:
Found in the field, on three machines, chasing a feature that silently stopped applying.

## The bug

`atomic_write_json` ends in a rename of its temp file onto the destination. A rename swaps a directory **entry** and never follows a symlink — so when the destination is a symlink, the rename **detaches** it. The write succeeds. The content is correct. The link's target just quietly stops receiving updates.

That is invisible until something reconciles the link. Where these files are managed by a dotfiles tool that links them into a config repo, the sequence is:

1. a write detaches the link; the file is now a plain local file
2. every later write lands there — the repo copy is frozen
3. the next deploy restores the tracked copy, **discarding everything written since**

What we actually observed: a settings section written by one feature vanished, while the runtime state that feature had already started kept running — so the daemon was alive, its wiring was in place, and nothing looked broken; the feature had simply stopped being configured. Three machines, three timestamps, each matching that machine's last deploy.

## Why this is not a one-off

This is the same shape as #192 / PR #193, which fixed `session.py`'s own writer. `atomic_write_json` was never touched, and it is the **shared** writer behind eight call sites:

- `settings.py` — `save_settings`, `set_setting`, `unset_setting`
- `autoswitch.py` — state file (2 sites)
- `usage_store.py` — the usage store
- `credentials.py` — the unclaimed-credential manifest
- `session.py` — session profile config

Any of those can detach a managed file today. Nothing about it is specific to the feature that surfaced it.

## The fix

Resolve the path first and rename onto the **resolved** target: the link survives and its target is updated. Two details kept deliberate:

- **A dangling link still writes where it points.** Linking a path is a request to write there; a missing target is a directory to create, not a reason to clobber the link.
- **The temp file is created beside the resolved target**, so the rename stays on one filesystem and remains atomic. Writing beside the *link* and renaming across a filesystem boundary would trade this bug for a worse one.

## Tests

Three regressions: the link survives a write and its target is updated; a dangling link writes through; the plain-file path is unchanged.

Full suite green minus this checkout's pre-existing baseline (44 tui-async + 11 ANSI, identical on `main`).

### DIFF
--- a/src/claude_swap/settings.py
+++ b/src/claude_swap/settings.py
@@ -449,20 +449,27 @@ def atomic_write_json(path: Path, data: dict) -> None:
     **Writes THROUGH a symlink, never over it.** A rename swaps a directory
     ENTRY and does not follow links, so renaming onto a symlinked path
     DETACHES the link: the write succeeds, the content is right, and the
-    link target silently stops receiving updates. Where these files are
-    dotfiles-managed (a link into a config repo), the next deploy restores
-    the tracked copy and every change written since the detach is gone —
-    measured in the field as a settings section that vanished while the
-    state it configured stayed live. Same shape as #192/#193, which fixed
-    ``session.py``'s own writer; this is the shared writer behind eight
-    call sites. A DANGLING link still writes where it points (that is what
-    linking it asked for), and the temp file is created beside the
-    RESOLVED path so the rename stays same-filesystem and atomic.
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks a lot for addressing this and for the fix. Looks good, merged it!

Issue #194, opened 2026-07-28
feat(autoswitch): treat accounts with extra-usage headroom as last-resort targets instead of "exhausted"


smoochy:
## Problem

An account whose subscription window is at 100% but which still has **extra usage credit available** (`extra_usage`, surfaced since #39 as the `$$:` row) is treated as unconditionally exhausted by autoswitch. It can never be a rotation target, even when it is the only account able to serve requests.

With two accounts where one is subscription-exhausted but has spend headroom, `auto` reports `BLOCKED` / all-exhausted and parks — while a usable account sits right there.

## Reproduction

Two OAuth accounts:

```
1: account-one
   ├ $$:  79%   $78.83 / $100.00      <- ~$21 of extra usage still available
   ├ 5h:   8%   resets 15:59
   └ 7d: 100%   resets Jul 30 17:59   (ahead of pace)

2: account-two  (active)
   ├ 5h:   7%   resets 18:20
   └ 7d:   1%   resets Aug 2 03:00
```

Once account 2 crosses the threshold, `cswap auto --once` exits `3` (`blocked: no viable target / all exhausted`). Account 1 is skipped purely on `sevenDay.pct == 100`, although its `spend` row shows it can still serve paid requests.

## Cause

Candidate viability is decided from `headroom`, which is computed only from the percentage windows (5h / 7d / optional per-model). `src/claude_swap/autoswitch.py`:

```python
candidate_headrooms = [headroom.get(n) for n in oauth_candidates]
truly_exhausted = all(
    h is not None and h <= 0 for h in candidate_headrooms
)
```

`extra_usage` is parsed in `src/claude_swap/oauth.py` and rendered by `switcher.py` / `json_output.py` / `menubar.py`, but never reaches the routing decision. The two `spend` occurrences in `autoswitch.py` (lines ~884, ~1163) are the English verb in comments, not the billing field.

So the model conflates two distinct states:

- **subscription quota exhausted, cannot serve** — correctly non-viable
- **subscription quota exhausted, extra usage active, can serve (billed)** — currently also non-viable

## Why it matters

This is the exact scenario extra usage exists for: continuing past the subscription window. When it is configured and funded, autoswitch strands the user on a saturated account and reports "all exhausted" — which isn't true.

## Proposed behaviour

[...]

Issue #161, opened 2026-07-22
_lockstep_usage_warnings() fails to detect lockstep usage when resets_at timestamps differ by microseconds


djbclark:
## Description

The `_lockstep_usage_warnings()` heuristic in `switcher.py` (line ~3106) builds a comparison key from `(h5_pct, h5_resets_at, d7_pct, d7_resets_at)` using exact string equality on the `resets_at` timestamp fields.

When two slots hold different generations of the same account (the poisoned state from #117), both tokens report the same account's usage from **two separate API calls**. The API returns `resets_at` with microsecond precision (e.g., `"2026-07-25T09:00:00.129393+00:00"` vs `"2026-07-25T09:00:00.340384+00:00"`), so the keys never match and the warning never fires — even when `pct` values are identical and the reset times are functionally the same.

## Steps to reproduce

1. Have two accounts (e.g., `user1@example.com` and `user2@example.com`) in slots 1 and 2 where both tokens actually point to the same underlying account
2. Run `cswap list --json`
3. Observe: both accounts show identical `fiveHour.pct` and `sevenDay.pct`, with `resets_at` differing only in microseconds
4. No `lockstepUsageWarnings` field appears in the JSON payload

## Expected behavior

The lockstep heuristic should detect functionally identical usage data. When `resets_at` values parse to the same ISO8601 instant (within ~1 second tolerance) and `pct` values match, the warning should fire.

## Actual behavior

The heuristic uses exact string comparison on `resets_at`, so microsecond-level differences (which are unavoidable across two separate API calls) prevent detection.

## Suggested fix

Parse `resets_at` into a `datetime` object and compare at second granularity (or coarser), rather than comparing raw string values. The key construction at `switcher.py:3106` should normalize the fields before hashing:

```python
from datetime import datetime
def _normalize_ts(val):
    try:
        return datetime.fromisoformat(val).replace(microsecond=0)
    except (ValueError, TypeError):
        return val
```

## Environment

- claude-swap v0.22.0
- macOS, installed via `uv tool install claude-swap`

PR #202, opened 2026-07-30
Switch before the limit lands: obey the poll plan, and go where quota returns first when nothing is healthy

codeslake:
Two independent bugs that together let a session take a hard rate limit while a peer with quota sat unused. Both measured on a live 3-account setup, one commit each.

Once Claude Code takes that limit, no switch can undo it: its retry timer comes from the rate-limit headers it already received (`URe.error.rateLimits.resetsAt` drives the "Retrying in 2h" banner; the process-wide limit state is only ever updated from response headers, never from the credential file). So both fixes are about not arriving there.

## 1. The loop overslept the poll plan it had just written

`poll_policy` tightens the active account's row to `URGENT_INTERVAL_S` (60s) while it burns near the switch threshold, so the crossing is caught quickly. `_next_delay` always slept `interval_seconds` — 360s by default — so that plan simply could not be honoured.

Measured mid-episode:

```
acct 2 (active): plan=60.0  next_poll_at = -112s   # due nearly two minutes ago
```

while the engine still had minutes of sleep left. The account sat over the threshold doing nothing until the user toggled LIVE off and on, which constructs a fresh engine and ticks immediately — which is why a manual toggle "fixed" it every time and made the engine look dead. It was never dead; it was asleep past its own alarm.

`_next_delay` now shortens a normal-cadence sleep to the row's own `nextPollAt`. **It only ever shortens, and never below `URGENT_INTERVAL_S`**, so it cannot raise the request rate above what the plan already permits: the 429 budget lives in the plan (#245f5ba), and this makes the loop obey that budget rather than override it. A relaxed plan does not stretch the cadence past the configured interval, and a failed store read leaves the cadence alone.

Two of the four tests fail on pre-fix code (393s and 372s where 60s is required).

## 2. With every account above the threshold, there was nowhere to go

Landing must be healthy — an account at/over the threshold would re-trigger on the next tick — so `_rank_candidates` skips any candidate above the line. When *everything* is above the line that rule has no answer, and the engine holds still.

Measured: three accounts' 5-hour windows at 100 / 99 / 95%, threshold 90. Zero qualifying candidates, so the engine waited while the active account burned to 100% and the session hit a hard limit. The peer whose window reset in **8 minutes** was never tried.

In that state the goal changes from "most headroom" to "soonest back": rank by the reset of each account's **binding** window — the one actually holding it back, the same window `account_headroom` measures, so ranking and headroom can never disagree. In the measured case that picks 8 minutes instead of 2 hours.

Deliberately narrow:

- engages only when the active account *and* every measured candidate are at/over the threshold, so one healthy peer still wins the ordinary way;
- an account at its limit (`h <= 0`) is still never a target, so genuinely-all-exhausted still reports exhausted and parks on the reset;
- an unknown reset sorts **last**, not first, so a snapshot nobody refreshed cannot masquerade as "back immediately".

The percentage-point hysteresis is replaced in this state only, not dropped. It is unmeetable by construction here (nothing can be 10 points better when everything is within a few points of its limit), so it would block the escape rather than protect anything. The flap it exists to prevent is prevented on the axis being ranked instead: a target must come back `RECOVERY_HYSTERESIS_S` (5 min, comfortably over one poll cycle) sooner than where we are, so the reverse move never qualifies. Two tests pin that — 60s sooner does not switch, an hour sooner does.

## Testing

12 new tests in `tests/test_autoswitch.py`; 142 pass in that file, full suite green. Mutation-checked both ways: reverting `_respect_poll_plan` fails 2, disabling the escape fails 4.

Running on three machines (linux + two macs) since the fix.

### DIFF
--- a/src/claude_swap/autoswitch.py
+++ b/src/claude_swap/autoswitch.py
@@ -425,20 +425,15 @@ def _binding_recovery_ts(
     would otherwise rank a snapshot nobody has refreshed above a measured,
     genuinely imminent one.
     """
-    windows = [
-        (pct, _parse_reset_ts(resets_at))
-        for _label, pct, resets_at in oauth.relevant_windows(
-            usage if isinstance(usage, dict) else None, models
-        )
+    usable = [
+        (pct, ts)
+        for _label, pct, resets_at in oauth.relevant_windows(usage, models)
+        if (ts := _parse_reset_ts(resets_at)) is not None and ts > now
     ]
-    usable = [(pct, ts) for pct, ts in windows if ts is not None and ts > now]
-    if not usable:
-        return float("inf")
-    return max(usable)[1]  # highest pct wins; its reset is the answer
+    return max(usable)[1] if usable else float("inf")  # highest pct wins
 
 
 def _every_account_above_threshold(
-    current: str,
     candidates: Sequence[str],
     headroom: dict[str, float | None],
     active_headroom: float | None,
@@ -1147,15 +1142,15 @@ class AutoSwitchEngine:
         #
         # Deliberately narrow. It engages only when every measured OAuth
         # account is at/over the threshold, so a single healthy peer still
-        # wins the normal way, and the hysteresis margin below still applies
-        # so two accounts in the 90s cannot ping-pong.
+        # wins the normal way, and RECOVERY_HYSTERESIS_S below replaces the
+        # percentage-point margin so two accounts in the 90s cannot ping-pong.
         all_above = _every_account_above_threshold(
-            current, oauth_candidates, headroom, active_headroom, settings.threshold
+            oauth_candidates, headroom, active_headroom, settings.threshold
         )
         active_recovery_ts = (
             _binding_recovery_ts(usage.get(current), self._models, now)
             if all_above
-            else float("inf")
+            else 0.0  # unread unless all_above; never a live sentinel
         )
         qualifying: list[tuple[tuple, str]] = []
         any_known = False
@@ -1168,6 +1163,11 @@ class AutoSwitchEngine:
                 continue  # itself at its limit — never a target
             reset_ts = (
                 _seven_day_reset_ts(usage.get(num), now) if consume_first else None
+            )
+            recovery_ts = (
+                _binding_recovery_ts(usage.get(num), self._models, now)
+                if all_above
+                else 0.0
             )
             if trigger in ("proactive", "consume-first"):
                 # Landing must be healthy: an account at/over the threshold
@@ -1196,10 +1196,7 @@ class AutoSwitchEngine:
                     # prevent is still prevented: the target must come back
                     # meaningfully sooner than where we are, so the reverse
                     # move never qualifies.
-                    if (
-                        _binding_recovery_ts(usage.get(num), self._models, now)
-                        >= active_recovery_ts - RECOVERY_HYSTERESIS_S
-                    ):
+                    if recovery_ts >= active_recovery_ts - RECOVERY_HYSTERESIS_S:
                         continue
                 elif active_headroom is not None:
                     # best: the candidate must beat the active account by the
@@ -1212,10 +1209,7 @@ class AutoSwitchEngine:
                 # this account back, not the weekly one. With everything in the
                 # 90s the only thing worth optimizing is how long until work can
                 # continue; headroom breaks ties, then sequence order.
-                key: tuple = (
-                    _binding_recovery_ts(usage.get(num), self._models, now),
-                    -h,
-                )
+                key: tuple = (recovery_ts, -h)
             elif consume_first:
                 # Soonest weekly reset first (unknown resets sort last), most
                 # headroom breaks ties, then sequence order.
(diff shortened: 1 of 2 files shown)


codeslake:
Reviewed this against the open issues and PRs that touch the same code, and found a real bug in my own first version. Pushed two more commits.

**The bug (2792811):** `if consume_first:` caught before the new `all_above` branch, so a consume-first user reached the new ranking (soonest binding recovery) without ever passing the recovery-hysteresis gate — filtering on weekly reset ordering while sorting on 5-hour recovery, two different axes, leaving that strategy with no anti-flap guard at all. Two accounts whose windows rolled over a minute apart traded places. `all_above` is now checked before either strategy and owns both the gate and the key: with nothing below the threshold the strategy question is moot, since consume-first exists to spend perishable *weekly* quota and every account in that state is blocked on a window returning in minutes. The test fails on the previous commit.

**Related issues, checked:**

- **#149** (closed, consume-first) — the escape ranks by *binding* recovery, not the weekly window, so it does not re-litigate that strategy; it only applies where consume-first has nothing to choose between.
- **#115** (closed, "auto did not switch despite a valid candidate") — same symptom class, different cause: there the candidate was genuinely healthy on the binding window and the fix was window selection. Here every candidate really is above the line. The at-limit path is untouched, and 5cc8a48 pins that.
- **#194** (open, extra-usage headroom) — orthogonal and complementary. It argues an account with spend credit should not count as exhausted; this changes which account to pick among those already known to be spent. If #194 lands, `headroom` gains those accounts and they flow through the ordinary path before this escape ever engages.
- **#146** (open, 5h readings flap) — the escape sorts on `resets_at`, not `pct`, and window boundaries do not flap the way utilization does. A reset already in the past sorts last rather than first, so a stale row cannot win.
- **#187 / #161** — credential-lineage and lockstep-detection; no overlap.

**#178** (open, burn-rate-aware consume-first) touches the same gate chain. It currently sits 51 commits behind `main`, so most of the conflict is stale base, but a real merge decision is needed between the two — happy to rebase on top of it if you would rather land that one first.

realiti4:
Hi, thanks a lot for the PR, both bugs are real and the direction is exactly right. Found three things during review; could you check these on your end too?

1. **`_respect_poll_plan` can lengthen short intervals** — the default is 60s (not 360s) and the range starts at 15s, so `max(min(delay, due_in), URGENT_INTERVAL_S)` turns a 15s interval into 60s sleeps and clamps the lower jitter half at the default. `min(delay, max(due_in, URGENT_INTERVAL_S))` keeps the "only shortens" invariant; a test at 15s would pin it.
2. **`_binding_recovery_ts` can rank on a non-binding window** — it drops unknown/past resets *before* picking the highest pct, contradicting its docstring. Picking the max-pct window first, then its reset or `inf`, fixes it.
3. **The recovery key leaks into at-limit** — the gate is scoped to proactive/consume-first but `key = (recovery_ts, -h)` isn't, so at-limit now ranks by soonest-recovery instead of most-headroom. Maybe better, but it's untested and described as untouched — either scope it or pin it with a test.

I agree that we should land this before #178, especially since I was slow to merge that and it is already 51 commits behind and needs a rebase regardless.

codeslake:
All three confirmed on my end — reproduced each before touching it, and each fix has a test that fails on the pre-review code. Pushed as `322d5cb`.

**1. `_respect_poll_plan` lengthening — confirmed, and worse than the report.**

You are right that the default is 60s and the floor is 15s. I developed against 360s, which is my own machine's setting, and then wrote "only ever shortens" in the commit message as though it were an invariant I had checked. Measured:

```
interval=15s, jittered delay=13.5s, row due in 5s
  max(min(13.5, 5.0), 60.0)  -> 60.0s     lengthened 4.4x
  min(13.5, max(5.0, 60.0))  -> 13.5s
```

Taking your form. It also holds at the 60s default, where the old one flattened the entire lower jitter half, and it still refuses to poll below the planner's floor when the row is overdue.

**2. `_binding_recovery_ts` ranking on a non-binding window — confirmed.**

The filter ran before the `max`, so whenever the binding window's reset was unknown or past, a lower window answered for the account:

```
7d 95% (no resets_at), 5h 40% (resets in 1h)
  before -> "back in 60min"      <- the 5h window, which is not what binds
  after  -> inf
```

The docstring directly above it says the binding window's reset is the answer, so the code contradicted its own contract. Now the max comes first and its reset — or `inf` — is the result.

**3. Recovery key leaking into at-limit — confirmed, and my test was the problem.**

The gate is inside `if trigger in ("proactive", "consume-first")`; the key assignment is one level out, at the same indent as `qualifying.append`. So at-limit and failover were re-ranked by soonest-recovery. Scoped it to the same triggers.

The part worth flagging: I *did* add `test_at_limit_trigger_still_ignores_the_landing_rule` for exactly this concern, and it passed throughout — because its healthy candidate (30% used) made `all_above` False, so the tick never reached the branch. A test that cannot reach the code it names is worse than no test, because it reads as coverage. The new one keeps every account above the line, which is the combination that gets there.

**Verification:** each test fails on the pre-review code with the symptom named — `a 15s interval slept 60.0s`, `assert 1003600.0 == inf`, `at-limit must take the most headroom, not the soonest recovery`. Full suite 1662 passed / 0 failed.

Agreed on landing before #178 — and if it helps, its conflict with this branch is mostly stale base rather than real overlap, since it predates the current `main` by 51 commits.

realiti4:
Thanks, looks good merged it!

Issue #198, opened 2026-07-29
feat(pin): keep Remote Control / artifacts on one account while inference follows the swap


codeslake:
## The problem

cswap swaps the on-disk credential, so *everything* follows the swap — including two things that are not inference and that you usually want to stay put:

- **Remote Control** — a session's owner is fixed at creation by whichever bearer created it. Swap accounts and the phone/web loses the session; stale "ghost" sessions pile up on the old account.
- **Artifacts** — owned by the publishing bearer. After a swap, a republish 403s and the artifact "disappears" from the account you are logged into on the web.

So the thing I actually want is: **inference keeps following `cswap switch` / `cswap auto`, while the claude.ai-side assets stay on one account of my choosing.**

## Why this needs more than an env var

Claude Code resolves all of these through one credential accessor, and there is no per-operation token selector. Dead ends I measured on 2.1.220:

| Approach | Result |
|---|---|
| `CLAUDE_CODE_OAUTH_TOKEN` | applies to inference too — pins the wrong half |
| `x-organization-uuid` header | server derives ownership from the bearer (bearer #2 + org-header #1 → session owned by #2) |
| `CLAUDE_CONFIG_DIR` | isolates the whole session, not per-operation |
| `CLAUDE_BRIDGE_OAUTH_TOKEN` | declared in the binary, zero consumers |
| `ANTHROPIC_BASE_URL` | `/v1/code/sessions` is gated on "first-party Anthropic API provider" |

Splitting auth *per operation inside one session* therefore means intercepting the requests.

## What I have working

Branch: [`codeslake/claude-swap@feat/account-pin-proxy`](https://github.com/codeslake/claude-swap/tree/feat/account-pin-proxy) — implemented with tests, not opened as a PR here yet (hence this issue).

A small local MITM forward proxy inside cswap that swaps the `Authorization` bearer on exactly the routes whose server-side ownership is decided by it, and passes everything else — `/v1/messages` above all — through untouched.

```
claude session
  HTTPS_PROXY ─► cswap pin proxy ──► (whatever HTTPS_PROXY was already set) ──► api.anthropic.com
                   swaps bearer on: /v1/code/sessions*, /v1/sessions/*,
                                    /api/frame/*, /v1/ultrareview/*
                   passes through:  /v1/messages, /api/oauth/usage, everything else
```

```
cswap pin 2          # RC / artifacts / ultrareview → account 2
cswap pin            # show the current pin
cswap pin --clear    # remove it
```

[...]

PR #203, opened 2026-07-30
feat(pin): pin Remote Control and Artifacts to one account

codeslake:
Pin Remote Control and Artifacts to one account while inference keeps following
the active one. A local MITM on `api.anthropic.com` swaps the `Authorization`
bearer on the RC/Artifact routes only; everything else is blind-tunnelled.

## Shape

- `cswap pin <acct>` starts a per-user daemon and records it. Sessions launched
  by `cswap run`, by the TUI, or by hand all pick it up — `.claude.json`'s `env`
  block is what reaches a hand-typed `claude`, and `cswap pin-env` is the hook
  for a launcher that wants it in the shell.
- Pinned routes: `/v1/code/sessions`, `/v1/sessions/`, `/api/frame/`,
  `/v1/ultrareview/`. The `/worker` subtree is deliberately excluded — it
  carries a session JWT, and swapping it produces a 403 storm.
- **Fail-open on purpose.** A pin that cannot resolve must never block work.
  The cost is that the RC session is then born owned by the active account
  permanently (ownership is fixed at `/bridge`; there is no transfer API), so
  the fail-open path now warns on stderr and reports `can_pin` on `/health`.

## What this branch fixes on top of the feature

- `e17ae62` the fail-open was silent. A daemon that could not read its
  credential store served every pinned route unswapped, and 19 sessions had to
  be rebuilt by hand. Now it says so, once, and `/health` carries `can_pin`.
- `6e7e678` a re-pin shortened the recorded egress chain, so a session behind a
  filtering proxy lost its route.
- `a8966bd` a launcher's `ALL_PROXY` kept naming the hop we chain through, so a
  pinned session ran with two proxy variables pointing at different proxies.
  `wire_global_config` now claims it too; `wire_env` only rewrites one that is
  already there, never invents one (its output can be eval'd into the user's
  shell, and an `ALL_PROXY` we created would route that shell's `git`/`uv`/`gh`
  through a MITM built for one client).
- `7a9f9da` reverts `1a60209` — the premise was disproved by re-measurement.

## Verification

- 133 tests in `tests/test_pin_proxy.py` + `tests/test_pin_proxy_server.py`,
  including a real TLS handshake against the generated trust file and a control
  that fails without it.
- Live, on a host with a corporate proxy chain: sending the **active** account's
  bearer through the pin returns the **pinned** account's session list and not
  the active one's. All 13 locally-bridged sessions land on the pinned account,
  `connection_status=connected`, zero on the active account.
- `curl` proxy-variable precedence measured rather than assumed:
  `https_proxy=A` + `ALL_PROXY=B` dials A; B only when A is unset.

## Known limits

- Ownership cannot be moved after `/bridge`, so sessions bridged before a pin
  existed stay where they are. Re-registering them is per-session and manual.
- Bundle narrowing is unguarded on purpose (`cc53904`): the ambient CA count is
  2 on one host here and 132 on another, so any floor breaks a healthy host.


codeslake:
Closing while the packaging question is settled — the pin will land as a separate `cswap-pin` package per #198, not as a core PR. Reopening or superseding once that shape is agreed.

PR #204, opened 2026-07-30
fix(autoswitch): don't trade real headroom for a reset days away

codeslake:
Follow-up to #202, from running it on a live 3-account setup.

## What #202 got right, and where it has no bound

#202 added: when *every* account is above the threshold, rank by which one's binding window resets soonest. The evidence was minutes-scale — the measured case had a peer whose 5-hour window reset in **8 minutes**, and trading 9 points of headroom for an 8-minute wait is plainly right.

The rule shipped with no upper bound on that wait, so it applies identically when every reset is days out.

## Measured

Live, three accounts, threshold 90, `strategy: best`:

```
 1 (active)  7d 91%   resets in 109h   <- 9 points of headroom
 2           7d 94%   resets in  80h
 3           7d 98%   resets in  50h   <- 2 points, but "soonest"
```

The engine ranked by recovery and moved to account 3 — from 9 points of headroom to 2. Nothing returns within the session either way, so those 9 points were the only resource that could still do work. The user switched back to account 1 by hand; the engine moved them off again on the next tick.

| | headroom | recovery |
|---|---|---|
| acct 1 | **9%** — most | 109h — worst |
| acct 3 | 2% — least | **50h** — best |

## The change

Past `RECOVERY_HORIZON_S` (4h — one whole 5-hour cycle, so same-day resets keep the recovery ranking), rank by headroom instead.

The escape itself stays. Dropping out of `all_above` entirely would re-arm the landing-health gate, which has no answer when everything is above the line, and park the engine rather than fix it — so only the ranking axis changes.

## The anti-flap bar

Ranking by a *reset* rather than by headroom needs a matching no-return rule, because burn changes headroom continuously while a reset does not: without one the engine returns to the account it just left as soon as the active burns past it. `_left_account_recovered` holds the departed account barred until it genuinely recovers.

A failover departure records `(leftHeadroom, leftRecoveryAt) = (None, None)` — a real departure whose severity was unmeasured — so its release cannot use headroom at all and falls to two legs: the landing floor, and "the peer's binding reset is meaningfully sooner than the active's".

That second leg is where the analysis matters, because **`_binding_recovery_ts` returns `inf` for five distinct states**:

```
A  finite reset, far out      D  unreadable (None)
B  no resets_at               E  no windows ({})
C  reset in the past
```

One predicate over five states is wrong in both directions, and both readings ship a real defect. Reading every `inf` as "never" releases onto a peer arbitrarily far out on no evidence. Reading every `inf` as "unknown" and holding pins the engine on a near-spent active for up to a full window while the peer is back in minutes.

The release condition is therefore `math.isfinite(active_recovery_ts) or peer_recovery_ts - now <= RECOVERY_HORIZON_S`: a known-finite active decides on the comparison, and an unknown active still yields to a peer that is provably close. Measured, with both controls invariant and only the damage rows moving:

```
POS  active 400h / peer 50min   SWITCHED
NEG  peer only 60s sooner       BLOCKED
DMG-a active NO resets_at       SWITCHED
DMG-b active reset in PAST      SWITCHED
DMG-long 40 ticks               switched=1
```

This releases on all five `inf` states when the peer is inside the horizon, not only the two the motivation names. Whether that violates F6 (an unreadable active must not flip a verdict) was **adjudicated as unreachable**: `trigger` is derived from `active_headroom`, so an unreadable active always classifies as `failover`, which `_no_return_account` scopes out of the bar entirely. The one route that survives classification is closed by exact complementarity — with `active_headroom=None`, `_every_account_above_threshold` returns `False` unconditionally, making the landing gate and the failover landing leg exact complements. Swept every headroom in [0,100] at 0.01 across five thresholds: **0 values where the horizon term could decide**, and 0 over 1920 random ticks.

## A `consume-first` departure could write the failover snapshot shape

`_left_account_recovered` inferred the trigger from `(None, None)`. But a consume-first phase-2 refetch writes that same shape whenever the refetched active row has a `pct` but is otherwise unmeasurable in the same tick its weekly reset is known — `account_headroom` needs a numeric `pct`, `_seven_day_reset_ts` needs only `resets_at`, and one row can satisfy one and not the other. The reader then ran the more permissive failover legs on an ordinary departure. Reproduced end-to-end; reclassification differs on ~31% of swept fleets, both directions. `_perform` now records `state["leftTrigger"]` directly; records written before the field existed keep the old two-null inference.

## Verification

- Suite 1793 passed, 3 skipped; 7.1s.
- 21 mutants in place with controls (`POS-rename` killed 140, `NOOP` survived, `SYNTAX` errored). Every horizon and `leftTrigger` mutant killed except one float-boundary equivalent (needs a peer at exactly 14400.000000s).
- `leftTrigger` blast radius: sole reader/writer is `autoswitch.py`; a downgrade ignores the key.
- Real account store unchanged across every run.

## Base

Rebased onto the current `main` (`3d9f596`). The xdist parallelisation, the
real-Keychain serialisation hook and the `uv sync --locked` CI gate this branch
used to carry landed upstream via #209, so those commits dropped out of the
range — `pyproject.toml`, `tests/conftest.py`, `.github/workflows/ci.yml` and
`uv.lock` are byte-identical to `main` here, and the diff is only
`autoswitch.py` and its tests.

### DIFF
--- a/src/claude_swap/autoswitch.py
+++ b/src/claude_swap/autoswitch.py
@@ -109,6 +109,17 @@ RECOVERY_HORIZON_S = 4 * 3600.0
 # to a quarter of the headroom it just beat.
 HORIZON_HEADROOM_RATIO = 2.0
 
+# Below this much headroom an account is spent for practical purposes, and
+# comparing headroom between two spent accounts compares noise: at the burn
+# rates measured on 2026-07-30 a one-point edge is under ten minutes of work,
+# less than two poll intervals. When EVERY candidate is down here the ranking
+# has to fall back to the reset — sit where the quota returns first, so the
+# reset finds us already on it — no matter how far out that reset is. Without
+# this, three accounts at 99% left the engine with no qualifying candidate and
+# it parked on whichever it happened to hold, including the one resetting
+# 59 hours after a peer.
+SPENT_HEADROOM_PCT = 3.0
+
 # Adaptive scheduling: the baseline request volume is O(1) per tick — the
 # active account plus ONE due candidate (stalest data first) — instead of
 # every account in parallel, and the per-account cadence itself (movement,
@@ -1183,8 +1194,18 @@ class AutoSwitchEngine:
         # the landing-health gate has no other answer here). An UNKNOWN
         # recovery is no evidence rather than a distant one, so it keeps the
         # recovery axis and that gate keeps deciding. See RECOVERY_HORIZON_S.
+        # Everyone spent: headroom can no longer tell the candidates apart, so
+        # the reset is the only thing left worth ranking on (see
+        # SPENT_HEADROOM_PCT). Checked before the horizon, which is about
+        # whether a sooner reset is worth REAL headroom — a question that only
+        # arises while some real headroom exists.
+        all_spent = all_above and all(
+            (headroom.get(n) is not None and headroom[n] <= SPENT_HEADROOM_PCT)
+            for n in oauth_candidates
+        ) and (active_headroom is not None and active_headroom <= SPENT_HEADROOM_PCT)
         recovery_useful = all_above and (
-            active_recovery_ts == float("inf")
+            all_spent
+            or active_recovery_ts == float("inf")
             or active_recovery_ts - now <= RECOVERY_HORIZON_S
         )
         qualifying: list[tuple[tuple, str]] = []
--- a/src/claude_swap/autoswitch.py
+++ b/src/claude_swap/autoswitch.py
@@ -106,6 +106,72 @@ HORIZON_HEADROOM_RATIO = 2.0
 # whichever account we happen to hold.
 SPENT_HEADROOM_PCT = 3.0
 
+def _recovery_is_useful(
+    candidate_recovery_ts: float,
+    active_recovery_ts: float,
+    active_headroom: float,
+    best_candidate_headroom: float,
+    threshold: float,
+    now: float,
+) -> bool:
+    """Rank THIS candidate by soonest reset, or by headroom?
+
+    One place, because four independent gates over two axes produce sixteen
+    combinations and a review found reachable holes in four of them. Each
+    clause below is a hole that was measured, not a hypothesis.
+
+    Reset wins when:
+
+    * **Everything is spent.** Below ``SPENT_HEADROOM_PCT`` a headroom edge is
+      under ten minutes of work — less than two poll intervals — so comparing
+      two spent accounts compares noise, and the only real question is which
+      returns first. Asked of the active and the BEST candidate — "is there
+      anything worth having?" — not of every account. Requiring EVERY account
+      to be spent made the answer non-monotone in headroom (2/2/3.0 took the
+      soonest reset; improving one peer to 3.5 flipped the axis and parked on
+      the account resetting 59h later) and let one unknown headroom — a
+      sentinel row, a locked keychain — veto the check for everybody, which
+      parked three accounts at 99% on the one resetting LAST. The maximum
+      answers both: an unknown is not a maximum, and a peer improving can only
+      move the answer toward "keep the headroom".
+    * **This candidate is back soon.** Asked per candidate, not once on the
+      active: an active bound by its WEEKLY window sits days out while a
+      peer's five-hour window returns in minutes, and asking the active
+      refused that peer — the #202 case this is supposed to preserve.
+
+    Past the horizon we rank by headroom and, when no candidate can meet the
+    ratio, nobody qualifies — deliberately. An active holding materially more
+    quota than any peer can offer SHOULD keep the work; that is the whole
+    finding this horizon encodes. The one thing that must not happen is the
+    engine parking while a peer holds meaningfully more, and the ratio cannot
+    produce that: it is unreachable only when the active already holds more
+    than half of what the threshold permits, i.e. more than any candidate.
+
+    KNOWN RESIDUE, deliberately left: inside the spent band the axis is the
+    reset, above it the ratio, and the step between them is not monotone —
+    2 / 2 / 3.0 switches to the soonest reset while 2 / 2 / 3.5 parks, because
+    3.5 leaves the spent band without meeting 2x. Every fix tried was worse:
+    ranking that band by reset re-armed the days-away trade for 9 / 10 / 10
+    and broke the return-leg flap guard. The band is 0.5 points wide at the
+    default constants and both outcomes keep a working account, so it is
+    recorded rather than papered over.
+
+    An UNKNOWN active recovery (``inf``, which also means "already elapsed")
+    deliberately does NOT keep the recovery axis. It used to, and that re-armed
+    the exact trade this horizon forbids: 9 points of headroom for a peer
+    resetting 50h out, whenever the active's ``resets_at`` was missing or
+    stale. No evidence a sooner reset helps is a reason to keep the headroom,
+    not to spend it.
+    """
+    if active_headroom <= SPENT_HEADROOM_PCT and best_candidate_headroom <= (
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for the PR, it looks good overall!

Could you take a look at these findings as well before merging?

**1. The no-return bar goes stale after a manual switch.**

* engine switches 1 → 2 (bars account 1, `lastSwitchTo=2`)
* user manually switches 2 → 3
* active now has 2 pts, barred account 1 has 4 pts, all are above threshold, and resets are far out
* result: BLOCKED. Clearing `lastSwitchFrom` allows it to switch correctly to account 1.

The release legs do not cover this case: 4 pts does not clear the 2×+3 dominance ratio, and the account has not improved over its baseline. As a result, it remains on the worse account until the at-limit escape.

Manual switching is a normal scenario here—the original reporter did exactly this. Arming the bar only when `lastSwitchTo == current` should fix it. One detail: `lastSwitchTo` is stored as a string and `lastSwitchFrom` as an integer, so the comparison will need to be normalized.

This is the only item I consider blocking before merge.

**2. Inside the horizon, headroom is still ignored entirely.**

Active 9 pts / 109h, peer A 0.1 pts resetting in 3h59m, peer B 9 pts / 50h → it switches onto the 0.1-pt account. At 4h01m, it stays put.

To be fair, `main` makes the same trade. This follows #202’s existing semantics rather than being introduced here, and the at-limit logic self-corrects one tick later. So this is not a blocker, just a possible follow-up: add a minimum-runway floor to tier 0.

**3. Nit: “4h covers a whole 5-hour cycle” is not quite accurate.**

A 5h reset can still be up to 5h away, leaving an approximately 1h band where a 5h-bound peer is headroom-ranked. The effect is conservative, so it is not a bug; the comment could either be reworded or the constant bumped.

Also, take or leave: the docstrings contain quite a bit of review-round history (`R7`, `round-6`, etc.). That was useful during review, but the behavior is already pinned down by tests, so a trim pass may make them easier for the next reader to follow.

codeslake:
Rebased onto `main` (`06a7aa5`), after #197 and #209 landed. Force-pushed
`c1f5513` → `23b7788`.

The rebase changes no code. The branch's own diff is identical before and
after:

```
before (c1f5513 vs 9f35426)   src/claude_swap/autoswitch.py  +670 -28
after  (23b7788 vs 06a7aa5)   src/claude_swap/autoswitch.py  +670 -28
```

What dropped out are three commits whose content `main` now carries via #209 —
the xdist parallelisation, the real-Keychain serialisation hook, and the
`uv sync --locked` CI gate. git dropped the third itself ("patch contents
already upstream"); the other two were skipped after confirming `main` has
them. So these four files are byte-identical to `main` on this branch now:

```
pyproject.toml   tests/conftest.py   .github/workflows/ci.yml   uv.lock
```

`uv lock --check` is rc=0 (the rc of `uv` itself, not of a pipe).

Suite on `23b7788`: **1819 passed, 3 skipped, 6.20s.**

codeslake:
All three reproduced here before any code changed, and all three answered.
Pushed as `31d3a6e` (append, no force-push).

## 1. The stale bar — fixed

Reproduced exactly as you described, and the mechanism is one step worse than
"the release legs do not cover it": on this path they are never **reached**.
`_no_return_account` returns at

```python
barred = str(came_from)
if not recovered:
    return barred
```

so the `2×` ratio leg below it does not run. Measured on your fleet (active
2 pts, barred account 1 has 4 pts, all above threshold, resets far):

```
row                               barred   ratio leg   recovered   can reach 1?
CONTROL-A engine on its landing   1        False       False       False
PROBE     user moved to 3         1        True        False       False
CONTROL-B lastSwitchFrom cleared  None     None        True        True
```

PROBE shows `ratio leg=True` while still barred — that is the early return.

Your fix, taken as suggested: the bar is armed only while `lastSwitchTo ==
current`. `current` was already at the call site (`_left_account_recovered`
takes it two lines above), so it is threaded rather than re-derived.

```
row                               barred BEFORE   barred AFTER
CONTROL-A engine on its landing   1               1
PROBE     user moved to 3         1               None
CONTROL-B lastSwitchFrom cleared  None            None
```

CONTROL-A is unmoved: the bar still holds where the engine actually sits.

**The type asymmetry you flagged is real and is pinned.** Both sides are
`str()`-normalised; without it the int-on-landing row releases a bar that must
hold:

```
lastSwitchTo   current          barred BEFORE   barred AFTER
2   (int)      "2" on landing   1               1
2   (int)      "3" moved away   1               None
"2" (str)      "2" on landing   1               1
"2" (str)      "3" moved away   1               None
```

**A record written before `lastSwitchTo` existed keeps the bar** — absence
cannot prove the engine moved away:

```
lastSwitchTo   current   barred
(absent)       3         1
```

Mutation, `tests/test_autoswitch.py`, baseline 253 passed: gate disarmed →
2 failed; `str()` dropped → 1 failed; absent-key disarms → 1 failed;
comparison inverted → 13 failed; `current` not threaded at the call site →
2 failed. NOOP controls 253 passed.

That last one initially SURVIVED, because the test drove
`_no_return_account` directly and left the call-site wiring unpinned. The test
now drives `tick()` end to end and asserts the live login, which a hand switch
moves and `activeAccountNumber` does not.

Suite 1819 → **1824 passed, 3 skipped**, 6.1s → 6.5s.

## 2. Headroom inside the horizon — confirmed, not changed here

Reproduced on this branch. Active 9 pts / 109h, peer A 0.1 pts, peer B 9 pts /
50h:

```
peer A resets in    outcome    live acct              axis
3h59m (inside 4h)   SWITCHED   2 (the 0.1-pt peer)    reset-ranked
4h01m (outside)     BLOCKED    1 (stays put)          headroom-ranked
```

Your read of the self-correction is right, and it is one **tick**, not one
cooldown: at-limit skips the cooldown gate (`_perform` gates it on
`trigger in ("proactive", "consume-first")`), so tick 2 at +60s with peer A at
100% switches back.

Left alone here. A minimum-runway floor on tier 0 — rank by reset only among
candidates above `SPENT_HEADROOM_PCT`, a constant the module already has — is
a behaviour change on the axis #202 settled, so it wants its own measurement
and its own PR. Happy to open an issue for it.

## 3. The 4h comment — reworded, constant unchanged

You are right that it is inaccurate. Measured, `_recovery_is_useful` alone
with neither side spent:

```
peer resets in    useful?
3.9h              True
4.0h              True
4.1h              False
5.0h              False
```

Through `tick()`, a 5h-bound peer switches at 3.5h and blocks at 4.5h and
4.99h — so it spends roughly 1h of every cycle headroom-ranked. Conservative,
as you say. The comment now says 4h keeps *most* of a 5-hour cycle and names
the 4h–5h fallback. `RECOVERY_HORIZON_S` is still `4 * 3600.0` — bumping it is
a behaviour change and did not belong in this reply.

## Docstring round history — removed

```
grep -nE "round[ -][0-9]|R[0-9]+ review|\(I-[AB]\)|C-1" \
  src/claude_swap/autoswitch.py tests/test_autoswitch.py
```

is now empty, and so are the softer forms (`R1`–`R8`, `F1`–`F6`, `C1`,
`I-1`…`I4`). Every measurement was kept and rewritten as what the code does:
`TestAcceptance204eR1toR7` became `TestTheReleasePredicateOneStateShapePerTest`,
and 11 method names lost their matrix labels. Thanks for pointing at it — it
reads much better without the bookkeeping.

The branch is also rebased onto `main` (`06a7aa5`), so `pyproject.toml`,
`tests/conftest.py`, `.github/workflows/ci.yml` and `uv.lock` are
byte-identical to `main` here and `uv lock --check` is rc=0.

realiti4:
Thanks for the adjustments, and PR again. I've merged it.

PR #209, opened 2026-08-01
test: a developer's terminal must not decide whether the suite passes

codeslake:
The suite passed or failed depending on whose terminal ran it. `FORCE_COLOR=1`
or `COLORTERM=truecolor` exported in a developer's shell turned 11 tests red;
CI never saw it because CI exports neither.

## Cause

Two independent leaks, and the second is why the first fix was not enough.

**1. The environment reached the tests.** `printer.colors_enabled()` and the
theme detection read `FORCE_COLOR` / `COLORTERM` / `NO_COLOR` and cache the
result. Nothing scrubbed them, so assertions on plain output compared against
ANSI-wrapped strings.

**2. `monkeypatch.undo()` unwound the fixture that scrubbed them.** pytest
hands **one** `MonkeyPatch` instance to an autouse fixture and to the test's
own `monkeypatch` parameter — verified by identical `id()`. A test calling
`.undo()` mid-body therefore restores `FORCE_COLOR='1'` for the rest of that
test. Blast radius is wider than colour: the same call also unwinds
`block_real_keychain`, restoring the real `macos_keychain.get_password`.

## Change

- `tests/conftest.py`: `_deterministic_colour` autouse fixture scrubs the three
  variables and resets both caches, snapshotting before the reset.
- 8 of 8 `monkeypatch.undo()` call sites replaced with 14
  `pytest.MonkeyPatch.context()` blocks. The 5 remaining grep hits are prose.
- `tests/test_colour_cache_isolation.py`: pins that the globals are unlatched,
  with a real pty for the TERM path (POSIX-gated).
- `tests/test_printer.py`: `TestEntryAssertionsCatchAPoisonedGlobal` invokes the
  fixture body directly via `_deterministic_colour.__wrapped__(mp)` against a
  poisoned global, so it has no ordering dependency on other tests.

Tests only. No production file is touched.

## Verification

Both runs green, and the hostile one is the point:

```
env -u FORCE_COLOR -u COLORTERM  ->  1705 passed, 3 skipped
FORCE_COLOR=1 COLORTERM=truecolor ->  1705 passed, 3 skipped
```

Also green on 4 random seeds, `-n 4`, and pytest 8.0.2 (the declared floor) as
well as 9.0.2.

14 mutations, all killed by named tests: each of the 6 fixture scrub lines
individually, both wrong-direction pins for both globals, both entry
assertions, the snapshot-before-reset ordering, and the `.undo()` pattern
reverted. The two `delenv` lines die **on a clean environment** (2 and 1
failures with nothing exported) — the specific hole this branch closes.

## Notes

`TERM=dumb` vs `linux` is caught even though `appearance.py:97` and
`printer.py:97` disagree on it, and the `NO_COLOR` scrub has its own test
separating it from `FORCE_COLOR` despite both producing identical plain output.

`appearance._reset_cache()` becomes dead code: this branch deleted its only
caller and replaced it with a direct `monkeypatch.setattr`. Left in place
rather than removed as unrelated cleanup.


realiti4:
Thanks a lot for the PR! Merged it.

PR #210, opened 2026-08-01
feat(pin): keep Remote Control and Artifacts on one account, via an optional extra

codeslake:
Implements the claude-swap side of #198: the pin as an optional extra, with the MITM proxy in a separate package as you suggested.

Proxy lives at [codeslake/cswap-pin](https://github.com/codeslake/cswap-pin), on PyPI as [`cswap-pin`](https://pypi.org/project/cswap-pin/). Nothing in this PR intercepts anything, and `/v1/messages` is never touched — usage still bills whichever account the swap landed on.

## The seam

| | |
| :-- | :-- |
| `claude_swap/pin.py` | import-safe without the extra, modelled on `menubar.py` |
| `cli.py` | `cswap pin [NUM\|EMAIL] [--clear\|--heal\|--ensure\|--get_port\|--set_port]`, pre-dispatched like `run`/`auto` |
| `session.py` | launch-env wiring, one call, no-op without the extra |
| `pyproject.toml` | `pin = ["cswap-pin>=0.1.103"]` |
| `tui/{dashboard,widgets,autoview}.py` | `Cloud` menu + `○ cloud` badge, gated on the same `find_spec` check |

Without the extra installed there is no `Cloud` entry and no badge. The menu rebuilds on the snapshot watch, so installing or clearing takes effect without restarting the TUI.

## Outside the seam — these want their own read

- **`claude_locks.py`** — `proper_lockfile`'s jittered retry sleep now clamps to the REMAINING budget, through one `_nap` both sleeps go through. The deadline is checked at the top of the loop, so a sleep longer than what is left runs to completion first and the raise lands late. Measured on the contended path before: `timeout=0.01` → 0.302s, `timeout=0.1` → 0.258s. Clamping to the remainder rather than to `timeout` is what makes it hold on the second retry too — `min(want, timeout)` is a no-op once most of the budget is spent.

  This line described a change that was NOT in the branch until now: the file was untouched and the sleep unconditional. It is true of the code as of this revision.
- **`switcher.py`** — `purge` unwires `.claude.json` *before* deleting `backup_dir`. The old order deleted the pin record and cert dir while leaving the env block Claude Code applies at boot, stranding every hand-launched `claude` on a dead port. A failed unwire now prints the config path and key to remove by hand instead of "Purge complete."
- **`oauth.py`** — the pin's CA is now **added** to the default trust store rather than replacing it. The earlier revision set `SSL_CERT_FILE`-style replacement, which narrows what the process will trust for *every* host, not just ours. `_pin_aware_ssl_context()` builds `ssl.create_default_context()` and calls `load_verify_locations()` on the pin CA; without the extra installed it returns the default context untouched, and a test drives both arms with the CA path injected so the no-pin control is a real assertion rather than a skip.

## Three decisions

- **`--clear` works without the extra, deliberately.** Uninstalling the pin must not leave the wiring behind, so `clear_wiring()` lives here: it reads a marker cswap wrote (`_cswapPinWiredKeys`), restores what it displaced, and touches no proxy, daemon or credential.
- **"not installed" vs "installed but broken" split with `find_spec`, not `except ImportError`.** Catching cannot tell them apart, and conflating them tells a user to install a package they already have when the real cause is e.g. a missing `cryptography`.
- **The floor tracks the current release and is raised with each one.** It is not enforced at runtime — `menubar.py` declares `rumps>=0.4.0` the same way and then only asks whether the import works. An earlier revision of this branch argued for *no* floor; that was wrong in the other direction, and `cswap-pin>=0.1.3` then sat unchanged through 36 releases while `uv tool install 'claude-swap[pin]'` resolved a build from before the port handdown.
- **`clear_pin()` / `set_pin()` return `(ok, message)` and both front ends render it**, so CLI and TUI cannot diverge on what "pinned" means. Enforced structurally: `run()`'s AST must call the pair and must not call `apply_pin`.

## Tests

Two guard the seam rather than the behaviour:

- an **AST check** that nothing imports `cswap_pin` at module scope — behaviour alone passes right up until someone adds one, and then the extra is not optional
- a **subprocess** importing every module with `cswap_pin` blocked at `sys.meta_path`. Blocking `builtins.__import__` would miss `importlib.import_module`.

Each guard was mutation-checked when it landed: reverting it turns its own test red, run and recorded at that point.

CI on this head: **2316 passed, 4 skipped** on Linux (3.12), plus the four Windows shards and the macOS keychain job, all green.

## Ordering

Nothing has to land before this. The branch is cut from current `main` (0 behind), and CI is green on it standalone.

It does overlap #199 though, so whichever of the two lands second needs a resolution: measured on a merge of both, 9 hunks across `pyproject.toml`, `tui/autoview.py`, `tui/widgets.py`, `tests/test_tui.py` and `uv.lock`. One of them is worth knowing about — in `tui/autoview.py` both sides add a *different* import at the same line and both are used, so taking either side wholesale drops the other's and still compiles. The rest are ordinary.

## Install

```bash
uv tool install 'claude-swap[pin]'
```

Ready for review. The split itself is the part I most want your read on — the package boundary, the command surface, and whether `clear_wiring()` living here rather than in the extra is the right call.

## Known costs, for the maintainer to weigh

**The dev group installs the extra, so every contributor and every CI job gets
`cswap-pin` + `cryptography` + `cffi` — Windows included, where `pin._impl()`
refuses the feature outright.** That is a third-party native crypto stack on a
platform that cannot use it, paid by people who are not using the pin at all.

It is deliberate, and the two narrower fixes were tried and rejected in-tree:

- **A `sys_platform` marker** would keep Windows out — and make
  `test_the_marker_still_matches_the_package_that_writes_it` decorative there
  rather than red.
- **`importorskip`** was the original shape. Measured: it reported "skipped" on
  every job on every platform, so the one check standing between a rename in
  `cswap-pin` and `cswap pin --clear` silently reporting "No cloud account
  pinned" over a live wiring had never run once.

cswap READS a marker `cswap-pin` WRITES and the two version independently, so
that check is the only thing holding the seam. Raising it here because the cost
lands on people who did not ask for the feature, and that is a maintainer's call
rather than something a reviewer should have to find in `pyproject.toml`.

**What is NOT at stake:** running without the extra. `TestImportSafeWithoutTheExtra`
and `TestClearRunsWithTheExtraGone` block `cswap_pin` at `sys.meta_path`, and the
whole suite was run that way — the only failures are the contract cases that
require the package by construction.

## Budget, re-measured after the clamp

`_LAUNCH_LOCK_BUDGET_S = 0.5` is a TOTAL across every config `clear_wiring`
touches, carved as `left / (len(paths) - i)` so the running total cannot exceed
it. That arithmetic was always right; the overshoot was the unclamped sleep
above. Measured with BOTH config locks held, three runs:

    idx=0 share=0.25 elapsed=0.25 [timeout]   idx=1 share=0.25 elapsed=0.25 [timeout]   TOTAL=0.500

Before the clamp the same shape ran 0.584s and 0.694s against a 0.5s budget.


codeslake:
Raised the extra's floor to `cswap-pin>=0.1.76` (206be49).

The previous range still admitted two builds that can take the pinned port down, and this extra is what installs them.

**0.1.74** is the one that matters. Its daemon asks its holder for a successor via `SIGUSR1` and assumed the holder could hear it. On an upgrade the holder is *always* the previous release — that code path fires only when the disk is newer than what someone loaded — so the handler does not exist, and an unhandled `USR1` terminates. Measured: the ask killed the holder, the port went `REFUSED` for 30 s, and service returned only because a standby noticed the free port. A machine whose standby has already been consumed has nothing to notice.

**0.1.72** was the earlier attempt at the same problem from inside the daemon; it made the gap worse (2 s → ~9 s) and was reverted in 0.1.73.

0.1.75 makes the capability the holder's to advertise and refuses to signal one that has not claimed it. 0.1.76 retires a holder the deploy left on the previous release — without that half a machine falls back to the slow path for as long as the holder process lives, and nothing recycles it (holder ages when this was found: 7 days and 3 days, above daemons that were minutes old).

Verified on three machines at 0.1.76 before raising the floor: two daemons overlapping on one socket through the handover, then the stale holder retired under a fresh lineage — 200/200, 200/200 and 45/45 completed requests through the proxy, zero failures.

codeslake:
Moving this back to draft while I re-measure against Claude Code **2.1.234**, which changed behaviour under the feature this PR is for. Not withdrawing it — the premises need to be re-established on the new binary before it is worth anyone's review time.

**What changed in the binary** (marker counts across the four releases installed here):

| | `Watching ~/.claude.json` | `identityChanged` |
|---|---|---|
| 2.1.231 | 0 | 0 |
| 2.1.232 | 0 | 3 |
| 2.1.233 | 0 | 3 |
| 2.1.234 | **6** | 3 |

2.1.234 added a live watch on `~/.claude.json` through a storage interface, with an explicit fallback in its own prose (`…failed; staying on the fs.watch`).

**What I can reproduce.** On a machine where the active account rotates, Remote Control drops with the binary's own string:

```
signed-in claude.ai account or organization changed on this machine — run /remote-control to reconnect
```

and the owner comparison is both fields, not the email:

```js
function hX(e,t){ return Boolean(e?.accountUuid)
    && e?.accountUuid === t?.accountUuid
    && (e?.organizationUuid||void 0) === (t?.organizationUuid||void 0) }
```

**What I have NOT established, and why that is the reason for the draft.** I first attributed the disconnect to the new watch driving `identityChanged`. That is wrong: `identityChanged` occurs three times in the binary — a symbol entry, its definition, and exactly one call site, which is the OAuth login flow (`startOAuthFlow(...).then(… identityChanged(…))`). The watch path refreshes the config cache and does not call it. So the trigger is still unidentified, and the PR should not be reviewed against a rationale I cannot support.

Two further things to re-check before this comes out of draft, both of which may narrow the change considerably:

- The veto appears to be **attach-time**, and its own log line says `the transcript record is left in place; this bridge is not persisted` — which contradicts the "history is lost" framing.
- The bridge constructor already takes an `ownerPin`, and `bridgeOwnerAccountUuid` / `bridgeOwnerOrganizationUuid` are persisted per bridge. If that pin can be driven, the compatible fix may live there rather than anywhere near the credential store.

One thing I have ruled out, so it does not resurface in review: holding `~/.claude.json`'s `oauthAccount` steady. The switcher reads that field to decide *whose* the live credential is, so pinning it makes a refresh pass adopt another account's credential believing it is its own. That is worse than the outage it would paper over.

I will report the measurements here and take it out of draft when the mechanism is pinned down.

codeslake:
Out of draft. The mechanism is pinned down, and it is not what I attributed it to when I moved this back — as promised, here are the measurements.

## The trigger, and it is neither the watch nor `identityChanged`

I was right to withdraw the first rationale and wrong about where to look next. The disconnect comes from the reattach check itself:

```
confirmChanged()  ->  a7t()  ->  POST ${BASE_API_URL}/api/oauth/validate
                                 Authorization: Bearer <token from ~/.claude.json>
```

If the server attributes that call to the bridge's owner, Claude Code logs `[bridge:owner-pin] … re-baselining` and returns `"unchanged"`, and the bridge survives. Otherwise it returns `"changed"` and you get the string in my previous comment. The `~/.claude.json` watch added in 2.1.234 refreshes the config cache and is not on this path; `identityChanged`'s single call site really is the OAuth login flow, as I said.

**So the third thing I listed to re-check was the right one.** The comment guessed that "if that pin can be driven, the compatible fix may live there rather than anywhere near the credential store", and that is exactly where it lives. `/api/oauth/validate` is now a pinned route in the extra (`cswap-pin` 0.1.92) — the proxy swaps the bearer on that one request, the server answers as the owner, and Claude Code re-baselines instead of disconnecting.

**Matched exactly, not by prefix, and that distinction is load-bearing.** `/api/oauth/token` is a *refresh*, and swapping the bearer there would mint one account's credential under another — the same objection that killed the hold-`oauthAccount` idea in my last comment. A prefix match on `/api/oauth/` would have reintroduced it silently.

Measured live: with the extra at ≥0.1.92, switching the active account through `cswap` left every open bridge `CONNECTED` and the disconnect string absent from the logs.

On the second item — the veto is attach-time and its own line does say the transcript record is left in place. What actually loses a session's continuity is the pointer comparison being made against `~/.claude.json`'s `oauthAccount` rather than against the pin, so a rotation between runs is what strands it. That is a narrower claim than "history is lost" and I have corrected it wherever I had written the broader one.

## Floor raised to `cswap-pin>=0.1.97`, relocked in the same commit

It had been at `0.1.77` while the package reached `0.1.97` — twenty releases of drift, which is precisely the rot the comment above that line was written about after `>=0.1.3` went unchanged through 36. Writing the rule down did not enforce it.

| | |
| :-- | :-- |
| `0.1.92` | the route above — Remote Control survives a `cswap switch` |
| `0.1.93` | the daemon log rotates instead of unlinking itself at 64 KiB, which used to destroy the outgoing daemon's teardown record at the instant it is read |
| `0.1.95` | a server-generated title is decided from what Claude Code recorded (`ai-title` / `custom-title`) rather than from the string's shape, which was overwriting names a human had typed |
| `0.1.97` | the drain can reach zero, so a proxy recycle stops cutting whatever is open |

The last one is the same shape as the bug above and worth recording, because this PR's structure is meant to make that shape findable and did not. `await_inflight` was changed in 0.1.94 to wait on requests *owed* rather than connections *open*, since Remote Control's WebSocket lives for the whole session and made the old zero unreachable. The fix cleared the debt at the MITM path's 101 handover — but `_blind_tunnel`, which that file's own docstring says is where Remote Control's receive channel actually lands (*not* `api.anthropic.com`), never cleared it. Every machine with RC open therefore held one permanent debt, and 0.1.93, 0.1.94 and 0.1.96 all produced an identical departure line. Three versions, one signature, which is the tell that a fix is not on the path it was written for.

Two defects in the log line kept it invisible: the count reported open sockets while the loop waited on owed answers, and `after {budget}s` printed the argument rather than the elapsed. Both fixed, and every drain now logs whether or not it cut, because "no line" previously meant both "drained clean" and "this daemon never drained".

## Also now flagged in the description

`oauth.py` **adds** the pin's CA to the default trust store rather than replacing it. Replacement narrows what the process trusts for every host, not only ours. `_pin_aware_ssl_context()` builds a default context and calls `load_verify_locations()`; with the extra absent it returns the default untouched, and the test drives both arms with the CA path injected so the no-pin control asserts rather than skips.

CI on this head: **2316 passed, 4 skipped** on Linux, plus the four Windows shards and the macOS keychain job.

The seam, the three decisions, and the #199 overlap note are unchanged from the description.

noahchamberlain-bunch:
Hello. Am eagerly awaiting this PR to be merged (I've been lurking in the background while this [issue was discussed](https://github.com/realiti4/claude-swap/issues/198)). Do you have an idea when this might happen?

codeslake:
Thanks for waiting, and for the issue thread — it is what shaped this.

**You do not have to wait for the merge. The branch installs and works today:**

```bash
uv tool install 'claude-swap[pin] @ git+https://github.com/codeslake/claude-swap@feat/pin-package-seam'

cswap list        # the account numbers come from here
cswap pin 2       # Remote Control / Artifacts -> account 2
cswap pin         # show the current pin
cswap pin --clear # remove it
cswap pin --heal  # restart a pin proxy that died
```

The `[pin]` extra pulls `cswap-pin` from PyPI on its own, so that half is always the released one; only the wiring lives on this branch. `/v1/messages` is never touched, so inference still bills whichever account `cswap switch` / `cswap auto` steered you onto — Remote Control and Artifacts are what stay pinned.

Please do try it. A report from a machine that is not ours is genuinely useful, and if it misbehaves for you that is worth more to this PR than another week of our own testing.

**On timing:** Claude Code ships close to daily, and several of the surfaces this PR sits on move with it. Rather than chase each release we have been converging the branch on behaviour that survives an update — probing the installed signature and falling back instead of pinning a version, treating an unreadable state as UNKNOWN rather than as a failure, and keeping the pin's own correctness floor separate from a freshness one. That is what the recent churn here is.

It is green on CI and mergeable now. I would rather land it once it has run unattended across a few CC releases than land it and have it break on the next one.

PR #211, opened 2026-08-02
Auto-switch: reset-aware grace window + settings.json hot-reload

ianderrington:
## Motivation

`cswap auto`'s switch decision is currently pure `utilization >= threshold`, with no awareness of how soon the active account's own binding window resets. That produces a needless switch in a common case:

- Threshold: 90% (default)
- Active account: 91% used, its 5h window resets in ~5 hours
- A second account sits below the threshold

Today this immediately proactively switches to the second account — even though the active account's own window is going to fully refresh regardless of whether it's used further in the meantime. The switch doesn't preserve anything (the "saved" headroom is wiped out by the reset anyway); it just starts spending quota on an account that didn't need touching.

This is most visible right at threshold-crossing with a fast-cycling 5h window, but the same shape applies any time a window's reset is close relative to how much runway is left.

## Change 1: `autoswitch.graceBeforeResetMinutes` / `--grace-before-reset`

New setting (0–1440 minutes, default `0` = off, same validation/clamping pattern as the other `autoswitch.*` keys):

```bash
cswap auto --grace-before-reset 300
# or
cswap config set autoswitch.graceBeforeResetMinutes 300
```

When set, a **proactive** switch (active account over the threshold but not yet at its hard limit) is skipped if the account's own *binding* window (the one actually driving the trigger — reused via the existing `_binding_recovery_ts` helper, same one the all-accounts-exhausted escape already uses) will reset within the grace window. The engine rides the account to ~100% instead of switching away.

Deliberately scoped:
- Only applies to the `"proactive"` trigger. `"at-limit"`/`"failover"` are left untouched — those mean the account is already unusable, so switching is a strict improvement, not a tradeoff.
- `"consume-first"` strategy is untouched too — it already has its own reset-driven logic for below-threshold moves that this would otherwise fight.
- A grace long enough to cover the whole 5h window (e.g. `300`+) effectively stops that fast-cycling window from ever triggering an early switch, while a multi-day 7d window reset is essentially never inside a reasonable grace value, so it still switches normally when *that's* what's binding.
- An unknown/missing `resets_at` never suppresses a switch — can't prove "imminent" without it.

## Change 2: `settings.json` hot-reload for the CLI loop

Separately: a plain `cswap auto` foreground/background loop currently loads `settings.json` once at process start (`merged_with_cli(load_settings(...), args)`) and then loops forever on that frozen value. `AutoSwitchEngine.apply_threshold()` exists to hot-apply a new threshold, but it's only ever wired up from the TUI's session slider and the menubar app's explicit engine restart — the plain CLI loop never called it. So `cswap config set autoswitch.threshold 95` from another terminal while `cswap auto` is already running is silently ignored until the process is killed and restarted.

Fix:
- `merged_with_cli` is refactored into two reusable pieces — `cli_overrides(args)` (extract the override dict) and `apply_overrides(settings, overrides)` (apply + clamp) — so the reload path can share the exact same override-application logic as startup.
- `AutoSwitchEngine` gains `apply_settings()` (generalizes `apply_threshold()` to every knob except the model axis, which stays fixed at construction — same invariant `apply_threshold()` already documented) and `cli_overrides`/`watch_settings` constructor params.
- The CLI loop (`_auto_command`, loop mode only — `--once` already reloads fresh every invocation) now passes `watch_settings=True`; `run_loop()` mtime-checks `settings.json` each iteration (same technique the menubar app already uses for its own active-account-change detection) and hot-applies changes via `apply_settings()`, preserving CLI-flag precedence over the file and emitting a new `settings-reloaded` event so it's visible in the log.
- TUI and menubar are untouched — they already have their own live-apply paths and shouldn't get a second, uncoordinated writer of `engine.settings`.

## Testing

- 13 new tests: `TestResetGrace` (6 — default-off, suppression, beyond-grace, never-at-limit, unknown-reset, 5h-vs-7d-window interaction) and `TestSettingsHotReload` (7 — reload + event, no-op when unchanged, CLI-override precedence, model-axis pinning, `apply_settings`, and the loop's `watch_settings` on/off wiring) in `tests/test_autoswitch.py`.
- Existing `test_config_cli.py` list/JSON tests updated for the new 10th setting key.
- Full suite: 1702 passed, 3 skipped, 6 pre-existing failures unrelated to this change (verified against a clean `git stash` baseline — `test_move_accounts.py`/`test_swap_accounts.py` rollback tests, untouched by this diff).
- README's "Automatic switching" section updated with both behaviors.

Happy to adjust naming/scope/defaults if you'd rather see this shaped differently — opening this as a concrete starting point rather than a final API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ Let claude-swap watch your usage and switch for you. When the active account's 5
 ```bash
 cswap auto                     # foreground loop, polls every 60s
 cswap auto --threshold 80      # switch earlier
+cswap auto --grace-before-reset 300   # ride an about-to-reset account to ~100%%
 cswap auto --model Fable       # also switch when the Fable weekly limit is hit
 cswap auto --once              # single check-and-switch, for cron/scripts
 cswap auto --dry-run           # log what it would do, never switch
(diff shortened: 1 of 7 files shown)


ianderrington:
Closing this — folding the grace-window idea for now. It's solving a narrower problem (don't switch away from the active account right before its own window resets) with no real usage evidence behind it yet, while the actual bug worth fixing was in the consume-first drain ranking itself: see #(new PR) for a fix to #305 plus a proper 7-day -> 5-hour -> most-used priority chain. May revisit grace-before-reset separately once there's a concrete case for it.

PR #212, opened 2026-08-03
fix(session): prevent bootstrap credential churn

Coderman562:
## Problem

Claude Code 2.1.220 changed subscription auth status from `authMethod: claude.ai` with inline identity fields to `authMethod: oauth_token`, `apiProvider: firstParty`, with identity omitted.

Session bootstrap treated every unknown command result or schema as a confirmed invalid profile. Each retry could then delete the profile Keychain item, proactively refresh the backup refresh token, seed the new generation, fail the same validation again, and delete the profile. A minute-cadence launcher can consume many one-time refresh-token generations and eventually turn healthy accounts into `invalid_grant` failures.

## Fix

- Remove proactive OAuth refresh from session bootstrap. Bootstrap copies the stored credential once; Claude Code is the only refresher of the profile token family.
- Make session validation tri-state: valid, invalid, or indeterminate.
- Accept Claude 2.1.220 first-party `oauth_token` status and recover identity from the profile config when the CLI omits it.
- Treat command failures, timeouts, malformed JSON, and unknown future schemas as non-destructive indeterminate results.
- Never rebuild a profile while any Claude process is live against it.
- Require the existing stale marker before reseeding an existing non-live profile; a validation probe alone cannot authorize deletion.
- Preserve explicit logged-out, API-key, wrong-email, and wrong-organization detection without deleting the profile after failure.

This is intentionally narrower than #196: session startup consumes no refresh grant at all instead of routing bootstrap through a broader consume gate.

## Validation

- `tests/test_session.py` plus a real subprocess `cswap run` test: 162 passed.
- Black-box test runs the CLI twice against a fake Claude 2.1.220 executable and asserts the backup generation is byte-identical.
- Live macOS check against an already-running profile: backup and session fingerprints were identical before and after `cswap run 5 -- --version`.
- No credential contents are logged or asserted outside synthetic fixtures.

### DIFF
--- a/src/claude_swap/session.py
+++ b/src/claude_swap/session.py
@@ -45,6 +45,7 @@ import sys
 import tempfile
 import time
 import unicodedata
+from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, NoReturn
 
@@ -58,7 +59,6 @@ from claude_swap.exceptions import (
 from claude_swap.fsutil import replace_with_retry
 from claude_swap.locking import FileLock
 from claude_swap.models import Platform
-from claude_swap.oauth import refresh_oauth_credentials
 from claude_swap.paths import get_default_global_config_path
 from claude_swap.printer import accent, dimmed, muted, warning
 from claude_swap.process_detection import ClaudeSession, list_sessions
@@ -141,6 +141,12 @@ _AUTH_STATUS_TIMEOUT = 10.0
 # timeout) plus auth-status probes, so it needs more headroom than the
 # default 10s acquire used by the switch paths.
 _BOOTSTRAP_LOCK_TIMEOUT = 30.0
+
+
+class _SessionValidation(Enum):
+    VALID = "valid"
+    INVALID = "invalid"
+    INDETERMINATE = "indeterminate"
 
 
 def slugify_email(email: str) -> str:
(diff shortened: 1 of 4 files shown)


realiti4:
Thanks a lot for the PR — the churn issue is real and the direction here (tri-state validation, no proactive refresh, never touching live profiles) is exactly right.

Can we check couple of findings before merging:

1. **Drift check on indeterminate probes** — when the probe times out or fails, `setup_session` reuses the profile without checking `session_identity_drifted`. So if someone did `/login` to a different account inside the session and exited, `cswap run A` can launch account B when the probe happens to fail. I reproduced this on the branch. A drift veto before the indeterminate reuse returns should cover it.

2. **Marker ordering in `_invalidate_session_credentials`** — it deletes the keychain entry and creds file before writing the stale marker, and `mark_session_stale` swallows write errors. A crash in between leaves a credentialless profile with no reseed authorization. Writing the marker first (and skipping the deletion if it fails) would make the invariant hold.

3. **The 2.1.220 schema claim** — I couldn't reproduce it. I ran the real 2.1.220 and 2.1.221 binaries and both report `authMethod: "claude.ai"` with email/orgId inline for stored subscription logins. I only get `oauth_token` with `CLAUDE_CODE_OAUTH_TOKEN` set in the env, which our probe scrubs. Was your observation maybe from a shell with a token env var exported, or something macOS-specific? Would like to understand what actually triggered it on your machine.

One small related note: real claude exits with code 1 when logged out, so the `loggedIn: false` → invalid branch never actually runs (rc≠0 short-circuits to indeterminate first). That's fine behavior-wise, but the fake in the tests returns rc=0 there — worth making it match the real CLI.

Thanks again, happy to merge once these are addressed.

PR #213, opened 2026-08-03
fix(auth): keep session profiles authoritative across credential generations

### LINKED ISSUE #187: bug: idle session expiry can falsely quarantine account backup

Coderman562:
## Problem

An inactive account with a matching `cswap run` profile can be falsely quarantined after that profile becomes idle and its access token expires.

Claude Code may have rotated the profile onto refresh-token generation B while the stored backup still contains consumed generation A. The collector currently trusts B only while a process is live; once idle, expiry falls through to the backup refresh path, POSTs A, receives `invalid_grant`, and records the healthy slot as `relogin_required`.

A later login/profile refresh can create generation C, but the usage cache stores only a generation-blind dead-token strike. That leaves C quarantined even when native Claude can authenticate successfully.

## Fix

Keep an existing matching session profile credential-authoritative whether or not a Claude process is currently alive:

- fresh profile access tokens remain usable for read-only usage fetches;
- expired profile credentials surface `token_expired`;
- cleared, partial, or unreadable profiles never fall back to a possibly consumed backup;
- native Claude refreshes the profile on its next real turn;
- each `invalid_grant` strike records the credential fingerprint that failed;
- a complete newer profile generation releases only its predecessor generation’s quarantine;
- legacy fingerprint-less rows bind to the stored generation that originally produced them;
- strict Keychain reads prevent stale plaintext fallback from falsely healing quarantine;
- RFC-style JSON classification keeps `invalid_client` systemic instead of condemning an account token.

Profiles whose identity has drifted to another account still fall back to the intended slot backup, and accounts with no session profile retain the existing backup path.

## Validation

- Rebased onto current upstream `main` on August 6, 2026.
- 715 changed-area tests pass; Ruff passes on all changed files.
- The current upstream full macOS run reaches 1,833 passed and 3 skipped. Its one existing strict-clear rollback failure is the separate credential-clear issue covered by #214, not this diff.
- The combined production deployment branch with #214 passes 1,714 tests with 3 platform-specific skips.
- Regression coverage binds quarantine to the failed fingerprint, releases it after a complete generation change, migrates legacy rows, keeps the same generation blocked, rejects cleared-profile fallback, and distinguishes exact `invalid_grant` from `invalid_client` or marker substrings.
- A controlled native Claude request succeeded through the newer profile while the old cache still reported `relogin_required`; the fixed collector then released only that stale quarantine. A genuinely cleared sibling profile remained quarantined.

Fixes #187.

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -3441,16 +3441,13 @@ class ClaudeAccountSwitcher:
                         error=outcome.error,
                         retry_after_s=outcome.retry_after_s,
                     )
-                if has_live_session:
-                    # The live claude refreshes lazily on its next API call;
-                    # requesting now would just 401 (same rule as the owned
-                    # active account in _fetch_active_usage).
-                    return FetchRecord(sentinel=USAGE_TOKEN_EXPIRED)
-                # Expired profile credential and no live session: fall through
-                # to the backup path — cswap must not rotate the profile's
-                # family, but a backup family that is still alive (e.g. the
-                # account was re-added after the profile last ran) can serve
-                # and heal via the normal refresh machinery below.
+                # The profile remains the credential authority while idle:
+                # Claude may have rotated its refresh-token family without
+                # synchronizing the stored backup. Falling back on expiry can
+                # therefore POST a consumed predecessor and falsely quarantine
+                # a healthy account as invalid_grant. Native Claude refreshes
+                # the profile lazily on its next real API call.
+                return FetchRecord(sentinel=USAGE_TOKEN_EXPIRED)
 
         outcome = oauth.try_fetch_usage_for_account(
             str(num), email, creds,
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import json
-import logging
 import os
 import re
 import shutil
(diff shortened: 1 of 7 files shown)


src/claude_swap/switcher.py
```diff
@@ -3441,16 +3446,23 @@ def persist(acct_num: str, acct_email: str, new_creds: str) -> None:
...
+                return FetchRecord(sentinel=USAGE_TOKEN_EXPIRED)
+
+            # A non-stale session profile remains Claude Code's credential
+            # authority even after its credential was cleared or became
+            # unreadable. Falling through to the stored bootstrap copy would
+            # POST a potentially consumed predecessor and recreate the exact
+            # invalid_grant churn session mode is designed to avoid.
+            if session_exists and not session_stale and not session_drifted:
+                return FetchRecord(sentinel=USAGE_NO_CREDENTIALS)
+        elif session_exists and not session_stale and not session_drifted:
```
m4ttheweric, line 3464: **issue (non-blocking):** i think this strands a slot right after `cswap add` re-adds it. the add path ends in `_invalidate_session_credentials`, which drops the profile's creds but keeps the dir and `.claude.json` and clears the stale marker... exactly this branch, so the fresh backup never gets tried.

probed the re-add state both ways on upstream/main:

```
with this branch:  sentinel='keychain unavailable'  backup_fetched=False
without:           usage={'five_hour': {'pct': 4}}   backup_fetched=True
```

could a deliberately invalidated profile (dir present, creds gone, no stale marker) still fall through? happy to share the repro test. hit the original bug on two of my own accounts, so glad someone's on this one.

PR #216, opened 2026-08-04
fix(add): refuse a credential whose owner is not the account being added

codeslake:
`add_account` reads the IDENTITY from `.claude.json`'s `oauthAccount` and the CREDENTIAL from the keychain/file store. Those are two independent sources and nothing asked whether they agree.

## The defect

Measured in the field: a session registered one account and the slot received a **different** account's credential. The shape was an ssh session whose `.claude.json` had been renamed to the new profile while the live keychain item still held the original account's token.

Reproduced here — config naming `ax@example.com`, token belonging to somebody else:

```
Added Account 7: ax@example.com
```

Exit 0, no warning. The slot is **labelled** one account and **contains** another, so every later switch to it logs the wrong user in and `--status` shows a name that is not whose token is stored. Nothing surfaces the disagreement.

## The fix

`oauth.fetch_oauth_profile` already answers exactly this question ("whose token is this") and the autoswitch identity oracle already uses it. This adds the same call to the one path that creates slots.

**Advisory**, in the direction the oracle is advisory everywhere else: a `None` answer means UNRESOLVABLE (offline, 401, schema drift), never "wrong", and must not block a registration that worked before this guard existed. Only a RESOLVED identity that disagrees refuses.

The organization is compared too, not just the email. The codebase already states why, where it compares identities elsewhere:

> `_live_identity_matches`: "Compares the organization too: two managed slots may share an email across orgs."

That arm is mutation-driven, not speculative — dropping `and seen_org == ...` left 477 tests passing until the org test was added.

**Email first, then organization.** `fetch_oauth_profile` returns
`organizationUuid: None` for any account with no organization block, which is
every personal account — not a corner case. An absent organization is
unverifiable about the *organization*; it says nothing about the address. So a
disagreeing email refuses regardless, and `None` is consulted only once the
email agrees. That is the shape `_resolved_matches_slot_identity` already uses:
its `if r_org is None: return None` sits inside the branch where the email has
already matched.

```
                             refuses?
foreign email + org None     yes
foreign email + org ""       yes
own email + org None         no — registers, org unverifiable
own email + different org    yes — same address, different organization
```

**An expired access token is not a dead one.** `is_oauth_token_expired` reads
the ACCESS token, and a live refresh token revives it — `session.py` refreshes
exactly this way at switch time. Skipping the check on expiry would store a
foreign credential that is one refresh away from working. The guard refreshes
first and resolves identity from the refreshed token.

The credential it stores is the credential it verified. A refresh grant may
return a rotated `refresh_token`, and accepting one retires its predecessor
server-side, so persisting the pre-refresh bytes would leave the slot holding a
refresh token the server has already invalidated. On every path where no
refresh happens the guard returns its input unchanged.

```
expired foreign, refresh succeeds   refused, resolved with the refreshed token
expired own,     refresh succeeds   registers, stored refreshToken is the rotated one
expired,         refresh fails      registers — unresolvable is not wrong
no refresh token                    registers, 0 profile fetches
```

No lock is held across either network call: both call sites sit outside every
`with` block in `add_account`.

## Mutation table

In place, restored after each, controls both directions:

| Mutation | Result |
|---|---|
| drop the guard call | 1 failed (the mismatch test) |
| make `None` REFUSE | 19 failed (the advisory contract) |
| compare email only | 1 failed — survived before the org test existed |
| org check above the email check | 2 failed (create and refresh-in-place) |
| skip the check on an expired token | 1 failed |
| store the pre-refresh bytes | 1 failed |
| NOOP comment | passed, as it must |

## Tests

16 in one file, roughly half of them controls, because a refusal that refused
everything would otherwise read as a pass. Both call sites — creating a slot
and refreshing one in place — are covered by the same cases; the organization
cases ride one `parametrize(["", None])` rather than being written twice.

Suite: **1842 passed, 3 skipped**.


codeslake:
Housekeeping: **this branch was force-pushed.** Author/committer identity and message metadata were normalized across its 5 commits. If you have a local copy it will not fast-forward — re-fetch rather than merge.

No content changed. Measured:

```
pre-rewrite  13930df  tree b90e13f015c3
post-rewrite 1f7ff9a  tree b90e13f015c3   <- identical
```

`upstream/main` was not touched and is still an ancestor. #199 got the same treatment.

The PR's own content is unchanged and still MERGEABLE.

realiti4:
Hi, thanks a lot for this. I think verifying ownership at capture time is the right direction. #237 closed one specific cause of this drift, but this check is what catches the whole thing.

Could we address two things before a merge though: the expired-token refresh arm should go (or become lifecycle-safe, details below), and the comparison should be uuid-first like the rest of the codebase. The advisory fail-open contract is right and should stay exactly as you built it.

Detailed findings below are from my review pass with Claude Code:

---

Full review on 1f7ff9a: line-level pass of the guard and both call sites, full suite locally (2103 passed; the 3 local failures reproduce identically on main, so not this PR's), CI green on all three platforms, plus a comparison against how the rest of the codebase handles the two acts the guard performs — consuming a refresh grant, and comparing identities.

## 1. The refresh arm consumes a grant outside the lifecycle discipline everything else applies to that act

Everywhere else in this codebase, POSTing a refresh grant is a coordinated transition: the collector's path runs it under the account FileLock plus CC's credential locks (`switcher.py:4160-4168`), and afterwards "the successor MUST survive" — it's written to **both** the backup and the active store (`switcher.py:4244-4285`), because an accepted rotation retires its predecessor server-side (the premise this PR itself states). The guard calls the bare `refresh_oauth_credentials` with no lock, and:

- **Success path:** the rotated successor lands only in the slot backup. The active store keeps the spent generation, so the running claude session's next refresh gets `invalid_grant` and the user is logged out. No later switch heals this: the unmanaged auto-add returns without switching (`switcher.py:5551-5556`), and the menubar's refresh (`menubar.py:881`) is the same shape.
- **Refusal path:** the guard raises after the refresh and the rotated credential is discarded entirely — the only usable copy of that lineage's live generation is dropped, and "Nothing was changed" in the error text is inaccurate (server-side state did change).
- **Race:** an unserialized POST can collide with the collector's locked refresh of the same lineage; with `AUTH_DEAD_STRIKES=1` the loser strikes a live slot.

The cleanest PR-sized fix is to **drop the refresh arm**: expired → unresolvable → registers, exactly like an offline profile fetch. That deletes the entire risky surface and keeps the guard fully effective for the field incident's shape (a live token). Expired-foreign detection is real but belongs in a follow-up built on the consume-gate machinery. If you'd rather keep it now, it needs the collector path's locks and both-store persistence, plus preserving the successor on the refusal path — the grant response's `tokenAccount` (`oauth.py:198-224`) can then replace the second profile request, though it doesn't remove the persistence requirement.

## 2. Compare the uuid first

Both sides already carry one: the profile result's `account.uuid` is guaranteed non-empty whenever the guard has a resolution at all (`oauth.py:240-243`), and `add_account` already reads the config's `accountUuid` (`switcher.py:3366`). The guard compares only email/org. The codebase's own comparators are uuid-first because emails can be recycled across accounts (`_resolved_matches_slot_identity`, `switcher.py:2901-2903`). Email-only fails in both directions: it refuses a legitimate add after an account email change (stale config address, same uuid), and accepts a deleted/recreated account under a recycled address. Uuid-first with email/org as fallback also simplifies the org arm.

## 3. Smaller items, batch with the fixes

- Derive the identity from the same config read whose bytes get stored: `add_account` reads `.claude.json` once at the top for identity and again near the write, so a `/login` landing during the guard's network window pairs one account's token with another's metadata — the exact class this guard exists to close. The race predates the PR (the window was previously sub-second); the guard widens it, and the fix is nearly free here. A credential-fingerprint recheck before committing covers the side a config snapshot can't.
- The guard's docstring still says the refresh result "is never persisted here" while the code returns it for storage — 650a61b changed the behavior intentionally (the mutation table pins it) but the docstring wasn't updated.
- The test docstring's "a plain `cswap` switch's auto-add" overstates that branch's reach: managed-account switches go through `_perform_switch` and never call `add_account`; the auto-add fires only for an unmanaged active account (`switcher.py:5539-5556`).
- Optional: print a visible notice when ownership couldn't be verified, so fail-open isn't silent.

## What checked out

The advisory contract and the email-then-org ordering with `None`-as-unverifiable mirror `_resolved_matches_slot_identity` (`switcher.py:2937`) exactly; the refusal precedes every destructive write in both branches; the org-mismatch message naming both organizations is right. On overlap with #237: none — that fixed the env-var profile divergence in the *read* path, while this incident is content-level (a config file swapped within one profile), which only a capture-time ownership check catches. The controls-heavy test file and the autouse profile stub are the right calls (one caveat on the stub: it converts a forgotten mock in a future test into a silent "unresolvable" rather than an error).

codeslake:
Thanks — this was a good review and it changed the shape of the PR. All of it is in at `5718c1d`. I reproduced each finding before agreeing, and the two blockers were both real.

## 1. The refresh arm is gone

You are right that it consumed a grant outside the discipline every other caller applies to that act, and the refusal path was the worse half: it raised *after* the rotation, discarding the only usable copy of that lineage's live generation, while the message said "Nothing was changed".

Dropped entirely. Expired now means unresolvable and the add proceeds, exactly like an offline profile fetch.

I kept the knowledge that dies with it, as a test rather than a comment — `test_an_expired_FOREIGN_credential_registers_and_that_is_deliberate` asserts the registration *and* that no grant moved, so the next person to notice the gap finds out it was priced rather than missed. Expired-foreign detection on the consume-gate machinery sounds right as a follow-up.

## 2. Uuid-first — and my first attempt at it was a regression

Done, but the obvious version of it was wrong and I want to flag that rather than let it pass silently. Comparing uuid-first with an early `return` on a match makes the guard **accept a uuid match under a disagreeing org**, and worse, it makes the org-mismatch message — the one you verified as correct — unreachable for every config Claude Code writes, since those all carry an `accountUuid`.

Measured against `0c1569a`: the cross-org case was refused before, and my first uuid-first draft registered it. The uuid arm now falls *through* to the same org block the email arm reaches, so one org check serves both.

One correction to the citation, since I checked it: `_resolved_matches_slot_identity` tolerates a missing org on *either* side (`not r_org or not own["organizationUuid"] or ...`), so "definitively another account" holds only when both orgs are present. This guard is deliberately stricter there — a capture is a one-time write, not a per-switch check — and the comment now says so instead of appealing to the sibling for more than it says.

`accountUuid` is read at `switcher.py:3473`, not `:3366` — the fact was right, the line was off.

## 3. One config read, plus a commit-time recheck

`_get_current_identity_triple()` reads once and returns `(email, org, uuid)`, and `_get_current_account` now delegates to it so there is a single reader. Your fingerprint suggestion pointed at something a snapshot cannot cover, so there is also an explicit recheck that the identity verified is still the identity about to be stored.

**That recheck initially covered only one of the two write paths**, and it was the wrong one: `slot=None` on an already-registered account is the branch the menu bar and the TUI take. Lifted into `_reject_identity_drift_since_verify` and called from both. Reproduced first — a `/login` landing inside the guard's network window left slot 7 labelled one account with the other account's config stored, no exception.

Two smaller things from the same area:
- `accountUuid: null` versus `""` made that recheck fire spuriously, with a message naming the *same* address as both before and after. Both sides normalise now.
- The auto-add claim in the test docstring was wrong in the direction I "narrowed" it: the auto-add fires only when the account is **unmanaged**, and the in-place branch requires that it **is** managed — the same predicate on the same two arguments, so it can never land there.

## 4-6. Docstrings and the silent fail-open

Guard docstring rewritten to the contract that now exists. Test docstring corrected as above. Every unresolvable exit now prints a notice — registering with the ownership question unanswered is the state the field incident was in, so it should not be silent.

## Verification

Whole suite: **2114 passed, 3 skipped**. Each change mutation-checked in isolation — restore the refresh arm → 4 tests fail; make the uuid arm return early → 1; un-normalise the second read → 1; take the drift check off the in-place path → 1.

codeslake:
Closing this at the project owner's direction.

The last four commits added a seam that consults an optional `claude_swap.pin`
package. That is pin functionality, and it belongs with the pin work rather
than here — the pin branch already owns the implementation and six of the
seven call sites, so this PR was carrying a seventh hook with nothing behind
it. It is being handed over.

It was also already breaking on the integration branch: the test here fakes
the module through `sys.modules`, but `from claude_swap import pin` resolves
the parent package attribute first, so once the real module is present the
fake is never consulted. Green alone, red merged, invisible to CI on either
side.

The first five commits are a generic identity guard with no pin surface —
refusing a credential whose owner is a different account, plus the
refresh-in-place hole, the org-None bypass and the expired-token blind spot.
If those are wanted they will come back as their own PR with no pin
involvement.

codeslake:
Thank you for this, and I am sorry it sat for days without an answer. The delay was mine, not a disagreement with any of it.

Taking the review as written. Point by point:

**1. The refresh arm goes.** Your reading is right and the reasoning is the part I want to acknowledge rather than just comply with: the guard POSTs a grant outside the lifecycle discipline every other call site applies to that act, and both outcomes are bad in ways the error text denies. On success the rotated successor lands only in the slot backup while the active store keeps the spent generation, so the running session's next refresh gets `invalid_grant`. On refusal the rotated credential is discarded and "Nothing was changed" is false, because server-side state did change. And the unserialized POST can collide with the collector's locked refresh of the same lineage.

Dropping it is the right size for this PR: expired becomes unresolvable and registers, exactly like an offline profile fetch. That deletes the surface rather than guarding it, and the guard stays fully effective for the shape the field incident actually had, which is a live token. Expired-foreign detection belongs in a follow-up on the consume-gate machinery, as you say.

**2. uuid-first.** Agreed, and the argument that convinces me is that email-only fails in *both* directions rather than one: it refuses a legitimate add after an account email change (stale config address, same uuid) and accepts a deleted-and-recreated account under a recycled address. Both sides already carry the uuid, and `_resolved_matches_slot_identity` is uuid-first for this reason, so the guard being email-first was an inconsistency and not a deliberate choice. It also collapses the org arm, which is a simplification I had not seen.

**3a. Same config read.** This one is already in: the caller was unpacking the config triple into three names and rebuilding a tuple from them, which could pair one account's address with another's uuid. Both call sites now pass the triple that was read, by identity rather than by value, and there is a test that fails on `==` and passes on `is` precisely so a rebuilt-but-equal tuple cannot satisfy it. A credential-fingerprint recheck before committing is not in yet and I agree it covers the side a config snapshot cannot.

**3b, 3c.** Docstring and test docstring both wrong as you describe, and both will be corrected with the fixes above rather than separately.

**3d.** A visible notice on unverifiable ownership is worth having. Fail-open being silent is the property that makes the autouse stub caveat you raise real, so I would rather it speak.

**One change since your review pass.** This PR previously carried a seam that consulted an optional package before refusing, so a deployment where something other than `/login` writes `oauthAccount` could say so. That has been removed (754d1fc). The hook lived here while its implementation lived elsewhere, which is the wrong side of the line, and the split was unsound on its own terms: a branch without the package has to fake the module, and `from claude_swap import pin` resolves the parent package attribute before consulting `sys.modules`, so once the real module exists the fake is never reached. Each side green, merged red, invisible to either CI. It does not belong in this PR and the diff you reviewed is the better one for it being gone.

The advisory fail-open contract stays exactly as it is, per your note.

codeslake:
Correcting my previous comment. I wrote it as a plan and most of it is already in the branch — I answered your review against the sha you reviewed rather than against the current head, which is exactly the mistake your review warns about in a different place. Sorry for the noise.

Checked against `754d1fc7`, quoting the code rather than the docstring:

**1. The refresh arm is already gone.** Expired is unresolvable, not refreshed:

```python
oauth_data = oauth.extract_oauth_data(creds)
if oauth_data and oauth.is_oauth_token_expired(oauth_data.get("expiresAt")):
    return unverified("the access token is expired")
```

That is the shape you recommended — expired becomes unresolvable and registers, like an offline profile fetch. No grant is spent, so none of the three failure modes you described (successor stranded in the slot backup, rotated credential discarded on the refusal path, collision with the collector's locked refresh) can arise from this path.

**2. The comparison is already uuid-first.**

```python
seen_uuid = (profile.get("uuid") or "").strip()
if account_uuid:
    if seen_uuid != account_uuid:
        raise ConfigError(... "the token resolves to account {seen_uuid}, not {account_uuid}" ...)
else:
    seen = (profile.get("email") or "").strip()
    if not seen:
        return unverified("the resolved identity carries no address")
    if seen.lower() != email.lower():
        raise ConfigError(...)
```

The uuid decides whenever the slot carries one; the email decides only when it does not. The org is corroborated after either arm, so a uuid match under a disagreeing org is still refused.

**3a. Same config read** — in, as I said, and both call sites pass the triple by identity.

**3b, 3c.** Both docstrings are already corrected; neither string you quoted is in the file any more.

**3d. The notice is in.** Fail-open is no longer silent:

```
Notice: could not verify that the stored credential belongs to <address> (<why>).
Registering anyway; re-run where the check can complete to confirm.
```

**What is genuinely still open is the one residual you named under 3a:** a credential-fingerprint recheck before committing, covering the side a config snapshot cannot. The `credential_fingerprint` comparisons in this file are all in the collector's locked-refresh path; there is none in `add_account`'s guard-to-commit window. That is the remaining item and it is mine.

So the diff is further along than my last comment implied. If it is easier for you, the review is worth re-reading against `754d1fc7` rather than `1f7ff9a`.

realiti4:
Hi, sorry for the delay on my side. All look good, thanks for the adjustments. Before merge can we check these too please, one of them is a blocker too it looks like.

Detailed findings below are from my verification pass with Claude Code:

---

Verified on `015fb6e`: full suite locally 2114 passed + the same 3 pre-existing `main` failures, CI green on all three platforms, net diff since `5718c1d` read line by line, no `pin` references remain after `754d1fc`.

## 1. Blocker: `_writer_candidates` can capture secrets, and belongs with #117

The `/proc` instrumentation (`switcher.py:306-367`, persisted at `:6516`) has a secret-capture problem the truncation doesn't fix: `argv[:160]` truncates but doesn't redact. This repo's own documented command line is `cswap add-token sk-ant-oat01-...` — the full token sits well inside 160 characters — and other processes' cmdlines can carry the same shape (`curl -H "Authorization: Bearer ..."`, `-p` passwords), all landing in the plaintext stash manifest. Two more reasons it belongs elsewhere: it instruments the switch-time stash, not this guard — the same line drawn when the pin seam was removed — and it returns empty on macOS, the platform every measured incident was on.

Dropping it from this PR is the fix here. The 31-minutes-apart double landing is genuinely valuable #117 data worth recording on that issue either way, and an attribution design (allowlisted executable names, or real write-event attribution rather than arbitrary argv) would fit as its own PR.

## 2. Small: run the in-place rechecks after the bytes they license

In the refresh-in-place path the two rechecks run *before* `config_path.read_text()` — the config bytes actually stored — so a `/login` landing between them still slips through. The create path already has the right order (read first, recheck after). Moving the two calls below the read is the whole fix. The residual race past that point is the pre-existing one; making the final revalidation-plus-write a critical section under the existing lock order is real but bigger than this PR — follow-up-issue shaped, not a merge condition.

## What checked out

- The credential-fingerprint recheck is exactly the residual from the earlier review: lineage hash on both write paths, fail-open on an unreadable store, access-token-only rotation deliberately equal, pinned by the rotation-mid-lookup test.
- The silent `return creds` on `organizationUuid=None` is correct as-is: by that point the uuid (or email) has already matched, so ownership of the *account* was verified — a "could not verify" notice there would misstate what happened. The design table's "own email + org None → registers, org unverifiable" stands as deliberate.
- The advisory fail-open contract is unchanged, as agreed, and the triple-by-identity change with its `is`-pinned test is sound.

codeslake:
Both done in `068814c4`. I reproduced each before agreeing, and pushed back on one word of the second — details below.

## 1. Blocker — `_writer_candidates` dropped

You are right, and the code convicted itself. Its own docstring reads *"NEVER
the environment and never a token"*, and the test that guarded it asserted only
`"env" not in c`. A promise in a comment, checked on the wrong half.

Confirmed here:

- `argv[:160]` truncates and does not redact.
- `README.md:332` documents `cswap add-token sk-ant-oat01-...`, and the token
  is well inside 160 characters.
- The result lands in the stash manifest as `"writerCandidates"`, plaintext.
- `if not os.path.isdir("/proc")` returns an empty list, so on the platform
  every measured incident came from, it says nothing.

Removed: the function, its private `import datetime as _dt` (used nowhere
else), the call site, and the test that asserted the field. 65 + 1 + 42 lines.

The 31-minutes-apart double landing goes on #117 as you suggest. I agree an
attribution design belongs on its own branch — allowlisted executable names, or
real write-event attribution rather than arbitrary argv.

## 2. The recheck order, with one deviation

Confirmed: the refresh-in-place path ran both rechecks ahead of
`config_path.read_text()`, so a login landing between the check and the read
stored a config the check never saw. Fixed.

**I moved only the identity recheck, not both.** Your note says "the two
calls", but the create path you cite as already correct keeps
`_reject_credential_drift_since_verify` *ahead* of the read — and I think that
is right rather than an inconsistency. It licenses `current_creds`, which comes
from a different read entirely; the identity recheck is the one that licenses
the config bytes. Moving both would separate a guard from the thing it guards
and would make the two paths differ again, in the other direction.

The two paths are now identical: credential recheck → read → identity recheck.
Say the word if you want both moved and I will.

I agree the residual past that point is the pre-existing race and that a
critical section over revalidate-plus-write is follow-up-shaped rather than a
merge condition.

`2116 passed, 3 skipped`.

realiti4:
Thank you for the adjustments and the PR, merged it.

PR #225, opened 2026-08-07
fix(session): treat auth-status probe timeout as indeterminate, not invalid

### LINKED ISSUE #224: Session validation treats a subprocess timeout as "not logged in" and deletes the profile under load
**Version:** claude-swap 0.24.1 (uv tool install), macOS, claude CLI 2.1.224

## Symptom

I run a fleet of concurrent `claude` sessions driven by an orchestrator that spawns turns via `cswap run <email> --share-history -- ...`. Under high system load, spawns intermittently fail with:

```
Error: Session profile for Account-2 (<redacted>@example.com) failed validation. Log in with that account and re-add it: cswap --add-account --slot 2
```

The account's login is actually fine: `cswap list` shows fresh usage for the same slot at the same time, and a few minutes later `cswap run` on that account works again with no human action. Re-adding the account is never actually needed. I've hit this in bursts (~96 times over 3 days), always at moments the machine was busy.

## Root cause

`SessionManager._is_session_valid` (session.py) probes the profile with `claude auth status --json` under a 10s timeout (`_AUTH_STATUS_TIMEOUT = 10.0`) and collapses timeouts into "invalid":

```python
except (OSError, subprocess.TimeoutExpired):
    return False
```

`claude` is a Node CLI with a nontrivial cold start; on a loaded box (load avg > cores, many concurrent claude processes) 10s is easy to blow. When that happens, `setup_session` escalates: fast-path check fails -> re-check under the lock fails -> `_bootstrap` re-seeds the profile -> final `_is_session_valid` times out again -> `_cleanup_failed_session` **deletes the session profile** (`shutil.rmtree` + keychain entry) and raises the error above.

So three timeouts in a row turn a busy CPU into a destroyed profile and a misleading "log in again" message. It self-heals only because the next quiet `cswap run` re-bootstraps from backup storage -- which also masks the bug.

## Repro sketch

1. Add an account, confirm `cswap run <email> -- --version` works.
2. Saturate the machine (e.g. `yes > /dev/null &` x cores, or just run ~10 concurrent claude sessions) so `claude auth status` takes >10s.
3. `cswap run <email> -- --version` -> "failed validation", profile dir gone.
4. Let load drop, run again -> works (re-bootstrapped), no re-login ever needed.

## Suggested fix

The validation docstring already frames the check as best-effort ("a revoked but unexpired token still passes and fails on first real use"), so indeterminate should lean valid:

```python
except subprocess.TimeoutExpired:
    return True   # indeterminate: machine under load, not a bad profile
except OSError:
    return False
```

I've been running exactly this as a local patch and the failures are gone. Alternatives if you'd rather not lean valid: raise/make configurable `_AUTH_STATUS_TIMEOUT`, retry the probe once, or at minimum skip `_cleanup_failed_session` when the failure was a timeout so the profile survives and the error can say "try again" instead of "re-add the account".

PR incoming with the lean-valid change.

jhutchins:
Fixes #224.

## What

`_is_session_valid` collapsed `subprocess.TimeoutExpired` into "not logged in". `claude` is a Node CLI with a nontrivial cold start, so on a loaded machine the 10s probe times out, and `setup_session` escalates a False from there all the way to `_cleanup_failed_session` deleting the profile — with an error telling the user to re-add an account whose login was never bad. Details and repro in #224.

## Change

Split the except clause: `TimeoutExpired` now leans valid (the check is best-effort by design — a revoked but unexpired token already passes and fails on first real use), `OSError` still fails closed. Two tests pin the distinction.

## Testing

- `tests/test_session.py`: 155 passed (includes the two new tests).
- Full suite: 1824+ passed; the only persistent failure (`test_move_strict_clear_fails_closed_on_unreadable_dir`) fails identically on unmodified `main` in my environment, so it's unrelated.
- Running this change as a local patch against v0.24.1 in the environment that surfaced the bug (orchestrator spawning many concurrent `cswap run` turns): the bursts of spurious "failed validation" errors stopped.

https://claude.ai/code/session_013RFFYYRMjVjvHhDLrVrde4

### DIFF
--- a/src/claude_swap/session.py
+++ b/src/claude_swap/session.py
@@ -654,7 +654,13 @@ class SessionManager:
                 text=True,
                 timeout=_AUTH_STATUS_TIMEOUT,
             )
-        except (OSError, subprocess.TimeoutExpired):
+        except subprocess.TimeoutExpired:
+            # A timeout is the machine under load, not a bad profile: this
+            # check is best-effort (see docstring), and setup_session
+            # escalates a False all the way to deleting the profile. Real
+            # invalidity still fails on first use.
+            return True
+        except OSError:
             return False
         if result.returncode != 0:
             return False
--- a/tests/test_session.py
+++ b/tests/test_session.py
@@ -544,6 +544,30 @@ class TestIsSessionValid:
             tmp_path / "missing", ACCOUNT_EMAIL, ORG_UUID
         )
 
+    def test_probe_timeout_leans_valid(self, manager, tmp_path, monkeypatch):
+        """A probe timeout is a busy machine, not a bad login.
+
+        setup_session escalates a False from here all the way to
+        _cleanup_failed_session deleting the profile, so an indeterminate
+        probe must not report invalid (#224).
+        """
+        tmp_path.mkdir(exist_ok=True)
+
+        def raise_timeout(*a, **k):
+            raise session_mod.subprocess.TimeoutExpired(cmd="claude", timeout=10)
+
+        monkeypatch.setattr(session_mod.subprocess, "run", raise_timeout)
+        assert manager._is_session_valid(tmp_path, ACCOUNT_EMAIL, ORG_UUID)
+
+    def test_probe_oserror_stays_invalid(self, manager, tmp_path, monkeypatch):
+        tmp_path.mkdir(exist_ok=True)
+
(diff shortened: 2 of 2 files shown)


realiti4:
Thanks for the fix! Merged it.

PR #230, opened 2026-08-08
feat(desktop): add account profile switching

mbelinky:
## Why

Claude Desktop keeps Claude Code session metadata locally, but its session list is scoped to the currently logged-in account. Switching accounts can therefore hide existing sessions even though their transcripts remain on disk.

## What this adds

- A macOS-only `cswap desktop <account>` command.
- One isolated Claude Desktop profile per managed account.
- Reconciliation of missing, resumable Claude Code sessions into the selected profile.
- Explicit confirmation when an account needs its first Desktop login.
- Safe Desktop shutdown and rollback if profile activation fails.

## Scope

This does not transfer Desktop credentials, modify transcripts, or merge Claude Chat or Cowork history. It manages local Desktop profiles and their Claude Code session indexes.

## Validation

- `uv run pytest tests/test_desktop.py tests/test_cli.py`
- 150 tests passed.
- Full macOS suite: 1,857 passed and 3 skipped. The one failure reproduces unchanged on `main`.

## Possible follow-ups

I have two related follow-ups prepared, but I am holding them until this direction is accepted:

1. Show each account's Claude Code authentication methods clearly: browser OAuth, setup token, or API key.
2. Group multiple authentication methods and the local Desktop profile under one account, while keeping usage at the account level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,40 @@ Sessions use your normal `~/.claude` setup (settings, CLAUDE.md, skills, MCP ser
 
 </details>
 
+### Switch Claude Desktop accounts (macOS)
+
+Switch the account used by Claude Desktop while keeping every resumable local
+Claude Code session visible in the Desktop sidebar:
+
+```bash
+cswap desktop 2
+cswap desktop user@example.com
+cswap desktop 2 --dry-run
+```
+
+Claude Desktop's web login is separate from Claude Code credentials. The command
+therefore keeps one isolated Desktop profile per account and shares only the
+local Claude Code session index between them. The first switch to an account
+opens Claude's normal login flow once. Later switches reuse that authenticated
+profile. After visually verifying the email on that first login, record it once:
+
+```bash
+cswap desktop 2 --confirm-login
+```
+
+The command closes Claude Desktop, switches credentials through the normal
+cswap path, copies only missing resumable index entries, and launches the target
+profile. Existing entries and transcripts under `~/.claude/projects/` are never
+changed. It refuses to close Desktop while a local task is active. The explicit
+first-login confirmation avoids depending on Claude's private login-storage
+format, which changes between Desktop releases.
+The Mac must be unlocked before a relaunch so encrypted login storage remains
+available. Claude Chat and remote Cowork history remain account-specific.
+
+Run the command outside any manually selected Claude profile;
+`CLAUDE_CONFIG_DIR`, `CLAUDE_SECURESTORAGE_CONFIG_DIR`, and
+`CLAUDE_USER_DATA_DIR` are rejected because cswap owns profile selection.
+
 <details>
 <summary>Map accounts to directories — auto-pick per repo</summary>
 
@@ -175,6 +209,7 @@ This will update the stored credentials without creating a duplicate.
 
 ```bash
 cswap run 2                     # Run an account in this terminal only (session mode)
+cswap desktop 2                 # Switch Claude Desktop accounts (macOS)
 cswap auto                      # Auto-switch when nearing rate limits (see above)
 cswap config                    # Show or edit settings (see Configuration below)
 cswap list                      # Show all accounts with 5h/7d usage and reset times
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -217,6 +217,87 @@ Examples:
         sys.exit(130)
 
 
+def _desktop_command(argv: list[str]) -> None:
+    """Handle ``cswap desktop NUM|EMAIL [options]``."""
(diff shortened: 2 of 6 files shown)


thalesfsp:
Not a review, just one finding from working the same area that may save you some code.

Claude.app accepts Chromium's `--user-data-dir` — the shipping app passes that switch to itself, visible in its own process command line — and it does not call `requestSingleInstanceLock`; I get 0 occurrences across `app.asar` on 1.30096.1. So profile selection can be done by launching with a different `--user-data-dir`, without quitting the app or symlinking a directory into place. On macOS 26.6 I have had three instances on three profiles running at once for about an hour.

If that holds on your side, the shutdown-and-rollback path in this PR may be avoidable, and the same mechanism gives concurrent windows rather than one at a time.

I have opened #248 for the concurrent-launch case. It is deliberately in `launcher.py` under the verb `launch`, not `desktop.py`, so it does not collide with this PR — the session reconciliation you do here is genuinely separate work that mine does not cover. If the maintainer would rather have a single desktop command, I am happy to rebase mine onto yours instead.

Two things I have not verified: any of this on Windows or Linux, and whether the session-index reconciliation you added interacts badly with several profiles being live at the same time.

tares003:
Does it work with desktop version?

bitworks-io:
**Heads-up: the `--env CLAUDE_USER_DATA_DIR=…` launch no longer isolates the profile on current Claude Desktop (tested on v1.52386.6). Switching to the `--user-data-dir` command-line switch fixes it.**

I was trying this PR out on macOS and the managed profiles weren't isolating — every launch kept writing to the default `~/Library/Application Support/Claude` profile. Root cause is on the app side, and it's a deliberate hardening that landed after this PR was written:

**What the app does now.** In `Claude.app/Contents/Resources/app.asar`, the main-process init strips the env var on packaged builds unless a signed CDP token is present:

```js
var DY = bW();
app.isPackaged && !DY && (delete process.env.CLAUDE_USER_DATA_DIR, delete process.env.SSLKEYLOGFILE, …);
// …later…
if (process.env.CLAUDE_USER_DATA_DIR) { app.setPath("userData", process.env.CLAUDE_USER_DATA_DIR); }

function bW(){
  let e = process.env.CLAUDE_CDP_AUTH, t = process.env.CLAUDE_USER_DATA_DIR;
  if (!e || !t) return false;                          // no signed token ⇒ false
  let r = e.split("."); if (r.length !== 3) return …;  // header.payload.signature
  if (Buffer.from(r[1],"base64").toString() !== t) return …("userData path mismatch");
  … // signature verification against a trusted key
}
```

So on a normal packaged launch there's no `CLAUDE_CDP_AUTH`, `bW()` is false, and the app **deletes `CLAUDE_USER_DATA_DIR` before** the `setPath("userData", …)` relocation reads it. The profile never moves.

**Empirical confirmation (v1.52386.6):**

- `open -n -b com.anthropic.claudefordesktop --env CLAUDE_USER_DATA_DIR=/tmp/probe` → the var *is* in the launched process env (`ps eww`), but the profile dir stays empty and the app uses the default profile. (The env still shows in `ps` because macOS records the launch environment; the app deletes it in-process.)
- `open -n -b com.anthropic.claudefordesktop --args --user-data-dir=/tmp/probe` → `/tmp/probe` is populated with a full Electron profile (`config.json`, `IndexedDB`, `Cookies`, `Local Storage`) in ~1s, and the new process carries `--user-data-dir=…` on its command line. The Chromium/Electron **command-line switch is not gated** by the CDP-auth check.

**Suggested fix** — use the command-line switch instead of the env var. Two call sites (write + read):

```diff
--- a/src/claude_swap/desktop.py
+++ b/src/claude_swap/desktop.py
@@ _launch_desktop
         if not profile.is_default:
-            command.extend(["--env", f"CLAUDE_USER_DATA_DIR={profile.path}"])
+            command.extend(["--args", f"--user-data-dir={profile.path}"])
@@ _running_profile_dir  # reads back which profile a running instance uses, via `ps`
-            match = re.search(r"(?:^|\s)CLAUDE_USER_DATA_DIR=(\S+)", result.stdout)
+            match = re.search(r"(?:^|\s)--user-data-dir=(\S+)", result.stdout)
```

(macOS `open` forwards everything after a single `--args` to the app, so the flag needs to come after `--args`, not `--env`.) With that change the isolated profiles, the shared `claude-code-sessions` symlink, and session reconciliation all work as designed. Happy to help validate further if useful.

*(Separately, the branch is ~57 commits behind `main` now (0.25.0b1 base vs 0.27.0b1) — a rebase before merge would avoid regressing the release.)*

PR #231, opened 2026-08-08
feat(auth): show credential capabilities per account

mbelinky:
## Summary

- Classify each saved CLI credential as browser OAuth, setup token, or API key.
- Show its scope, Remote Control support, and setup-token expiry in the CLI list, TUI, and JSON output.
- Keep usage at the account level.

## Why

Remote Control requires full-scope browser OAuth. Setup tokens and API keys are inference-only, so the credential method alone does not tell users whether a saved login can run the requested workflow.

## Stack

This builds on #230. It remains closed until that foundational PR lands.

## Validation

- 1,854 tests passed and 3 skipped, excluding five macOS locked-console cases and one pre-existing filesystem permission case.
- Focused auth, JSON, and TUI suite: 151 passed.
- P0 autoreview: clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,7 +333,8 @@ cswap switch 2 --json
   "schemaVersion": 1,
   "activeAccountNumber": 2,
   "accounts": [
-    { "number": 2, "email": "you@example.com", "active": true, "usageStatus": "ok",
+    { "number": 2, "email": "you@example.com", "active": true,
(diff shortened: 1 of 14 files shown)


mbelinky:
Superseded by #235, which combines the credential-capability and account-grouping work in one auth-only PR based on current main.

PR #232, opened 2026-08-08
feat(auth): group login methods by account

mbelinky:
## Summary

- Store multiple CLI login methods under one account instead of duplicating the account.
- Show usage once per account and show each saved login with its own scope, Remote Control support, and expiry.
- Preserve the current default when another login method is added.
- Add explicit method selection and `--require-capability remote-control`, which selects browser OAuth or fails before changing credentials.

## Why

One Claude account can have browser OAuth, setup-token, and API-key credentials. They share account usage but have different capabilities and lifecycles, so the account should own several selectable login methods.

## Stack

This builds on #230 and #231. It remains closed until those smaller foundational changes land.

## Validation

- 1,865 tests passed and 3 skipped, excluding five macOS locked-console cases and one pre-existing filesystem permission case.
- Focused auth, CLI, JSON, and TUI suite: 282 passed.
- P0 autoreview: clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,7 +333,8 @@ cswap switch 2 --json
   "schemaVersion": 1,
   "activeAccountNumber": 2,
   "accounts": [
-    { "number": 2, "email": "you@example.com", "active": true, "usageStatus": "ok",
+    { "number": 2, "email": "you@example.com", "active": true,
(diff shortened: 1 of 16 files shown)


mbelinky:
Superseded by #235, which combines the credential-capability and account-grouping work in one auth-only PR based on current main.

PR #237, opened 2026-08-10
fix(credentials): stop pairing one profile's identity with another profile's credential

jamesvillarrubia:
## The bug

The active-credential read resolves its Keychain item from a fixed service name. The identity read one layer up (`paths.get_claude_config_home`) honors `CLAUDE_CONFIG_DIR`. Under a custom profile those two land in different profiles, so the store reports one account's identity paired with another account's token.

Nothing surfaces. There is no error, no warning, and no mismatch anyone can see, because each half is individually correct.

## Repro

1. Log in normally, so the default profile holds account A.
2. `export CLAUDE_CONFIG_DIR=~/some-other-profile` and log in there as account B.
3. Run `cswap list`.

Before this PR the identity reads B from the custom profile's `.claude.json`, while the credential reads A out of the unsuffixed Keychain item. The token belongs to A. The label says B.

## Who this affects

Users who set `CLAUDE_CONFIG_DIR` themselves. Claude Code does not set it (checked against 2.1.227); the spawn path only propagates it when it is already set.

Affected consumers are the ones that go through `_read_active_credentials`: `list`, `--status`, usage collection, `token_dead`, and switch reads.

**Not** `cswap add`. That consequence was already fixed by #190 (`9c6fc24`, plus `253877c` and `521075f`), which routed capture through `_read_capture_credentials`. All three are ancestors of this branch, so the mislabeled-slot outcome no longer reproduces on it.

## The fix

Resolve the Keychain item the way claude resolves it for the same environment, rather than assuming the default profile's.

Resolution mirrors `_read_capture_credentials` exactly, so the read side and the capture side cannot disagree about which profile they are looking at:

| Environment | Item read |
|---|---|
| Neither variable set | Unsuffixed `Claude Code-credentials` |
| `CLAUDE_CONFIG_DIR` set to a custom profile | That dir's hashed item |
| `CLAUDE_CONFIG_DIR` set to the default profile | Hashed item, then unsuffixed as fallback |
| `CLAUDE_SECURESTORAGE_CONFIG_DIR` defined and empty | Unsuffixed, whatever `CLAUDE_CONFIG_DIR` says |
| `CLAUDE_SECURESTORAGE_CONFIG_DIR` defined and set | That dir's hashed item, no fallback |

Two asymmetries in that table are deliberate:

- An explicit `CLAUDE_CONFIG_DIR` naming the default profile falls back to the unsuffixed item. Claude hashes the exported string and would write a suffixed item, but a user who has always used the default profile may only have the unsuffixed one.
- A defined `CLAUDE_SECURESTORAGE_CONFIG_DIR` gets no fallback. It names the only store claude will read for that environment, so a miss means claude sees a logged-out profile. Reaching into another store would report a credential claude is not using.

`session.keychain_service_name` supplies the derivation, which is the same one `delete_macos_keychain_entry`, `read_session_credentials` and `_read_capture_credentials` already depend on.

### Why redirect and not skip

An earlier version of this PR skipped the Keychain entirely under a custom profile and relied on the plaintext fallback. That is worse, not safer.

On macOS claude writes rotations keychain-only, so a live custom profile frequently has a hashed item and no `.credentials.json` at all. Measured on two such profiles: hashed item present, plaintext file absent in both. Skipping would render a logged-in profile as "no credentials", trading a wrong answer for a missing one.

### What is gated rather than redirected

The managed-key Keychain read (`Claude Code`) is skipped under a custom profile instead of redirected. `session.keychain_service_name` covers the credentials item; claude's managed-key service name under a custom profile is not pinned anywhere in this repo, so redirecting it would be a guess. Capture already reached the same conclusion about the same item.

`primaryApiKey` is read from the active profile's own config and is unaffected, so a custom profile with a managed key still resolves.

## Known scope limit

This changes reads only. Flows that both read and write (the usage refresh, switch) now read the profile's own item but still write the fixed-name item via `_write_oauth_credentials`. That split does not exist before this PR, since both sides previously hit the same item.

Writes are deliberately out of scope: the write-side clobber predates this PR and the provenance gates from #117 and #196 cap the damage. Worth fixing separately.

## Behavior when nothing is set

Unchanged. With no `CLAUDE_CONFIG_DIR` the unsuffixed item is this profile's credential and is still what gets read, on the same code path as before. The retry, timeout and degraded/unavailable classification in `_read_active_oauth_keychain` are untouched; the per-item retry moved into `_read_one_oauth_keychain` and the outer method keeps its signature.

## Tests

7 tests in `tests/test_credentials.py`, none of which touch a real Keychain, so they run on every platform.

- A custom profile never returns the default profile's OAuth or managed-key item.
- A custom profile whose credential exists **only** as a hashed Keychain item is read correctly. This is the case a skip would break.
- A custom profile falls back to its own `.credentials.json` when the hashed item is absent (rc 44, claude's own signal to read the file).
- No `CLAUDE_CONFIG_DIR` still reads the unsuffixed item.
- `CLAUDE_CONFIG_DIR` pointing at the default profile tries hashed, then unsuffixed.
- `CLAUDE_SECURESTORAGE_CONFIG_DIR` defined-but-empty selects the default store even under a custom `CLAUDE_CONFIG_DIR`.
- `CLAUDE_SECURESTORAGE_CONFIG_DIR` defined and set selects that store's hashed item and does not fall back.

Full suite on the rebased tree: 2085 passed, 3 skipped.

Branch is rebased onto current `main` (`9f62506`, post #196).

### DIFF
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -38,10 +38,74 @@ from claude_swap.models import Platform
 from claude_swap.paths import (
     get_claude_config_home,
     get_credentials_path,
+    get_default_claude_config_home,
     get_global_config_path,
 )
 
 _logger = logging.getLogger("claude-swap")
+
+
+def _active_profile_is_default() -> bool:
+    """Whether the active config home is the default profile's.
+
+    ``CLAUDE_CONFIG_DIR`` pointed at the default profile is still the default
+    profile, so this keys on where the path resolves rather than on the
+    variable being set. Resolution also collapses symlinks, which is how a
+    profile reached through one shows up as itself.
+
+    Unresolvable paths answer False: treating an unknown profile as the default
+    is what licenses reading another account's credential, and that is the
+    failure this guards.
+    """
+    try:
+        return (
+            get_claude_config_home().resolve()
+            == get_default_claude_config_home().resolve()
+        )
+    except Exception:
+        return False
+
+
+def _active_oauth_keychain_services() -> list[str]:
+    """Keychain services holding the OAuth credential for the active environment.
+
+    Resolved exactly the way :meth:`ClaudeAccountSwitcher._read_capture_credentials`
+    resolves it, so the active read and the capture read agree about which
+    profile's store they are looking at. Claude (2.1.220
+    ``getMacOsKeychainStorageServiceName``) sources secure storage from
+    ``CLAUDE_SECURESTORAGE_CONFIG_DIR`` when that is *defined*, else
+    ``CLAUDE_CONFIG_DIR``; defined-but-empty selects the *default* secure store,
+    whose item is the unsuffixed one.
+
+    Returned in try-order rather than as a single name for one case: an explicit
+    ``CLAUDE_CONFIG_DIR`` naming the default profile. Claude hashes the exported
+    string, so it would write a *suffixed* item there, but a user who has always
+    used the default profile may only have the unsuffixed one. Capture handles
+    that with the same fallback (``_same_directory`` → ``_read_credentials``).
+    A defined ``CLAUDE_SECURESTORAGE_CONFIG_DIR`` gets no fallback: it names the
+    only store claude will read for this environment, so a miss means claude
+    sees a logged-out profile and reaching into another store would report a
+    credential claude is not using.
+    """
+    # Local import: session imports from this module, so a top-level import
+    # would close the cycle. _read_capture_credentials does the same.
+    from claude_swap.session import keychain_service_name
+
+    secure_env = os.environ.get("CLAUDE_SECURESTORAGE_CONFIG_DIR")
+    if secure_env is not None:
+        if not secure_env:
+            return [CLAUDE_CODE_KEYCHAIN_SERVICE]
+        return [keychain_service_name(secure_env)]
+
+    config_dir = os.environ.get("CLAUDE_CONFIG_DIR")
+    if not config_dir:
+        return [CLAUDE_CODE_KEYCHAIN_SERVICE]
+
+    services = [keychain_service_name(config_dir)]
+    if _active_profile_is_default():
+        services.append(CLAUDE_CODE_KEYCHAIN_SERVICE)
+    return services
+
 
 # Service name for per-account backup credentials now managed via the ``security``
 # CLI on macOS. Deliberately distinct from KEYRING_SERVICE so old keyring items and
@@ -446,7 +510,31 @@ class CredentialStore:
         return self._read_active_credentials().value
 
     def _read_active_oauth_keychain(self) -> tuple[str | None, bool]:
-        """Read the active OAuth Keychain item with a bounded retry.
+        """Read the active profile's OAuth Keychain item(s).
+
+        Reads the item(s) :func:`_active_oauth_keychain_services` resolves for
+        the current environment, in order, stopping at the first hit. Only the
+        explicit-``CLAUDE_CONFIG_DIR``-names-the-default-profile case yields
+        more than one; see that function for why.
+
+        Returns ``(value, failed)`` exactly as before: ``value`` is the
+        credential string, or ``None`` when every item was absent (rc-44) or the
+        Keychain was unreadable. ``failed`` is True only for unreadable.
+
+        An unreadable Keychain stops the walk. It is a property of the Keychain,
+        not of the item, so a second service name cannot fare better — and
+        ``_kc_call`` has already flipped routing to file mode by then.
+        """
+        for service in _active_oauth_keychain_services():
+            value, failed = self._read_one_oauth_keychain(service)
+            if value:
+                return value, False
+            if failed:
+                return None, True
+        return None, False
+
+    def _read_one_oauth_keychain(self, service: str) -> tuple[str | None, bool]:
+        """Read one OAuth Keychain item with a bounded retry.
 
         Returns ``(value, failed)``. ``value`` is the credential string, or
         ``None`` when the item is absent (rc-44) or unreadable. ``failed`` is True
@@ -460,7 +548,7 @@ class CredentialStore:
             try:
                 value = self._kc_call(
                     macos_keychain.get_password,
-                    CLAUDE_CODE_KEYCHAIN_SERVICE,
+                    service,
                     macos_keychain.keychain_account_name(),
                 )
                 return value, False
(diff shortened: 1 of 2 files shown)


realiti4:
Hi, thanks a lot for this.

Could we address a couple of things before a merge though? The repo already ships the hashed-name machinery this PR decides not to use, and half of the PR's stated motivation was fixed by #190 before this branch was cut.

On the alternative you asked about: I'd rather not pin the identity read to the default profile. #190 already committed `cswap add` to following `CLAUDE_CONFIG_DIR`, and reversing that now would break its design. I think we should keep your guard, but redirect the read through the existing resolver rather than skipping the keychain entirely.

Detailed findings below are from my review pass with Claude Code:

---

**1. The `cswap add` consequence is already fixed — including on this PR's own base.** 9c6fc24 (PR #190, merged 2026-07-31, plus follow-ups 253877c and 521075f, all contained in this branch's merge-base) routed capture through `_read_capture_credentials` (`switcher.py:3239`, `switcher.py:3348`), which reads the custom profile's *hashed* keychain item. So "add stores the default profile's credential in a mislabeled slot" no longer reproduces. The remaining affected consumer set — where this PR prevents the cross-profile read, though it does not yet read the custom profile's real credential (see 2) — is `list`, `--status`, usage collection, `token_dead`, and switch reads through `_read_active_credentials`.

**2. Skip → redirect: the hashed name is not a guess in this codebase.** `session.keychain_service_name()` (`session.py:232`) codifies the derivation (sha256 of the raw env value, NFC-normalized, hashed verbatim — pinned against claude's source), and `read_config_dir_credentials()` (`session.py:339`) implements hashed-item-then-file for an arbitrary config dir, including the rc-44 "absent item is claude's own signal to read the file" semantics. To be precise about the ask: that function can't be dropped in unchanged — with `strict_keychain=False` it falls back to plaintext on a keychain *error* too and can't mark the result degraded, and with `strict_keychain=True` it raises instead of returning a classified result. So the ask is to reuse/factor that resolution (hashed service name + absent-item semantics) into `_read_active_credentials`, keeping its own `_kc_call` retry/timeout and degraded/unavailable classification. That buys two things the skip can't:

- `cswap list` under a custom profile shows the profile's **real** credential and usage. With the skip, macOS custom profiles go mostly blind: claude writes keychain-only there, so the plaintext fallback usually doesn't exist and a logged-in profile renders as "no credentials".
- No un-degraded stale seed. The config-dir plaintext file can be a superseded generation while claude rotates the hashed item; the skip returns those bytes as the live credential with `degraded=False`, while the redirect only reaches the file on the absent-item signal.

**3. The predicate misses `CLAUDE_SECURESTORAGE_CONFIG_DIR`.** Claude 2.1.220+ resolves secure storage from that variable with precedence when it is defined (see the `_read_capture_credentials` docstring, `switcher.py:657`, written against that source). `_active_profile_is_default()` consults only `CLAUDE_CONFIG_DIR`, so custom config + defined-but-empty secure var — where claude uses the *default* secure store and the unsuffixed read was actually correct — now returns nothing. This dissolves into the same fix: resolve the store the way `_read_capture_credentials` does.

**4. A caution on the read-only scoping — fine to defer, worth a note in the description.** Post-PR, flows that both read and write (the usage refresh, switch) read the file but still write the fixed-name item (`_write_oauth_credentials`, `credentials.py:829`) — a split inside one operation that didn't exist before, since both sides previously hit the same item. The provenance gates (#117/#196) cap the damage and the write-side clobber predates this PR, so keeping writes out of scope as you propose is fine — but the description should say the split exists.

**5. Blast radius.** "Every Claude Code session sets `CLAUDE_CONFIG_DIR`" isn't right — checked on 2.1.227: sessions don't set it, and the spawn path only *propagates* it when already set. The bug fires for users who set the variable themselves. Worth trimming so the severity reads accurately.

**Verified while reviewing:** the branch merges conflict-free onto current main (post-#196); the full suite on the merged tree passes (2080 passed); the two custom-profile tests fail against main's `credentials.py` and pass with the fix; the two custom-profile tests fail against main's `credentials.py` and pass with the fix; all four new tests are portable off macOS (no real keychain probe). CI is green on all three platforms.

jamesvillarrubia:
Thanks, this was a genuinely useful review. You were right on every point that mattered, including the one I had backwards. Pushed a rewrite.

**On skip vs redirect: you're right, and my reasoning was wrong.**

I argued the hashed name was Claude Code's private business and guessing it would break on a scheme change. That argument does not survive contact with this codebase, which already codifies the derivation in `session.keychain_service_name()` and depends on it in the delete, session-read and capture paths. Skipping to avoid a "guess" the repo has already committed to just made this one read inconsistent with the other three.

Worse, I checked the failure mode you described on my own machine. Both of my custom profiles hold a hashed keychain item and **no** `.credentials.json` at all. So the skip would have made `cswap list` report "no credentials" for two profiles that are logged in and working. I traded a wrong answer for a missing one and called it "the honest answer."

The read now resolves through `_active_oauth_keychain_services()`, which mirrors `_read_capture_credentials`'s resolution exactly: `CLAUDE_SECURESTORAGE_CONFIG_DIR` when defined, else `CLAUDE_CONFIG_DIR`, with defined-but-empty selecting the default store. That closes finding 3 as a side effect, as you predicted.

I kept `_read_active_oauth_keychain`'s own `_kc_call` retry and degraded classification rather than calling `read_config_dir_credentials`, for the reasons you gave. The per-item retry moved into `_read_one_oauth_keychain`; the outer method keeps its signature so existing callers and tests are untouched.

**One place I did not follow you, and why.**

The managed-key read is gated rather than redirected. `session.keychain_service_name` covers the credentials item, and I could not find claude's managed-key service name under a custom profile pinned anywhere in this repo. Redirecting it would be the guess I wrongly accused the OAuth half of being. Gating also matches what capture already concluded about that exact item: "never the unsuffixed 'Claude Code' Keychain item, which belongs to the default profile and would answer for a login that is not the one being added." `primaryApiKey` still answers from the active profile's own config, so a custom profile with a managed key is not blinded. Happy to redirect instead if you have a source reference for the derivation.

**Findings 1, 4 and 5.**

Fixed in the description, which I've rewritten.

1. Confirmed `9c6fc24`, `253877c` and `521075f` are all ancestors of this branch, so the `cswap add` consequence was already fixed before it was cut. The motivation now names only the consumers that actually still reproduce: `list`, `--status`, usage collection, `token_dead`, and switch reads.

4. The read/write split is now stated explicitly, with the note that writes stay out of scope and the provenance gates cap the damage.

5. You're right and I should have caught this. Claude Code does not set `CLAUDE_CONFIG_DIR`. It is set in my sessions because my own launcher sets it, and I generalized from my machine to everyone's. Severity now reads: users who set the variable themselves.

**Also rebased onto current main** (`9f62506`, post `#196`), so the branch is written against `_active_read_failed` and the degraded-travel behavior rather than assuming the pre-`#196` shape.

**Verification**

- Full suite on the rebased tree: 2085 passed, 3 skipped.
- 3 failures on my machine are environment-specific and reproduce identically on a clean `main` checkout: `test_real_store_guard.py::test_c0_...`, and the `TestMoveUnreadableSourceIsNotAbsent` / `TestSwapUnreadableSourceIsNotAbsent` pair.
- 7 tests in `test_credentials.py`, up from 4. Two are new and cover exactly the regressions you identified: a custom profile whose credential exists **only** as a hashed keychain item, and a defined-but-empty `CLAUDE_SECURESTORAGE_CONFIG_DIR` under a custom `CLAUDE_CONFIG_DIR`. Both fail against the previous version of this branch.
- No real keychain probe in any of them, so they stay portable off macOS.
- `ruff check` clean. I did not run `ruff format`, since `main` is not currently format-clean and it would have buried the diff.

realiti4:
Thanks, this looks great, merged it.

On the managed-key derivation you asked about: Claude's `getApiKeyFromConfigOrMacOSKeychain` / `saveApiKey` call `getMacOsKeychainStorageServiceName()` with no suffix, same hash, so the custom-profile item is `Claude Code-<sha256(raw env)[:8]>`.

Gating is fine for now though, the redirect can always be handled separately later if you feel like taking it on.

PR #238, opened 2026-08-11
fix(settings): make the JSON write byte-stable — trailing newline + sorted keys

codeslake:
Two properties of the settings writer reach disk, and both make a
tracked file flap.

`atomic_write_json` wrote `json.dumps(data, indent=2)`:

- **no trailing newline** — a permanent `\ No newline at end of file`
  diff that returns after every write, so committing it never sticks;
- **key order taken from the dict it was handed** — the same content
  serialises differently depending on insertion order.

That is invisible if `settings.json` is a plain local file. It is not
invisible when the deployment links it into a dotfiles repo (`link:
absolute` here), which is the normal setup for anyone syncing config
across machines: every `cswap config set` / theme change then rewrites a
tracked file.

Measured on three machines: `git status` was never clean, and separately
a cross-machine sync was reverted with **no value changed** — `json.load`
equality stayed `True` while the `ui`/`theme` block had simply moved.

Sorting is recursive, which is the point: `ui`/`theme` is a nested
block, and top-level-only sorting would leave exactly the observed case
still flapping.

The writer is shared with the autoswitch state file and the other
machine-local JSON, so those get the same stability.

## Tests

Four, each written RED first, and both halves mutation-checked —
dropping `+ "\n"` kills the newline test, dropping `sort_keys` kills
both ordering tests:

- write ends with a newline
- reordered input produces identical bytes
- nested ordering normalises too
- content still round-trips (normalising the bytes must not change what
  is stored)

Suite: 1837 passed, 3 skipped.

### DIFF
--- a/src/claude_swap/settings.py
+++ b/src/claude_swap/settings.py
@@ -472,7 +472,16 @@ def atomic_write_json(path: Path, data: dict) -> None:
         os.chmod(path.parent, 0o700)
     fd, tmp_path = tempfile.mkstemp(dir=str(target.parent), suffix=".tmp")
     try:
-        os.write(fd, json.dumps(data, indent=2).encode("utf-8"))
+        # sort_keys + trailing newline: the bytes must depend on the CONTENT
+        # only. This file is deployed `link: absolute`, so every write lands
+        # in a TRACKED dotfiles path — without both, the repo carries a diff
+        # that returns after each write and `git status` is never clean on any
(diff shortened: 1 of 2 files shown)


codeslake:
Closing — this belongs on one of my existing PRs, not a separate one. Reopening as part of #199, which owns the TUI theme write that is one of the three paths this fixes.

PR #242, opened 2026-08-12
feat(chrome): optional Claude in Chrome extension account sync (macOS)

fantasywind:
## What

Adds an **opt-in, macOS-only** feature that switches the *Claude in Chrome*
extension's account alongside the CLI account. When enabled, `cswap switch` (and
the menu bar) also flips the account the extension is paired to, so Claude Code's
browser automation follows the same account.

Disabled by default — when off, cswap behaves exactly as before (CLI-only), and
non-macOS platforms no-op.

## How it works (OAuth tokens, not a cookie)

The extension authenticates with **OAuth tokens** — `accessToken`
(`sk-ant-oat01-`), `refreshToken` (`sk-ant-ort01-`), `tokenExpiry` — held in its
service worker's **in-memory** `chrome.storage.session`. `accountUuid` lives in
`chrome.storage.local`. The service worker reads these once at startup and
connects the bridge from there.

Switching an account therefore means:

1. Inject the target account's tokens into the SW's storage (via CDP, evaluating
   in the service worker).
2. Terminate the SW with CDP `Target.closeTarget` — **not**
   `chrome.runtime.reload()`. A runtime reload *clears* `storage.session`
   (losing the just-injected tokens); `closeTarget` *preserves* it.
3. Wake the SW with a throwaway background `claude.ai` tab (closed afterwards),
   so it restarts, reads the injected tokens, and reconnects as the new account.
   If the injected access token is expired, the SW auto-refreshes via the refresh
   token on adoption (self-heals).

Because the tokens live in in-memory session storage, capture requires the
automation Chrome running + CDP (there is no on-disk/offline path, and — unlike a
cookie approach — no Keychain/openssl/SQLite is involved at all).

## Why a dedicated Chrome profile

CDP against modern Chrome (macOS ~M136+/2025) forces a dedicated, separately
launched profile:

1. The **default user-data-dir ignores `--remote-debugging-port`** → switching
   must run in a *separate* `--user-data-dir` instance.
2. **`--load-extension` is blocked** → the Claude extension is installed once
   from the Web Store into the dedicated profile.
3. The wake tab hits claude.ai, which **Cloudflare blocks headless** → the
   automation Chrome runs headed.

Full rationale lives in the `chrome_session` module docstring.

## Token lifetimes & keep-alive

- `accessToken` ~8h; the extension refreshes it transparently.
- `refreshToken` ~28.5 days, **rotating on every refresh**. This is the real
  parking limit: switching to an account within ~28.5 days is seamless; past
  that its refresh token expires and it needs a browser re-login.
- `cswap chrome refresh <n>` (and every switch) refreshes a stored token offline
  via `platform.claude.com/v1/oauth/token`, rolling the window forward so a
  parked account never lapses — no browser needed.

## Commands

- `cswap chrome setup` — guided one-time setup (create profile → install
  extension → connect + capture each account)
- `cswap chrome doctor` — setup state + per-account coverage
- `cswap chrome open [--ensure] [--focus]` / `capture <n>` / `refresh <n>`
  / `create-profile [--rebuild]` / `enable` / `disable`
- `cswap chrome bridge install | status | uninstall | run` — local relay so a
  *running* session follows a switch (see below)

A `(chrome)` marker in `list` / `status` / the menu bar flags accounts that have
stored tokens.

## Making a *running* session follow a switch (bridge relay)

The token swap above re-pairs the **extension**, but a Claude Code session binds
its own bridge account at process start — so a switch made *while it is running*
was only picked up after restarting it.

The extension and Claude Code pair through Anthropic's bridge at
`wss://bridge.claudeusercontent.com/chrome/<accountUuid>`; the UUID is the
**room**, and both peers must resolve to the same one to see each other. Claude
Code ships an escape hatch: with `LOCAL_BRIDGE=1` it connects to a plaintext
`ws://localhost:8765/chrome/dev_user_local` with a tokenless connect —
account-agnostic.

`chrome_bridge` is the other half: a local relay on that port. It accepts Claude
Code's tokenless local connect, originates an *authenticated* upstream
connection to the real bridge in the **current cswap account's** room, and
relays frames both ways. On `cswap switch` it re-points that upstream to the new
room while the downstream connection stays up — so the browser follows the
switch live, no restart.

```bash
cswap chrome bridge install     # background relay via a LaunchAgent (com.claude-swap.bridge)
cswap chrome bridge status      # running? on which port     (also: run, uninstall)
LOCAL_BRIDGE=1 claude           # route this session through the relay
```

- The **extension is never modified** by this path — it stays on the real
  bridge and keeps auto-updating; only Claude Code's side is re-pointed.
- Verified TLS upstream (system trust store); the relay listens on `127.0.0.1`
  only, and runs independently of the `chrome.enabled` gate (the LaunchAgent
  keeps it up across logins).
- `cswap chrome doctor` reports relay + LaunchAgent state alongside the rest.
- Caveat: `LOCAL_BRIDGE` / `dev_user_local` are internal Claude Code dev
  features and may change between versions — hence opt-in, and harmless when
  unused.

## Design / safety

- **Opt-in**: gated on `chrome.enabled` in `settings.json`; `ChromeSync` is a
  no-op when disabled or on non-macOS. The switch/add hooks are best-effort and
  never fail the OAuth switch that already committed.
- **Freshness**: on every switch the outgoing account's tokens are re-captured
  (the active token rotates ~8h), matched to a vault slot by `accountUuid`.
- **Capture guard**: on `cswap add`, tokens are stored only if the extension's
  org matches the account Claude Code is adding.
- **Security**: stored tokens are bearer credentials (password-equivalent) in
  `chrome_sessions.json`, chmod 600, beside the existing backup root.
- **No new dependencies**: pure-stdlib CDP client; the offline refresh uses the
  system `curl` (as the old path used `openssl`).

## Browser behavior (non-intrusive)

- **No focus stealing**: the window is never raised automatically;
  `cswap chrome open --focus` is opt-in.
- **No lingering claude.ai tab**: the wake tab is a throwaway *background* tab
  that is closed afterwards, leaving the dedicated browser on a normal New Tab
  Page, ready for Claude Code to drive.
- `cswap chrome open` on an already-running browser only ensures it's up (no
  re-inject).

## Testing

- `tests/test_chrome_session.py` (18 tests): token value object, vault
  (incl. `find_by_uuid`), settings load/toggle, disabled/unsupported no-op, the
  add-time org guard, keep-alive staleness, and offline-refresh vault update.
- `tests/test_chrome_bridge.py` (6 tests): WebSocket frame round-trip (masked,
  unmasked, and the 8-byte extended-length path), active-account resolution from
  the sequence file + vault (and the `None` case), the closed-port liveness
  probe, and a stable launchd label.
- Full suite otherwise green; 3 pre-existing, environment-specific failures are
  unrelated to this change (they reproduce unchanged on `main`).
- Verified end-to-end on macOS: A→B→A account flips through the `ChromeSync`
  facade, confirmed by the extension's reported identity and by Claude Code's
  `list_connected_browsers` following the switch.
- Bridge relay verified **without restarting the session**: with `LOCAL_BRIDGE=1`,
  a mid-session `cswap switch` moved `cswap chrome doctor`'s reported identity to
  the other account, and `list_connected_browsers` kept returning the browser on
  a fresh connection in the new room.

## Notes

- macOS-only for now; other platforms no-op with a clear message. The token
  mechanism itself is portable (no macOS-specific crypto); a Linux/Windows launch
  path is a follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,6 +245,50 @@ Shows every account's 5h / 7d / spend usage and switches with a click (specific 
 
 </details>
 
+## Chrome extension sync (macOS)
+
+<details>
+<summary>Optional — also switch the "Claude in Chrome" extension's account when you switch</summary>
+
+`cswap switch` changes the account the `claude` CLI uses. This opt-in add-on also switches the account the **Claude browser extension** uses, so Claude Code's browser automation follows the same account.
+
+Two separate logins are involved, linked by account number:
+
+- **CLI** — the `claude` command's account (Keychain OAuth), managed already.
+- **Browser** — the Claude in Chrome extension's own OAuth pairing.
+
+The extension authenticates with OAuth tokens (not a cookie), held in its
+service worker's in-memory storage. Switching stores each account's tokens and,
+on a switch, injects the target account's tokens into the extension and restarts
+its service worker so it reconnects as that account. Because modern Chrome won't
+let a tool drive your everyday profile, this runs a dedicated, lean "Claude Code"
+Chrome profile alongside your normal browser (a separate window just for Claude
+Code).
+
+> **Security:** stored tokens are bearer credentials (`sk-ant-oat01-` /
+> `sk-ant-ort01-`, equivalent to a password) kept in `chrome_sessions.json`
+> (chmod 600) beside your other cswap state. Anyone with that file can act as
+> those accounts until the tokens expire.
+
+```bash
+cswap chrome setup      # guided: enable → create profile → install extension → capture accounts
+cswap chrome doctor     # what's set up + per-account coverage
+```
+
+The wizard installs the extension into the dedicated profile (one click from the Web Store), has you connect it to Claude Code, and captures each account's tokens. Other commands:
+
+```bash
+cswap chrome open              # open/focus the dedicated Claude Browser
+cswap chrome capture <n>       # store account <n>'s tokens (connect it there first)
+cswap chrome refresh <n>       # keep account <n>'s stored tokens alive (no browser)
+cswap chrome create-profile    # (re)create the dedicated profile  (--rebuild to reset)
+cswap chrome enable | disable  # turn the feature on/off
+```
+
+Once set up, `cswap switch <n>` (CLI or menu bar) flips the dedicated browser to account `<n>` too — the extension reconnects on its own, no side-panel reopen needed. A stored account's refresh token lasts about 28.5 days untouched; switching to it within that window is seamless (the extension self-refreshes an expired token), and `cswap chrome refresh <n>` (or any switch) rolls the window forward so it never lapses. Off until you turn it on (`chrome.enabled` in `settings.json`); when disabled, switching is CLI-only and behaves exactly as before.
+
+</details>
+
 ## Advanced
 
 ### Configuration
--- a/README.md
+++ b/README.md
@@ -287,6 +287,17 @@ cswap chrome enable | disable  # turn the feature on/off
 
 Once set up, `cswap switch <n>` (CLI or menu bar) flips the dedicated browser to account `<n>` too — the extension reconnects on its own, no side-panel reopen needed. A stored account's refresh token lasts about 28.5 days untouched; switching to it within that window is seamless (the extension self-refreshes an expired token), and `cswap chrome refresh <n>` (or any switch) rolls the window forward so it never lapses. Off until you turn it on (`chrome.enabled` in `settings.json`); when disabled, switching is CLI-only and behaves exactly as before.
 
+#### Making a *running* session follow a switch (bridge relay)
+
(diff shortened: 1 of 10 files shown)


deathemperor:
Test-plan correction: I read cswap's `transfer.export_accounts` (claude-swap `src/claude_swap/transfer.py`) — the engine does **not** refuse an existing path; `_atomic_write_file` replaces it. The fork's commit message ("export refuses to overwrite an existing file") is wrong on that point. On the Mac the guard is `NSSavePanel`'s own "Replace?" sheet when an existing file is picked, which is standard and sufficient here; nothing in the Swift code claims otherwise. (`gh pr edit` is refused for this token, hence a comment rather than a body edit.)

PR #245, opened 2026-08-13
auto: authoritative reset override bridge (opt-in, one-directional clamp)

### LINKED ISSUE #244: auto: authoritative reset override to avoid over-wait when usage API reports the next window's reset
## Problem

`cswap auto` derives every window reset time from `/api/oauth/usage`
(`oauth.build_usage_result` → `resets_at`). That endpoint is a poller's only
source, but it isn't always the authoritative current-window reset. We observed
(2026-08) that once a 5h window is fully spent, the usage endpoint can report the
**next** window's `resets_at` instead of the current one.

When that happens and **all** accounts are exhausted, the auto loop's
all-exhausted recovery ordering (`autoswitch._binding_recovery_ts` →
`relevant_windows` `resets_at`) sleeps to that too-late time and idles past the
real reset. In our case the endpoint said 21:30 while the true reset was 19:09 —
~2h20m of avoidable idle.

The **authoritative** current-window reset is the one Claude Code itself prints
in its rate-limit message ("resets 6:50pm") — a value only the worker actually
running Claude Code ever sees, never the usage endpoint.

## Proposed approach

An opt-in bridge: a worker that observes the authoritative reset writes it to a
small JSON file; cswap clamps its usage-derived resets with it.

```json
{ "<account-email>": {"five_hour": "<iso8601>", "seven_day": "<iso8601>"} }
```

Key property — the clamp is **one-directional**: it only ever pulls a window's
reset *earlier*, never later. So a stale or wrong override can shorten an
over-wait but can never extend a wait or hide a real limit. Default off = exact
pre-existing behavior (fail-open on missing/malformed file or entry).

## What I've drafted (PR to follow)

- `authoritative_reset.py` — tolerant reader + pure `clamp_account_resets()`.
- `oauth.try_fetch_usage_for_account()` — opt-in `authoritative_resets` param
  (default `None` = no-op); clamps this account's windows after the fetch.
- 12 tests (earlier-only clamp; missing/elapsed/unparsable/no-override no-ops).

I deliberately kept it to the self-contained core + opt-in param and did **not**
wire a settings key or the `auto` loop file-read yet — happy to add that wherever
you prefer (env var, `AutoSwitchSettings` field, etc.). Would you be open to this
direction?

yunjeongiya:
Closes #244 (proposal — see issue for the problem writeup and open questions).

## What

Opt-in bridge letting a worker that observes the **authoritative** rate-limit
reset (the "resets 6:50pm" text Claude Code prints — never visible to the usage
endpoint) feed it to cswap, so an all-accounts-exhausted `auto` loop doesn't idle
past the real reset when `/api/oauth/usage` returns the next window's `resets_at`.

## How

- `authoritative_reset.py` — tolerant `read_authoritative_resets(path)` + pure
  `clamp_account_resets(usage, overrides, now)`.
- `oauth.try_fetch_usage_for_account(..., authoritative_resets=None)` — when
  given `{email: {window: iso}}`, clamps this account's 5h/7d windows after the
  fetch. Default `None` = exact pre-existing behavior.

**One-directional by design:** the clamp only pulls a reset *earlier*, never
later — a stale/wrong override can shorten an over-wait but can never extend a
wait or hide a real limit. Fail-open on missing/malformed file or entry.

## Scope / open question

Deliberately just the self-contained core + opt-in param — no settings key or
`auto`-loop file-read wired yet, so this changes nothing by default. Happy to add
the wiring (env var / `AutoSwitchSettings` field / read in the auto loop) whichever
way you prefer.

## Tests

`tests/test_authoritative_reset.py` — 12 cases (earlier-only clamp, fills a
missing endpoint reset, ignores elapsed/unparsable/no override, input untouched).
`pytest tests/test_authoritative_reset.py tests/test_oauth.py` → 120 passed.

### DIFF
--- a/src/claude_swap/authoritative_reset.py
+++ b/src/claude_swap/authoritative_reset.py
@@ -1,0 +1,115 @@
+"""Authoritative rate-limit reset overrides — opt-in bridge for worker-observed resets.
+
+cswap derives every window reset time from the ``/api/oauth/usage`` endpoint
+(:func:`oauth.build_usage_result`). A poller has no other source, but the
+*authoritative* current-window reset is the one Claude Code itself prints in its
+rate-limit message ("resets 6:50pm") — a value only the worker actually running
+Claude Code ever observes. When the usage endpoint reports a reset later than the
+true current-window reset (observed 2026-08: once a window is fully spent the
+endpoint can return the *next* window's reset), an all-accounts-exhausted auto
+loop sleeps to that too-late time and idles past the real reset.
+
+This module lets an external worker that DOES see the authoritative reset feed it
+in through a small JSON file. The clamp is deliberately **one-directional**: an
+authoritative reset only ever pulls a window's reset *earlier*, never later — so a
+stale or wrong override can shorten an over-wait but can never extend a wait or
+hide a real limit. Absent file / missing entry / unparsable value = exact
+pre-existing behavior (fail-open).
+
+The caller reads this file and passes the parsed dict to
+:func:`oauth.try_fetch_usage_for_account` as ``authoritative_resets``; this
+module owns only the (tolerant) reader and the pure clamp. File format::
+
+    {
+      "<account-email>": {"five_hour": "<iso8601>", "seven_day": "<iso8601>"},
+      ...
+    }
+
+Any window key may be omitted; unknown keys are ignored.
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+from datetime import datetime, timezone
+
+_logger = logging.getLogger("claude-swap")
+
+# usage-dict window key -> override key (same names; explicit so a rename of one
+# side can't silently desync the other).
+_WINDOW_KEYS = (("five_hour", "five_hour"), ("seven_day", "seven_day"))
+
+
+def read_authoritative_resets(path: str | None) -> dict[str, dict]:
+    """Load the override file as ``{email: {window: iso}}``; ``{}`` on any failure.
+
+    Fail-open by contract: a missing, unreadable, or malformed file must never
+    break a usage fetch — it just means "no overrides this pass". Only the outer
+    shape is validated here; per-window parsing/comparison happens in
+    :func:`clamp_account_resets` so one bad entry can't discard the rest.
+    """
+    if not path:
+        return {}
+    try:
+        with open(path, encoding="utf-8") as f:
+            data = json.load(f)
+    except (OSError, ValueError) as e:
+        _logger.debug("authoritative reset file unreadable (%s): %r", path, e)
+        return {}
+    if not isinstance(data, dict):
+        _logger.debug("authoritative reset file is not an object: %s", path)
+        return {}
+    return {
+        email: windows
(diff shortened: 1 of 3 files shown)


realiti4:
Hey, thanks a lot for this. I'm open to the direction, but before we commit to it I want to make sure the problem still exists at this size on current versions. 

Review pass generated with Claude Code:

---

**Which version was the incident on?** Up to v0.23.0 the planner parked an at-limit account's next poll *at* the reported reset, so a wrong 21:30 meant no polls until 21:30 — exactly your 2h20m. Since v0.24.0 (13fb380, "keep exhausted accounts polling"), exhausted accounts re-poll every ~10 min ([poll_policy.py:256-270](https://github.com/realiti4/claude-swap/blob/main/src/claude_swap/poll_policy.py#L256-L270)) and all-exhausted sleeps are chunked at ≤600s, so a wrong-late `resets_at` costs ~10–30 min at most, not hours.

**After the true 19:09 reset, did the endpoint's `utilization` drop (while `resets_at` stayed wrong), or did it keep reporting ≥100% until 21:30?** This decides whether the clamp helps: candidate selection is utilization-based, so if the endpoint kept saying 100%, correcting the timestamp wouldn't have unblocked anything — the loop would wake at 19:10, see 100%, and stay blocked anyway. If utilization did recover on time, the clamp is real but buys ~10–30 min on current versions, not 2h20m.

Code notes, assuming we proceed:

- The "fills a missing endpoint reset" arm breaks the earlier-only invariant you're selling. On main, an exhausted window with no reset means "recovery unprovable → re-check every 300s"; filling in an override converts that to sleeping toward the override in 600s chunks — so a wrong override on that path can *slow* re-checks. Bounded and small, but I'd drop that arm so "can never extend a wait" holds unconditionally.
- I'd rather not freeze the JSON file format before a producer exists — once released it's a compatibility contract. Related: plaintext email keys (no case normalization) in a file that'll get pasted into issues; worth considering account numbers or at least normalizing.
- For wiring: my preference is a path field on `AutoSwitchSettings`, read once per tick in the auto loop. If the answers above still justify it, I'd like the minimal wiring included in this PR so it lands as one complete opt-in rather than an unused module.

****

yunjeongiya:
Sorry for the month of silence — answering both questions.

**Version**: the incident was on 2026-08-11 (KST), on v0.24.1 — v0.25.0 was released later that same day and we upgraded afterwards. So it was already on the "keep exhausted accounts polling" behaviour, not the pre-0.24 park-at-reset one.

**Did `utilization` drop at 19:09 while `resets_at` stayed 21:30?** We did not capture the endpoint responses during the window, so I can't answer that honestly. Given that, and that on ≥0.24 the cost of a wrong-late `resets_at` is bounded to the ~10-minute re-poll rather than hours, I don't think this PR is justified in its current shape. Closing it and #244; if we see the 100%-until-21:30 variant again I'll come back with the captured responses rather than a clamp built on a guess.

The problem that actually bit us this week is a different one — one `threshold` governing both the 5h and 7d windows — which #355 addresses, and I've left a field report there.

PR #247, opened 2026-08-14
autoswitch: drain v2 — the engine creates the park pause (CON-433)

danpoyar:
Drain v2 (active checkpoint) for the forced proactive switch: instead of waiting for a transcript silence a 10-20 agent fleet never produces (v1 live episodes: 621s and 620s of waiting, swap under load anyway), the engine signals every mid-turn background session to checkpoint and freeze, confirms fixation machine-wise from the `claude agents --json` roster (status busy→idle = turn boundary), swaps, verifies the new account, and wakes the frozen sessions — one cold cache write per agent on the new account.

- Wave channel: one-shot headless `claude -p` herald using SendMessage (a raw daemon socket write is held unread by bypass-mode receivers per the cross-session-messaging inbound rules). Per-terminal `cswap run` sessions excluded by pid — their credentials don't swap.
- Honest count: the switch event carries additive `drain2: {outcome: "ready"|"timeout", waitedSeconds, fixed, forced}`; the cap WARN names the sessions it forces.
- Episode persists in `autoswitch_state.json` (signaled→swapped); the resume wave survives a daemon restart, and a tick-top reconcile machine-closes an episode abandoned by a foreign-trigger switch or a vanished forcing condition (review r1).
- Fallback: any channel failure emits `drain2-unavailable` and falls back to the passive v1 drain (in-process backoff). At-limit/failover keep the passive drain. `drain2WaitSeconds=0` (default) is v1 bit-for-bit — upstream untouched.

Proof: red scenario tests shown failing before the implementation (ticket CON-433, 13:26Z comment); full suite `env -u FORCE_COLOR timeout 600 uv run --group dev pytest -q` → **1747 passed, 3 skipped, 1 xfailed** (all 1718 pre-existing v1 tests green, unmodified). Review: code-reviewer round 1 «принимать нельзя» → fixes in 8aba67f → round 2 «можно принимать», with one condition OUTSIDE the merge: fleet enablement (`cswap config set autoswitch.drain2WaitSeconds 180`) only after a live one-wave channel proof (herald → busy bypass background session: delivered, not held).

Ticket: CON-433 (do not auto-close; Done is the orchestrator's call after fleet acceptance).


danpoyar:
Opened against the wrong base by tooling — this belongs to the danpoyar fork. Apologies for the noise.

PR #248, opened 2026-08-15
feat(launch): run the Claude desktop app per account, several at once (macOS)

thalesfsp:
## Why

`cswap` switches which account Claude Code runs as. This adds the same idea for the desktop app, with one difference that turned out to matter: the accounts run **at the same time**, in separate windows, rather than one at a time.

The use case is having three Max accounts and wanting a Claude window open on each — reading one, working in another — without switching back and forth.

## What this adds

A macOS-only `cswap launch <num|email|alias>`:

- Launches the desktop app for that account on its own profile directory, or **focuses** the window if it is already running rather than spawning a duplicate.
- `cswap launch --all` for every managed account.
- `cswap launch --stop <N>` / `--stop --all`.
- Creates the profile on first use and seeds it with `claude_desktop_config.json`, `Claude Extensions` and that account's own `claude-code-sessions/<uuid>` subtree, so a new window is not bare.
- Account resolution goes through `switcher.resolve_account`, so slot numbers, emails and aliases all work — `cswap launch rb`.
- Non-macOS exits non-zero with a plain message, no traceback.

## How it works

Claude.app is Electron and accepts Chromium's `--user-data-dir`; the shipping app passes that switch itself. It does not call `requestSingleInstanceLock` — 0 occurrences in `app.asar` — so instances on distinct profile directories coexist. No symlinking, no quitting the app, no moving directories around.

Credentials are deliberately **not** copied between profiles: `Cookies`, `config.json` and `ant-device-registry.json` are excluded from seeding, so each profile signs in on its own. That keeps the accounts genuinely separate and means the command never touches credential material.

## Relationship to #230

#230 also targets the desktop app, and I want to be explicit that this is not meant to compete with it. They solve different problems and can coexist:

- #230's `cswap desktop <account>` **switches** one desktop app between accounts, and reconciles session indexes into the selected profile — work this PR does not do.
- `cswap launch <N>` runs several accounts **concurrently**.

I put the code in `launcher.py` rather than `desktop.py`, and used the verb `launch` rather than `desktop`, specifically so both can land without a collision. If you would rather have one desktop command overall, I am happy to rework this on top of #230 instead — say the word and I will.

One observation that may be useful there regardless of what happens to this PR: because the app honours `--user-data-dir` and holds no single-instance lock, profile selection does not require quitting the app or symlinking a directory into place. That may let #230 drop its shutdown-and-rollback path.

## Validation

- `uv run pytest` — 2117 passed, 3 failed, 3 skipped. The same 3 fail unchanged on the base commit (`test_move_accounts`, `test_real_store_guard`, `test_swap_accounts`, all in unreadable-source guards); they are not touched by this change.
- 32 new tests covering happy path, unknown account, non-macOS, missing app bundle, already-running (focuses without duplicating) and existing-profile (not re-seeded).
- Each new behaviour has a negative control: the behaviour was broken at the source, the relevant tests were observed failing for the right reason, then restored.
- Exercised live on macOS 26.6 with Claude.app 1.30096.1 over roughly an hour: three signed-in windows on three accounts concurrently; `cswap launch 3` on an already-running slot printed `Focused … (pid 46730)` and the pid did not change; `cswap launch rb` resolved through an alias to the same window; `cswap launch 99` errored cleanly.

## Scope

Does not transfer or read desktop credentials, does not modify transcripts, and does not merge Claude Chat or Cowork history. It creates local profile directories and launches the app against them. macOS only — Windows and Linux fall through to a clear error rather than a partial implementation, since I could not test them.

I have not measured behaviour on a machine where the desktop app has never been signed in at all; the first launch there simply shows the normal sign-in screen, but I have only seen that path on my own machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,21 @@ Subfolders inherit the nearest mapped ancestor. In an unmapped directory, `cswap
 
 </details>
 
+### Run several desktop apps at once (macOS)
+
+Launch the Claude desktop app for a stored account on its own profile, so several signed-in windows can run side by side — each account gets its own `Claude-a<N>` folder under `~/Library/Application Support`, separate from your normal `Claude.app` login.
+
+```bash
+cswap launch 2                  # launch (or focus, if already running) account 2
+cswap launch user@example.com   # by email
+cswap launch dev                # by alias
+cswap launch --all              # launch every managed account
+cswap launch --stop 2           # quit one account's app
+cswap launch --stop --all       # quit every launched account
+```
+
+Re-running `cswap launch` on an account that's already open brings its window to the front instead of opening a duplicate. A profile's cookies and login are never shared — copied only once, on first launch, from your default `Claude.app` profile: `claude_desktop_config.json`, extensions, and that account's own Code sidebar sessions. Each window signs in independently. macOS only.
+
 ### Interactive dashboard (TUI)
 
 Run `cswap` on its own (or `cswap tui`) for the full-screen dashboard: live usage for every account, switching, and the auto-switcher, all keyboard-driven. `cswap watch` opens it straight to the live monitor. Works on macOS, Linux, and Windows.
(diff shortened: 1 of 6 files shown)


thalesfsp:
Closing this. Withdrawing on my side rather than leaving it open for review — the underlying use case did not hold up: the desktop app's sign-in depends on a claude.ai session cookie, not the OAuth token cswap stores, so an account cswap manages still needs its own one-time sign-in and the command does not deliver what I wanted from it. Apologies for the noise.

The observation in my comment on #230 stands on its own if it is useful there: the app accepts --user-data-dir and holds no single-instance lock, so profile selection does not require quitting the app.

PR #250, opened 2026-08-15
feat(autoswitch): spend the last of a weekly window, and resume sessions it stopped

deathemperor:
Three related changes, driven by one problem: on a large plan the last few percent of a weekly window is real money, and today it is stranded.

## 1. `consume-first` no longer stops at the threshold

The threshold is `best`'s "start looking" line. Under `consume-first` the point is to spend a perishable weekly window before it resets, so the last few percent are exactly what must be burned — they evaporate at the reset either way. Previously a user on `consume-first` with `threshold: 98` abandoned each account at 98%, permanently losing that 2%.

The threshold keeps its **other three roles** under this strategy. Each is covered by a test that failed while getting here:

| role | behaviour |
|---|---|
| Landing gate | unchanged — still refuses a target at/above it, so we never move onto an account that re-triggers next tick |
| At-limit escape | unchanged — keys on `active_headroom <= 0`, not on the threshold, so an exhausted account still escapes to any healthy peer |
| All-peers-above detector | unchanged — `_every_account_above_threshold` still arms the recovery-axis ranking and its `RECOVERY_HYSTERESIS_S` guard, so two accounts rolling over a minute apart cannot trade places forever |

`best` is untouched.

## 2. The menu bar surfaces the strategy

There was no strategy control at all, so a menu-bar user could be running `consume-first` with nothing on screen saying so. Adds an *Auto-switch strategy* submenu, and renames the threshold item to **"Landing limit"** under `consume-first` — calling it "Auto-switch threshold" there would promise a switch that deliberately never comes.

## 3. New: resume Claude Code sessions stopped by a usage limit

`autoswitch.resumeStoppedSessions`, **off by default**.

When every account is exhausted, in-flight sessions stop with a rate-limit turn and stay stopped long after quota returns. With this on, cswap records the sessions it *witnesses* stopping and nudges them once the active account has headroom again.

The trigger is **"the account we are on can work"**, not "a switch happened". The two come apart: when the active account's own window rolls over there is nothing better to move to, so no switch fires — and hanging the nudge on the switch alone stranded exactly the sessions the feature exists for.

Conservative by construction:

- only sessions observed stopping **while the engine was running** are eligible, so one that stopped beforehand, or was already resumed by hand, is never touched;
- the record is held **in memory**, not in `autoswitch_state.json` — a pid from a previous engine run may since belong to an unrelated process. Losing the record costs an un-resumed session; persisting it risks messaging a stranger.

### On the undocumented interfaces

This talks to Claude Code two ways, neither of them public API:

- the **transcript tail**, to tell a terminal limit stop (`model: "<synthetic>"` + `isApiErrorMessage` + `error: "rate_limit"`, no `retryAttempt`) from a retryable mid-turn 429 that Claude Code recovers from itself;
- the **per-session Unix socket** from `~/.claude/sessions/<pid>.json`, to deliver the nudge as a peer message.

Both are versioned by `peerProtocol`, which is checked against the version this was written against. An unrecognised one logs a warning and resumes nothing rather than sending guessed frames into a live session. Every path degrades to "do not resume" rather than raising, so a Claude Code change costs the feature and never the auto-switch loop.

The discriminator is tested against **real captured transcript entries**, not reconstructions.

### Caveat worth knowing

A resumed session **continues its previous work unattended**, and so keeps spending quota. That is the point of the feature, but it is why it is opt-in. Documented in the README.

## Tests

+27 (2085 → 2112 passing). Three failures are pre-existing on a clean tree (`test_move_accounts`, `test_real_store_guard`, `test_swap_accounts`) and unrelated.

Two existing tests changed shape:

- `test_a_departure_at_full_quota_is_immediately_eligible` drove a `consume-first` return by burning the active account past the threshold, which is no longer a trigger for that strategy — it now drives the return via reset ordering, the axis `consume-first` actually ranks on. Same invariant (a full-quota departure is not a permanent lockout).
- two hardcoded settings counts in `test_config_cli.py` now derive from `SETTING_SPECS`, so adding a setting cannot fail a test that is not about the count.


jameswniu:
Running this branch (d0f824a) on a real fleet since tonight: 5 OAuth accounts on macOS, launchd tick every 300s, threshold 90, autoswitch.model Fable.

Install from the branch was clean over 0.25.0 (accounts and config preserved), autoswitch.resumeStoppedSessions was accepted and enabled, and the auto loop runs green (below-threshold holds, at-limit switches, all-exhausted parks all behave as on 0.25.0).

Context that motivated the trial: earlier tonight, during a genuine all-accounts-exhausted window, two interactive sessions stopped on the rate-limit turn and stayed stopped after quota returned, exactly the case this PR describes. I will report back what the resume path does on the next real wall event.

Happy to test follow-up commits on the same fleet if useful.

jameswniu:
CI note on the test-windows job: all 4 failures are one root cause, the test fixture itself builds an AF_UNIX listener (tests/test_session_resume.py:255) and Windows Python has no socket.AF_UNIX, so the suite cannot even stand up its fake session inbox there. Linux and macOS jobs are green at d0f824a, which matches our fleet result (macOS, feature enabled, daemon green). A skipif on hasattr(socket, "AF_UNIX") would make the Windows job honest until a named-pipe transport exists, if Windows support is even in scope for this feature.

jameswniu:
The Windows CI fix from the note above is now a mergeable commit: deathemperor/claude-swap#1, stacked onto this PR's branch. It marks exactly the four AF_UNIX wire tests with a skipif and leaves the fifteen platform-neutral tests running on Windows. Verified locally on macOS: 19 passed, no skips triggered.

deathemperor:
Update on the scope of this PR: part 3 (session resume) is moving out of the engine and into the **Infinitus** menu bar app — https://github.com/deathemperor/infinitus — rather than staying proposed here.

What moved (Swift port, fully app-side, no engine changes needed):
- Detection: reads Claude Code's own session records (`~/.claude/sessions/<pid>.json`) and transcript tails to find sessions whose last turn is a plan-limit stop (same classification as `session_resume.py`: `isApiErrorMessage` + `rate_limit`, no `retryAttempt`; retryable 429s are left alone).
- Delivery: terminal PTYs first — cmux, tmux and herdr (the last two are new, herdr sessions never bind a peer socket) — with the peer socket (`<cross-session-message>` envelope, byte-compatible with this branch's `wrap_peer_body`) as fallback. Same anti-spam rules: a typed nudge never also goes to the socket, a held/unchanged stop is never re-nudged, a burned nudge is retried with distinct text.
- `/rc` re-arm after a switch, with the same idle filter and confirm/dismiss pass.

Rationale: the nudge needs terminal access and Claude Code's session files, which sit naturally with the desktop app; the engine stays a pure credential switcher. It lands in the app's next release.

For this PR: parts 1–2 (consume-first weekly burn) are unaffected and still proposed as-is. Nothing in this branch is deleted or changed by the move — the resume code here keeps working exactly as it does now for anyone testing it (@jameswniu's fleet trial included). If maintainers would rather take parts 1–2 alone, I'm happy to trim the branch.

deathemperor:
@realiti4 I needed more GUI to track the accounts, thought of adding it here but figured what I want to make become irrelevant to the engine so I decided to create a separate repo for the GUI app while still use cswap as an engine.

https://github.com/deathemperor/infinitus

Would you mind endorse the app? It's my first public repo/app that built from scratch. If allowed, I'll make a PR.

PR #257, opened 2026-08-17
fix(usage): a refused certificate is not an unreachable host

codeslake:
## The defect

`_classify_usage_error` maps every `urllib.error.URLError` that is not a timeout
to `network`. A TLS certificate the client *refused* therefore gets recorded
with the same word as a host it could not *reach*.

Those two need opposite fixes. `network` points at DNS and connectivity; the
repair for a refused certificate is a CA bundle. Anyone reading the stored
failure sees only the first.

## Measured

On a machine whose `HTTPS_PROXY` is a TLS-terminating proxy, every usage poll
raised:

```
URLError(SSLCertVerificationError(1, "[SSL: CERTIFICATE_VERIFY_FAILED]
certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)"))
```

and each was stored as `network`. One account reached **1155 consecutive
failures over ten days** with that one word as the entire record — the exception
repr only reaches `DEBUG`, which nothing enables in normal operation.

The same request with `SSL_CERT_FILE` pointing at the proxy's CA returns
`HTTP 200`. The account, the token and the endpoint were healthy the whole time.

| `SSL_CERT_FILE` | result |
| --- | --- |
| unset | `URLError` → recorded `network` |
| proxy CA bundle | `HTTP 200`, usage parsed |
| system CA bundle | `URLError` (system store does not trust the proxy) |
| unset, no proxy | `HTTP 200` (control) |

## Not specific to one setup

`urllib` reads `SSL_CERT_FILE` and reads neither `REQUESTS_CA_BUNDLE` nor
`NODE_EXTRA_CA_CERTS`. A user behind a corporate MITM proxy (Zscaler, Netskope,
and similar) who has configured either of those still fails here, and still sees
only `network`.

## Scope

Deliberately narrow. No renames, no control-flow changes. The guard fires on
`SSLCertVerificationError` alone, so an ordinary transport error keeps reporting
`network`.

This PR only stops the cause from being erased. It does **not** make cswap read
anyone's CA bundle — that is a separate decision and is not proposed here.

## Tests

- `test_tls_cert_failure_is_not_flattened_to_network` — written first, failed
  with `assert 'network' == 'tls-cert'`, passes after the fix.
- `test_plain_transport_failure_still_reads_as_network` — the control, so the
  narrowing cannot swallow the case `network` exists for.

Mutation-checked: removing the guard makes the first test fail and leaves the
control passing. Full suite `2090 passed, 3 skipped`.

---

### DIFF
--- a/src/claude_swap/cli.py
+++ b/src/claude_swap/cli.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import logging
 import os
 import sys
 
@@ -878,6 +879,9 @@ Examples:
         sys.exit(130)
 
 
+_logger = logging.getLogger("claude-swap")
+
+
 def _use_native_tls() -> None:
     """Route TLS trust decisions through the OS-native verifier.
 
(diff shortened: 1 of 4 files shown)


codeslake:
Closing: this is being folded into #210 so the maintainer sees one PR rather than two for the same change. The branch `fix/usage-tls-cause` stays pushed at 2efe80b and the commits move over intact — nothing here is abandoned.

Summary of what moves, for anyone following the thread: a `URLError` whose `.reason` is `ssl.SSLCertVerificationError` was recorded as `network`, the same token as an unreachable host, so a refused certificate and a dead route were indistinguishable in the stored failure. One account accumulated 1155+ consecutive failures over ten days with `network` as the entire diagnostic record. It now records `tls-cert` and carries a remedy note.

PR #262, opened 2026-08-19
Add opt-in 'pace' auto-switch strategy: rotate off accounts burning ahead of weekly pace

schwarztim:
## Motivation

The auto-switch engine is purely reactive today: nothing moves until the binding window crosses `autoswitch.threshold`. An account burning well ahead of its weekly pace therefore gets ridden all the way to the cliff while on-pace peers sit idle, and by the time the threshold trips, that week's quota is effectively gone.

Meanwhile `pace.py` has computed exactly the leading signal for this since #125: expected-versus-actual weekly consumption, noise-gated with the 15-point `AHEAD_THRESHOLD_PCT` and the 24h post-reset suppression. Today it only feeds the menubar `(ahead)` marker and the JSON output; no decision surface consumes it.

Observed on a real 3-account fleet: one account at 81% of its weekly window with 3 days left in the cycle ("ahead of pace" in `cswap list`), while the other two sat near 0%. `best` keeps you parked until the threshold; by then the hot account's week is spent.

## What this adds

An opt-in third strategy, `pace` (`cswap auto --strategy pace`, or `cswap config set autoswitch.strategy pace`). Default behavior is unchanged; `best` and `consume-first` are untouched.

- At/above the threshold, `pace` behaves exactly like `best` (same triggers, same ranking, same hysteresis).
- Below the threshold, when any weekly window of the active account (7d, plus configured scoped per-model windows via the existing `models` filter) is meaningfully ahead of pace, it rotates to the on-pace candidate with the most headroom, with trigger `"pace"`.

## Design choices

- **The trigger is `compute_pace(...).ahead`, unchanged.** No new thresholds: the engine and the UI marker can never disagree about what "ahead" means, and the signal keeps its built-in noise gates (24h suppression, 15-point margin). The sensitive `will_last_to_reset` signal was deliberately not used; acting on "1 point over expected" would be far too twitchy for an automated mover, which is the same reasoning that kept it JSON-only in #125.
- **Landing gate: a target must itself be on pace** (in addition to the existing below-threshold health gate). Moving onto another over-pace account trades one projected exhaustion for the next and invites ping-pong. Unknown pace counts as on pace: suppression covers exactly the first day after a reset when a window is freshest, and an unknowable pace cannot prove a candidate hot.
- **No headroom hysteresis on the pace branch, deliberately.** The point of the move is that raw headroom misleads when burn rates differ: an on-pace account with less headroom than the over-pace active is still the account that lasts the week. Anti-flap comes from the cooldown, the no-return bar, the target-must-be-on-pace gate, and pace's own decay (a departed account stays "ahead" until expected catches its actual, hours at minimum).
- **Everything else is reused, not duplicated:** the below-threshold move takes the consume-first two-phase commit (provisional decision on the stored snapshot, escalated refetch and re-rank before switching) and the per-target `UsageEntry.fresh` gate; cooldown and the no-return bar apply; the API-key last-resort fallback is excluded for the same reason it is for consume-first (an API-key account has no weekly window to pace).
- Weekly windows come through `oauth.relevant_windows` (the canonical window source), so a window that binds a decision can never be invisible to the pace check; the 5h window is excluded per `pace.py`.

## Testing

- 10 new tests in `tests/test_autoswitch.py` (`TestPaceStrategy`), including the discriminating case: an ahead-of-pace candidate with MORE headroom than the on-pace one must be passed over, which a plain headroom sort gets wrong.
- Coverage for: the suppression window holding below threshold, every-candidate-ahead holding with the new `no-on-pace-candidate` reason (NO_ACTION, matching the consume-first hold contract), the landing health gate, cooldown, above-threshold behavior matching `best`, the 5h exclusion, scoped-model windows, and settings validation.
- `tests/test_autoswitch.py`: 270 passed. Full suite: 2095 passed, 3 skipped, plus 3 pre-existing failures in `test_move_accounts`/`test_swap_accounts`/`test_real_store_guard` that fail identically on unmodified `main` in my environment (macOS with a legacy `~/.claude-swap-backup` store layout); untouched by this change.

## Docs

README strategy section and example updated; `--strategy` help, settings spec choices, and the module docstring cover the new value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ cswap auto --once              # single check-and-switch, for cron/scripts
 cswap auto --dry-run           # log what it would do, never switch
 cswap auto --strategy consume-first   # burn the soonest-resetting account first
 cswap auto --strategy pace            # rotate off accounts burning ahead of weekly pace
+cswap auto --threshold-5h 95 --threshold-7d 85   # ride the 5h window, protect weekly quota
 ```
 
 <details>
@@ -99,6 +100,7 @@ cswap auto --strategy pace            # rotate off accounts burning ahead of wee
 
 - Runs safely alongside Claude Code: switches take the same credential locks Claude Code uses, so a swap never collides with a token refresh.
 - A cooldown (default 5 min) and a hysteresis margin stop it flip-flopping near the threshold: a proactive switch only lands on an account that's below the threshold *and* better than the current one by the margin — a candidate that clears the margin is always taken, but two accounts hovering at the line never ping-pong. When every account is exhausted it keeps checking on a bounded slow cadence, waking sooner for an imminent reset.
+- **Per-window thresholds** (`--threshold-5h` / `--threshold-7d`, or `cswap config set autoswitch.threshold5h`): the two windows recover on different scales, so one ceiling suits neither. A 5-hour window at 95% is whole again within the session; a weekly window at 95% is spent for days. Set them apart to ride the 5h window (fewer switches) while leaving weekly quota earlier (more of it left for the rest of the week). Per-model weekly limits follow `--threshold-7d`. Unset, both fall back to `--threshold` and behavior is unchanged.
 - **Strategies** (`--strategy`, or `cswap config set autoswitch.strategy`): `best` (default) stays put until the active account nears its limit, then moves to the account with the most quota left. `consume-first` proactively keeps you on the account whose **weekly window resets soonest** — use-it-or-lose-it — switching to a sooner-resetting account (with room to spare) even below the threshold, so perishable weekly quota isn't wasted. `pace` keeps `best`'s behavior and adds an early rotation: when the active account's weekly window is meaningfully ahead of pace (the dashboard's `(ahead)` marker), it moves to the on-pace account with the most quota left, so a hot week gets spread across accounts before anything nears its limit.
 - Usage polling is adaptive — a couple of accounts per check, busy alternates watched more closely, and exhausted ones checked about every ten minutes (or slower after 429s) — so API traffic stays flat no matter how many accounts you manage.
 - It fails safe: if a usage check errors it keeps trusting the last-known numbers while retries back off, and an expired token on an idle machine makes it hold rather than fail over (Claude Code refreshes the token on your next message).
--- a/README.md
+++ b/README.md
@@ -91,6 +91,7 @@ cswap auto --model Fable       # also switch when the Fable weekly limit is hit
 cswap auto --once              # single check-and-switch, for cron/scripts
 cswap auto --dry-run           # log what it would do, never switch
 cswap auto --strategy consume-first   # burn the soonest-resetting account first
+cswap auto --strategy pace            # rotate off accounts burning ahead of weekly pace
 ```
 
 <details>
@@ -98,7 +99,7 @@ cswap auto --strategy consume-first   # burn the soonest-resetting account first
 
 - Runs safely alongside Claude Code: switches take the same credential locks Claude Code uses, so a swap never collides with a token refresh.
 - A cooldown (default 5 min) and a hysteresis margin stop it flip-flopping near the threshold: a proactive switch only lands on an account that's below the threshold *and* better than the current one by the margin — a candidate that clears the margin is always taken, but two accounts hovering at the line never ping-pong. When every account is exhausted it keeps checking on a bounded slow cadence, waking sooner for an imminent reset.
-- **Strategies** (`--strategy`, or `cswap config set autoswitch.strategy`): `best` (default) stays put until the active account nears its limit, then moves to the account with the most quota left. `consume-first` proactively keeps you on the account whose **weekly window resets soonest** — use-it-or-lose-it — switching to a sooner-resetting account (with room to spare) even below the threshold, so perishable weekly quota isn't wasted.
+- **Strategies** (`--strategy`, or `cswap config set autoswitch.strategy`): `best` (default) stays put until the active account nears its limit, then moves to the account with the most quota left. `consume-first` proactively keeps you on the account whose **weekly window resets soonest** — use-it-or-lose-it — switching to a sooner-resetting account (with room to spare) even below the threshold, so perishable weekly quota isn't wasted. `pace` keeps `best`'s behavior and adds an early rotation: when the active account's weekly window is meaningfully ahead of pace (the dashboard's `(ahead)` marker), it moves to the on-pace account with the most quota left, so a hot week gets spread across accounts before anything nears its limit.
 - Usage polling is adaptive — a couple of accounts per check, busy alternates watched more closely, and exhausted ones checked about every ten minutes (or slower after 429s) — so API traffic stays flat no matter how many accounts you manage.
 - It fails safe: if a usage check errors it keeps trusting the last-known numbers while retries back off, and an expired token on an idle machine makes it hold rather than fail over (Claude Code refreshes the token on your next message).
 - An account whose refresh token has died is quarantined and reported until you either log in with it and re-run `cswap add --slot N`, or replace its stored credentials from a known-good export — a plain `cswap import backup.cswap` replaces dead-token slots on its own (`--force` is still required to replace other existing accounts; note a stale export can carry an already-superseded token). API-key accounts are never rotated onto unless you pass `--include-api-key-accounts`.
--- a/src/claude_swap/autoswitch.py
+++ b/src/claude_swap/autoswitch.py
@@ -323,6 +323,12 @@ class PollEvent(AutoSwitchEvent):
     # (e.g. "89%") hides which window binds — #115 was reported off that
     # ambiguity.
     windows: dict[str, dict[str, float]] = field(default_factory=dict)
+    # Resolved per-window ceilings, emitted only when they differ from each
+    # other. `threshold` alone cannot describe a split policy, so a reader
+    # of the JSON stream would see "threshold: 88" while the engine was
+    # actually gating on 95/85 and be unable to explain any decision. Kept
+    # conditional so the default single-threshold output is byte-identical.
+    window_thresholds: dict[str, float] = field(default_factory=dict)
 
     def _fields(self) -> dict:
         fields = {
@@ -330,6 +336,8 @@ class PollEvent(AutoSwitchEvent):
             "headroomPct": self.headroom,
             "threshold": self.threshold,
         }
+        if self.window_thresholds:
+            fields["windowThresholds"] = self.window_thresholds
         if self.fetch_errors:
             fields["fetchErrors"] = self.fetch_errors
         if self.windows:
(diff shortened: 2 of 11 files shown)


schwarztim:
**Why this matters, beyond the mechanics**

The implementation details are in the description above. This is the user-facing case for it.

claude-swap's promise is "switch before you hit a rate limit," but the engine can only watch one number: current utilization. It holds until that number crosses the threshold, and for a weekly window that is already too late. The account is spent for the week by the time anything moves. Acting on the trend rather than the level is the difference between "you are almost out" and "you will run out on Wednesday."

`pace.py` already computes that trend, and computes it carefully. The noise gates in there (24h post-reset suppression, the 15-point margin) are exactly what an automated mover needs to avoid firing on variance. Today they only render an `(ahead)` marker, so the tool notices the problem and then leaves the fix to the user, who has to spot the marker and switch by hand. That is the manual babysitting the auto-switcher exists to remove. This PR is mostly a wire between two things that already exist.

The 5h window does not need this. It recycles before lunch and a static threshold handles it well. The weekly window does need it, because a week is long enough that burn *rate* matters and a threshold cannot see rate. A weekly limit hit on Wednesday locks you out until the weekend.

It also makes a multi-account setup behave like a fleet. The current policy parks on one account and drains it while the others idle and their perishable quota expires. Spreading the burn is the reason people add a second and third account in the first place.

The case that prompted this: a three-account fleet with one account at 81% of its weekly window and three days left in the cycle, flagged `(ahead of pace)` right there in `cswap list`, while the other two sat near 0%. Stock `best` would not have touched it until 90%.

Opt-in, defaults unchanged, and it reuses the existing health gate, cooldown, no-return bar, and two-phase commit rather than introducing a second definition of "ahead" to keep in sync.

schwarztim:
**Added a second commit: per-window thresholds (`threshold5h` / `threshold7d`).**

Happy to split this into its own PR if you would rather review the two separately. It is a self-contained second commit, so it cherry-picks cleanly either way. Posting the reasoning here so the diff is not the first thing you read.

### The problem

`account_headroom` returns `100 - max(pct)`, and `threshold` gates that single number. But the 5h and weekly windows recover on completely different scales:

- A 5-hour window at 92% is *temporarily* busy and is whole again within the session.
- A weekly window at 92% is spent for days.

One ceiling cannot serve both. Set it high enough to stop evicting accounts over a window that heals itself, and weekly quota gets burned to the same line. Set it low enough to protect weekly quota, and accounts get rotated off (with a cold prompt cache in every live session, since auth is machine-wide) to save a window that would have recovered on its own.

Measured on a real 3-account fleet, with the account-wide threshold at 88:

```
                 uniform 88                    split 5h=95 / 7d=85
  duck   5h 27 / 7d 84    -4.0 under        -1.0 under   (7d 84% < 85%)
  gmail  5h 89 / 7d 18    +1.0 OVER         -6.0 under   (5h 89% < 95%)
  qvc    5h 19 / 7d  3   -69.0 under       -76.0 under   (5h 19% < 95%)
```

Under one threshold, `gmail` is excluded as a landing target because a 5-hour window that resets in 90 minutes sits at 89, while 82 points of its weekly quota go unused. Meanwhile `duck` is free to keep burning the weekly window that actually matters until it too reaches 88. Both of those are backwards, and no single value fixes either without making the other worse.

### The change

Optional `autoswitch.threshold5h` / `autoswitch.threshold7d` (plus `--threshold-5h` / `--threshold-7d`), each falling back to `threshold` when unset. Scoped per-model windows resolve to the weekly ceiling, since they reset on the same 7-day cadence and are perishable the same way.

### Design notes

- **One primitive.** `oauth.threshold_excess` reports how far the worst window sits past *its own* ceiling. Every `(100 - headroom) >= threshold` site becomes `excess >= 0`; the escalation band becomes `excess >= -ESCALATION_MARGIN_PCT`. It is exactly the old comparison whenever the windows resolve to the same value, which `test_excess_matches_the_old_test_when_uniform` pins directly.
- **`account_headroom` is untouched.** Distance to the real 100% is a different axis from threshold policy. The at-limit escape and the ratio anti-flap gates are calibrated on it and keep reading it.
- **The linear margins moved onto the policy axis.** This was the one non-mechanical part. Raw headroom misranks under a split policy: an account bound by a weekly window at 84 shows *more* headroom than one bound by a 5h window at 90, while tripping five times sooner. Room-before-tripping equals `headroom - (100 - threshold)` under a uniform policy, a constant offset, so every ordering and margin is unchanged there. The existing suite is the proof: all 270 prior tests pass across that change with no expectation edits. The RATIO gates (`HORIZON_HEADROOM_RATIO`, `SPENT_HEADROOM_PCT`) deliberately stay on headroom, where a ratio against distance-to-100 still means something; a ratio against a policy line every account is already past does not.
- **Signature churn kept minimal.** `excess` is keyword-only on `_no_return_account` / `_left_account_recovered`, so the existing positional order is untouched.
- The TUI threshold tick now draws the policy floor, because one tick is rendered against bars for both windows and would otherwise sit above the real ceiling on whichever window was lowered.

### Testing

`TestPerWindowThresholds`, 11 cases, including a deliberate discriminating pair: one fleet, two policies, different landings, so the test proves the policy moved the answer rather than the numbers. Also covered: unset overrides producing a byte-identical message, a hot 5h window no longer evicting an account with weekly quota left (with the uniform control that does evict it), a weekly window tripping below the account-wide default, scoped model windows following the weekly ceiling, and 100% still reading as at-limit with both ceilings at 99.

Full suite: 2106 passed, 3 skipped. The same 3 pre-existing failures noted in the PR description (`test_move_accounts` / `test_swap_accounts` / `test_real_store_guard`) still fail identically on unmodified `main` in my environment and are untouched by either commit.

schwarztim:
**Third commit: a regression test for a real hard rate limit this hit in production today.**

Shortly after posting the per-window commit, the fleet I have been running this against took an actual 100% rate limit, and the log shows exactly the failure the split policy is meant to remove. Adding it as a pinned test since it is a much better argument than the hypothetical I opened with.

### What happened

Three accounts under heavy concurrent load, single threshold at 88. All three landed within a point or two of that ceiling at the same moment, but each on a **different window**:

```
  duck    5h 44 / 7d 88     bound by a weekly window ~3 days out
  gmail   5h 89 / 7d  0     bound by a 5h window 39 minutes out, full week untouched
  qvc     5h 96 / 7d 13     active, burning
```

Every account is over the shared ceiling, so no candidate is a legal landing. The engine reported `no-qualifying-candidate` for eight consecutive ticks and then rode the active account to a real 100%, taking a hard limit before the at-limit escape moved it:

```
no-switch  no-qualifying-candidate   (x8)
SWITCH     at-limit  3 -> 1
```

The painful part is `gmail`. It had just rolled over to a **completely untouched week** and was blocked only by a 5-hour window that would heal 39 minutes later. Under one ceiling it is indistinguishable from `duck`, whose weekly window is genuinely spent for days. The policy could not tell "busy for the next half hour" apart from "depleted until Friday", so it treated the fleet as exhausted while a full week of quota sat one window away.

### Under the split policy

```
                    uniform 88              split 5h=95 / 7d=85
  duck     excess    +0.0 illegal            +3.0 illegal   (weekly genuinely spent)
  gmail    excess    +1.0 illegal            -6.0 LEGAL     (5h under its own ceiling)
  qvc      excess    +8.0 illegal            +1.0 illegal
```

`gmail` becomes a legal landing and the limit is never reached. `duck` stays correctly excluded, which is the part a simple "raise the threshold" cannot do: raising the shared ceiling to 95 would have unblocked `gmail` and also handed `duck` seven more points of weekly quota to burn.

### The test

`test_the_fleet_that_rode_an_account_to_a_hard_limit` runs that exact fleet, with real reset offsets, through both policies in one case: the uniform one asserts `is not SWITCHED` (reproducing the block), the split one asserts it lands specifically on the account holding the full week rather than the one whose weekly is days out. It fails if either half regresses.

Full suite still 2107 passed, 3 skipped, with the same 3 pre-existing environment failures noted earlier.

schwarztim:
**Status update: three further commits, including a blocking regression this branch introduced and then fixed.**

The PR picked up independent review after my last comment. It found a real blocker, which is now fixed, plus two coverage gaps that were genuinely missing. Summarising here so the commit list is not three unexplained entries.

### `c63cafc` — fix(tui): the session threshold control was inert under a split policy

This is the important one, and it was a regression **this branch introduced**.

`apply_threshold` wrote only `settings.threshold`. But `for_label` reads the per-window overrides and never consults that fallback, so with `threshold5h=95` and `threshold7d=85` configured, stepping the session control to 72 displayed 72, logged a session change, and moved no ceiling at all. The effective policy stayed 95/85. It also contradicted the bar tick, which this same branch had already moved onto the policy floor.

The control now edits the **floor**, which is exactly what the tick renders, so the number shown and the number adjusted are the same number. Both ceilings shift by the same delta so the configured gap survives, and each clamps to the configured bounds independently. Stepping 95/85 (floor 85) down three gives 92/82, floor 82, tick reads 82.

I chose this over disabling the control under a split policy because the tick already renders the floor here, so making the control move the floor completes that decision rather than adding a second rule. Under a uniform policy the floor **is** the threshold, so every path is byte-identical to before, which is what keeps the existing session-override tests passing untouched.

Implementation note: `shifted_to_floor` is a pure function applied by both the screen and the engine to the same input, rather than the screen reading back what the engine achieved. Reading back would couple the view to a live engine existing at all, and the value is identical either way.

### `e1b1bf4` — test: close the coverage gaps review raised

Review's point was that this change moves ranking, margins, escalation, CLI/config, and TUI rendering onto a new policy axis, and each deserves proof. Three were genuinely uncovered:

- **CLI precedence:** `--threshold-5h` / `--threshold-7d` beat settings.json, leave the fallback `threshold` alone, do not clear the sibling when only one flag is passed, and clamp like every other numeric flag.
- **TUI:** the tick renders the policy floor under a split policy (one tick is drawn against bars for both windows, so the fallback would sit above the real ceiling on whichever window was lowered), with a uniform-policy control.
- **Boundaries:** `excess >= 0` is the trip test, so equality trips. Pinned per window because a split policy has two independent boundaries, plus a legacy-equivalence sweep proving `excess >= 0` agrees with the old `(100 - headroom) >= threshold` at the boundary itself across 50/75/88/99.9.

Also pins that the no-return anti-flap bar still records a departure under a split policy, so the new axis does not bypass it.

### `646e9e2` — feat: report the resolved ceilings in poll JSON

`threshold` alone cannot describe a split policy. A reader of the JSON stream saw `"threshold": 88` while the engine gated on 95/85, with no way to explain any decision it made. Adds a `windowThresholds` object, emitted only when the two ceilings differ, so default single-threshold output stays byte-identical.

### Testing

Full suite: **2129 passed, 3 skipped**. The same 3 pre-existing failures noted in the description (`test_move_accounts` / `test_swap_accounts` / `test_real_store_guard`) still fail identically on unmodified `main` in my environment and are untouched by any commit here.

One practical note: there is no CI configured on this repository, so there are no checks to go green on the branch. The suite result above is the only signal available, and I would rather say that plainly than leave a clean-CI condition standing that nothing can satisfy.

Still happy to split the pace strategy and the per-window thresholds into separate PRs if that is easier to review. They are self-contained commits and cherry-pick cleanly either way.

PR #264, opened 2026-08-19
Read setup-token usage from rate-limit headers

pradeepkumar-qure:
## Problem

A `claude setup-token` credential is scoped to `user:inference` alone. The usage endpoint (`/api/oauth/usage`) needs profile scope, so these accounts 401/403 there on every pass and render a permanent `usage unavailable`.

The bigger consequence is invisibility to `cswap auto`: `_rank_candidates` skips any account whose headroom is `None`, so a setup-token account is never chosen as a switch target — and never triggers a switch away from itself, even at 100%.

```
2: user@example.com [personal]
   usage unavailable (http-403)
```

## Approach

Every `/v1/messages` response carries the same 5h/7d numbers in `anthropic-ratelimit-unified-*` headers, and inference is exactly what these tokens are permitted to do:

```
anthropic-ratelimit-unified-5h-utilization: 0.04
anthropic-ratelimit-unified-5h-reset: 1787184600
anthropic-ratelimit-unified-7d-utilization: 0.06
anthropic-ratelimit-unified-7d-reset: 1787655600
```

So for inference-only credentials, probe those headers instead of spending a guaranteed 401/403 per pass. The probe returns data in the *raw* usage-API shape, so `build_usage_result` normalizes it unchanged and nothing downstream needs to know the difference.

Detection is narrow — `scopes == {"user:inference"}` exactly — so full-scope logins keep using the real endpoint.

## Cost, and why it is throttled

The probe has to be a **successful** call. Rejected requests return no rate-limit headers at all; both were measured against the live API:

| request | status | rate-limit headers |
|---|---|---|
| empty `messages` array | 400 | none |
| bogus model name | 404 | none |
| minimal valid call | 200 | present |

So each probe costs one output token. `PROBE_MIN_INTERVAL_S` (default 600s, keyed by token digest) keeps the poll loop from consuming the very window it is measuring. Both the interval and the model are env-overridable (`CLAUDE_SWAP_PROBE_INTERVAL`, `CLAUDE_SWAP_PROBE_MODEL`) — a hardcoded model id eventually gets retired out from under us.

## Result

```
2: user@example.com [personal]
   ├ 5h:   3%   resets 16:00         in 31m
   └ 7d:  98%   resets Aug 22 05:00  in 2d 13h  (ahead of pace)
```

Verified against three real setup-token accounts, each reporting its own distinct utilization and reset times.

## Limitation

Per-model weekly windows are absent from the headers, so `--model` triggers stay blind for probed accounts. Account-wide 5h/7d — the windows that gate everything — now work.

## Tests

`tests/test_probe_usage.py`, 12 cases: scope detection, the 0-1 → 0-100 scaling (getting this wrong renders every percentage 100x low), epoch → ISO round trip through `format_reset`, normalization through `build_usage_result` into a non-`None` `account_headroom`, malformed/absent headers, and the throttle's cache-hit / expiry / per-token isolation.

Full suite: 2099 passed. One pre-existing failure (`test_settings.py::TestAtomicWriteThroughSymlink`) reproduces on a clean checkout of `main` — a WSL `/mnt/c` permissions quirk, unrelated to this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,17 @@ cswap add-token --email user@example.com     # optional label override
 `--email` is optional; omitted values use `setup-token-{slot}@token.local`
 (or `api-key-{slot}@token.local` for API keys). No Anthropic API calls are made.
 
+**Usage for setup-token accounts.** A setup-token is scoped to `user:inference`
+alone, so it cannot read the usage endpoint (that needs profile scope) and such
+accounts used to show a permanent `usage unavailable`, which also hid them from
+`cswap auto` — an account with unknown headroom is never picked as a target.
+Their 5h/7d numbers are instead read from the `anthropic-ratelimit-unified-*`
+headers on a minimal `/v1/messages` call. That probe costs one output token, so
+it is throttled to once per 10 minutes per token (`CLAUDE_SWAP_PROBE_INTERVAL`,
+seconds; `CLAUDE_SWAP_PROBE_MODEL` overrides the model). Per-model weekly
+windows are not in the headers, so `--model` triggers stay blind for these
+accounts.
+
 **API-key accounts.** An `sk-ant-api...` value registers a managed API-key account
 (the kind Claude Code uses after `/login` with a key) rather than an OAuth
 setup-token. It switches like any other account; since API keys have no subscription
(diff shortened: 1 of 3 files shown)


pradeepkumar-qure:
Opened against the wrong repository by mistake — this belongs on my own fork for now. Apologies for the noise.

PR #266, opened 2026-08-20
fix(locks): a held lock could go stale, and a small timeout was not one

codeslake:
Two defects in `proper_lockfile()`, both in the path meant to keep a lock alive while its holder still has it. They were found on a branch that had nothing to do with locking, so they are split out here: `claude_locks.py` is core cswap, and the first of the two was additionally duplicated onto a second PR byte for byte.

### 1. The heartbeat disarmed on any error

`_touch()` returned on any `OSError`:

```python
except OSError:
    return  # lock stolen/removed; nothing left to keep alive
```

The comment states the intent — stop when the lock is gone — and the code cannot tell that from a transient failure. One failed `os.utime` killed the heartbeat while the lock was **still held**, and with `CONFIG_STALENESS_S = 10` against `TOUCH_INTERVAL_S = 3` any waiter could then legally steal it ten seconds later.

It surfaced as a Windows CI failure elsewhere: `heal` took a lock a test was holding live, with `Lock ... vanished while held (taken over as stale?)` logged underneath by the real holder.

**Asking the filesystem which case it is does not work**, and is wrong on every supported Python for two different reasons:

| Python | `Path.exists()` on ESTALE / EACCES / EIO | Result |
| --- | --- | --- |
| 3.12 (what CI runs) | re-raises out of the thread | toucher dies — the same bug |
| 3.13 | re-raises out of the thread | toucher dies — the same bug |
| 3.14+ | swallows and answers `False` | reads "gone" for a lock we hold — the same bug |

The boundary is **3.14**, not 3.13; an earlier revision of this description put
it one version early. Measured on four interpreters, EACCES on the parent so
the `stat` is refused: 3.10 / 3.12 / 3.13 all raise `PermissionError`, 3.14
answers `False`. 3.13's `Path.exists` still routes through `_ignore_error`,
which re-raises EACCES; 3.14's is `return os.path.exists(self)`, which does
not. Both halves are still live under `requires-python = ">=3.12"`.

`requires-python` is `>=3.12`, so neither half is theoretical. The errno is already in hand and is the only thing that separates the two, so it decides.

### 2. `timeout` did not bound the call

The contention retry slept a full jittered 0.25–0.5 s regardless of the deadline, so the loop's own timeout check could not fire until the sleep that overran it had finished:

| `timeout` | worst elapsed |
| --- | --- |
| 0.01 s | 0.408 s |
| 0.25 s | 0.487 s |
| 0.5 s | 0.910 s |

A deadline is computed once and both sleep sites clamp to what is left of it. With a full sleep's worth of budget the clamp is a no-op and the jitter still spreads waiters apart. The stale-lock branch's own 0.05 s sleep had the same defect at a smaller ceiling and is clamped too.

### Tests

The control matters as much as the fix: retrying forever on a lock that is **gone** is the opposite defect, and is what the original `return` was for.

That control was first written as `assert not lock_dir.exists()` after `rmdir`-ing the lock — which `os.utime` cannot falsify, so it passed for a toucher that never stops at all. It now counts the toucher's own calls, keyed on the lock path rather than on a global counter, because the module does `import os` and a process-wide counter is consumed by whatever else calls `utime` first.

The errno test injects a failure into `os.utime` **and** `os.stat`, because that is what these errnos mean on a real filesystem — a stale handle or an unreadable directory does not answer `stat` either. Injecting only a `utime` failure leaves the directory genuinely present, so any implementation that asks the filesystem gets a truthful "still there" and the test proves nothing. That version was written first and passed against the broken code.

### Scope note: the once-only warning is in this diff

Review item 4(b) named the silent persistent retry and closed it as a
follow-up, and the reply here agreed, saying the whole `src/` change was one
word inside one comment. That was true then and is not true now: a later
commit added a `warned` latch and a `logger.warning` to the toucher, plus its
own test. It is disclosed here rather than left for a re-read to discover.

What it does: on the first non-ENOENT `utime` failure only, the toucher logs
once that the lock's mtime has stopped advancing, so a takeover that follows is
legitimate and no longer arrives with nothing recording its cause. Once and not
per attempt, because a failure that never clears would otherwise emit at the
touch interval for the whole hold. It is a WARNING from a daemon thread and
changes no control flow: the heartbeat stays armed exactly as it did before.

Mutation, each reversion caught by the case that should:

| reversion | tests failed |
| --- | --- |
| back to `lock_dir.exists()` | 3 |
| drop the `return` (never stop) | 1 |
| `return` on any `OSError` (the original) | 6 |
| WEAKEN the jitter clamp to `min(sleep, timeout)` | 1 |
| WEAKEN the rmdir-branch clamp to `min(0.05, timeout)` | 2 |
| WEAKEN `locking.py`'s clamp to `min(0.1, timeout)` | 1 |
| remove the jitter clamp | 2 |
| remove the rmdir-branch clamp | 2 |
| remove `locking.py`'s clamp | 2 |
| a FLAT `time.sleep(0.05)`, no clamp | 2 |
| a FLAT `time.sleep(0.1)`, no clamp | 2 |
| warn on the FIRST touch failure | 1 |

Every count re-measured at the current head against `test_claude_locks.py` +
`test_locking.py` (control: 40 passed, 0 failed). The `remove locking.py's
clamp` row was absent and the flat-sleep rows are the ones round 5 added: a
flat unclamped sleep used to pass all of them.

A flat `time.sleep(0.005)` still passes and is not claimed. The code's own
constants are 0.05, 0.1 and 0.25, so that is not a shape this can regress
into, and catching it would need a discriminating region narrower than the
scheduling noise these cases run in.

**The three WEAKEN rows are the new ones, and the `remove` rows are why they
had to exist.** Deleting a clamp
was caught; WEAKENING one to the whole `timeout` was not — both guard cases use
a budget smaller than the flat sleep (0.01 s and 0.001 s), where
`min(sleep, timeout)` and `min(sleep, remaining)` are the same number. Measured
before the new cases: both weakenings passed 36/36 while the acquire path
overshot a 0.6 s budget to 1.00 s. At `DEFAULT_TIMEOUT_S = 9.0` it is not a
no-op and not unbounded either: measured, the weakened form's worst is
`9.323 s`, so it reverts to base behaviour bounded by one sleep. The new cases pin the jitter
and read the rmdir branch off the sleep ARGUMENTS, since a 0.05 s overshoot is
not separable from scheduling noise on the wall clock.

`locking.py` carried the same hole and is closed in the same pass: its own
case uses `budget = 0.01`, smaller than the 0.1 flat sleep, so weakening that
clamp left both lock suites at 38 passed while DELETING it was caught — the
file discriminated against deletion and was blind to weakening. The new case
asserts each sleep against the budget left at that moment.

That per-sleep form replaced a `min(slept) < 0.05` shape assert, which failed
on CORRECT code: the shape of the last sleep depends on when the loop arrives,
and with 8-12 ms of per-iteration latency every sleep is a full 0.05. Measured
on pristine source with only `time.sleep` returning late — 0 ms passes, 10 ms
and 50 ms fail. This box has 48 cores; CI is `-n auto` on 2-4. The budget is
deliberately not a multiple of the flat sleep, or the weakened form divides
evenly and survives.

### Scope note: everything in this diff beyond the two headline fixes

Beyond the disarmed heartbeat and the unclamped retry sleep, the range carries:

- **A serialized stale takeover.** `_take_over_stale` decides with `os.stat` and acts with `os.rmdir`; between the two another waiter could create its lock at the same name and have it removed from under it, putting two processes inside the critical section. The window is now serialized on a `FileLock` at `<lock>.takeover`, and the staleness is re-read while holding it. Three such files (`~/.claude/.oauth_refresh.lock.takeover`, `~/.claude.lock.takeover`, `~/.claude.json.lock.takeover`) stay on disk by design: an flock guard has to be one inode for every waiter, and unlinking it after use reopens the race it exists to close. This adds a `claude_locks` → `locking` import. A takeover that finds the directory already gone reports the name as free, so the caller's next mkdir decides without a back-off.
- **A back-off on the swept-name arm.** A lock name that a sweep removed, or a dangling symlink at it, no longer spins the retry loop hot.
- **Two witness tests of behaviour this PR does not change**: `test_the_primary_lock_is_taken_before_the_legacy_one` and `test_the_legacy_lock_carries_the_60s_staleness_too` pin the two-lock protocol the fix sits inside, so a later change there fails loudly.
- **Test plumbing shared, not copied.** The thread-scoped sleep recorder and the scripted crossing clock live in `conftest.py`.

**5 files changed, +1376 / −11** across 61 commits; 58 cases across the two lock test files (27 added in this range). Suite: 2159 passed, 3 skipped.

## Round 5

The `src/` fix was sound; three of the four findings were in the cases that
certify it, and they were wrong in opposite directions at once.

**They failed on correct code.** `start` was taken outside the call, ahead of
`mkdir(parents=True)` and, for `FileLock`, an `open` as well, with a fixed 5 ms
expected to cover the gap. Under a starved runner it did not: **3 natural
false failures in 44 SCHED_IDLE runs of the pristine tree**, worst
`slept 0.050s with -0.095s left` — a correctly clamped sleep reported as an
overrun. Injected thresholds were 3 ms, 5 ms, and 1 ms for the third case,
whose whole budget was 1 ms.

They read the code's own anchor now: the first `monotonic()` inside the call IS
the deadline the clamp is measured against, cleared at the call rather than at
the patch so nothing else in the thread can become it. Measured with a **0.5 s**
stall injected ahead of that anchor, all three pass.

**And they passed a 100x defect.** `time.sleep(0.1)` in place of the clamp
overshot a 0.001 s timeout by 100x and passed all 39 cases, because the
surviving wall-clock ceilings are bounds a smaller flat sleep slides under.
Related, and the reason it went unseen: the anti-weakening guards go blind
under loop latency — detection needs the loop to arrive with less than one
flat sleep of budget left, and 10-50 ms of latency steps over that window
while `n >= 2` still holds. Each case asserts it ENTERED the discriminating
region now, and says so when it did not.

**The toucher's warning claimed something it could not know.** "its mtime stops
advancing, so a waiter may take it over as stale" describes a failure that
outlives the staleness window, and the arm fired on the first one — measured,
7 touches and a lock fresh throughout, under a warning saying the opposite. It
is gated on persistence now, which is what makes the sentence true.


realiti4:
Hi, thanks for th PR, it looks good.

Could we adjust couple of small things before merge though, based on the review pass:

---

**1. Verified.** Ran both test files at head e5fa043 in a clean worktree: 35 passed. Replayed the reversion table's third row (back to `return` on any `OSError`): exactly the 4 claimed tests fail. The three timing-sensitive tests were stable across 10 consecutive local runs, and CI is green on all three platforms. The argument against re-checking the path instead of using the errno also checks out empirically — with one correction, next.

**2. The `Path.exists()` boundary is 3.14, not 3.13.** Tested against real interpreters (child of a mode-000 directory, so `stat` fails with EACCES):

| Python | `Path.exists()` |
| --- | --- |
| 3.12 | raises `PermissionError` |
| 3.13 | raises `PermissionError` |
| 3.14 | returns `False` |

So `claude_locks.py:151` ("…and, from 3.13, an answer that reads absence out of a permission error") and the PR body's table should say 3.14; 3.13 still re-raises like 3.12. The argument itself survives untouched — `requires-python` is `>=3.12` with no upper bound, so both failure modes still exist across supported versions; only the boundary moves.

**3. The `test_locking.py` invariant can fail a correct implementation.** `begin` is anchored before `waiter.acquire()` (`tests/test_locking.py:92`), but `acquire` computes its own deadline only after `mkdir(parents=...)` + `open()` (`locking.py:41-44`). A correct clamp may therefore sleep up to (mkdir + open time) longer than the test's computed `left`, and the tolerance at `tests/test_locking.py:101` is 1ms — on Windows CI, Defender routinely stretches a first-touch `open()` past that. Since the invariant's discriminating power is against the flat 0.1s sleep, widening the tolerance to ~0.05 loses nothing and absorbs the anchor skew.

Same family, softer, take or leave: `tests/test_claude_locks.py:235` asserts `elapsed < 0.03` on a 1ms budget, ~29ms of headroom against Windows' ~15ms scheduler quantum. The ceiling can't simply be raised — the defect it guards produces a flat 0.05s, so it must sit below that — but recording the sleep *argument* (as the `test_locking.py` test already does) would decouple it from wall clock entirely.

**4. Noting for the record, not asks — pre-existing behavior this PR narrows but doesn't (and shouldn't) close:** (a) if a lock is nonetheless stolen and re-created, the toucher freshens the replacement and the exit path's `rmdir` (`claude_locks.py:163`) removes the new holder's lock — upstream proper-lockfile guards this by comparing the mtime it last set; (b) a touch failure that persists past the staleness boundary now retries silently, where a one-time warning would give observability. Both predate this PR (the old code failed the same scenarios faster, by disarming on the first error), and the full compromised-lease model Claude Code's bundle wires into its refresh isn't warranted here: unlike CC, cswap never holds these locks across network calls (`switcher.py` documents "no network while locks are held"), so holds are short local I/O against 10s/60s staleness windows. Follow-up issue, separate from this diff.

codeslake:
Thanks — all three are in at `2a3cfa9`. I reproduced each before changing anything, and two of them turned out to be worse than you described.

## 1. The version boundary is 3.14. You are right, and there is more of it.

Reproduced on three real interpreters, child of a mode-000 directory:

| Python | `Path.exists()` |
| --- | --- |
| 3.12 | `PermissionError` |
| 3.13 | `PermissionError` |
| 3.14 | `False` |

Corrected `claude_locks.py` and the test comment. The enumeration also silently skipped 3.13 while claiming to cover every supported version, so it now names it: *"3.12 re-raises these errnos out of the thread as does 3.13, 3.14+ swallows them"*. Verified per-errno for exactly the three the test parametrises (ESTALE/EACCES/EIO) — all three raise on 3.12/3.13 and all three return `False` on 3.14.

**Three more of these survive in `credentials.py:1173,1176,1606`, and they are mine, not inherited.** `git log -S` on the text lands on `0ccda15` — my own #196, already merged. I first told myself they came from someone else's commit, because I ran `git blame -L <lines>` on a branch where those line numbers point somewhere else entirely. The code there is correct either way (it uses `stat`, which raises on every version); only the prose is wrong. Say the word and I will fix them here, or I will open a separate PR — your call, since they are outside this diff.

## 2. The anchor skew is real, and my first fix broke the test

You are right that a correct clamp can sleep past the test's deadline: `acquire` starts its own deadline after `mkdir(parents=...)` + `open()`, and `begin` is anchored before the call. Measured worst case for that skew across 400 runs: **0.88 ms**, which left only 12% headroom under the old 1 ms tolerance. Genuine flakiness.

I widened it to 0.05 as you suggested and that was too much. Measured:

| clamp mutated to | tolerance 0.001 | tolerance 0.05 |
| --- | --- | --- |
| flat `sleep(0.02)` | FAIL | **PASS** |
| flat `sleep(0.04)` | FAIL | **PASS** |
| flat `sleep(0.1)` | FAIL | FAIL |

So a regression reintroducing a small flat sleep would have shipped green, and my comment claiming *"discrimination is unaffected"* was true only of the one historical defect.

Bounded by `budget` (0.01) instead — that is the most a correct clamp can produce, since its cap is the remaining internal budget and can never exceed `timeout`. Measured max single sleep with the correct clamp: 0.009998 s. Plus one anchor-free assertion that has no skew to absorb at all:

```python
assert seconds <= budget + 1e-6, "a single retry sleep exceeds the whole timeout"
```

Re-measured after: flat 0.011 / 0.02 / 0.04 / 0.1 all fail, correct clamp passes.

## 3. The wall-clock assertion now reads the sleep argument — and the recorder had a second bug

Rewritten as you suggested. Your mechanism note was sharper than mine, too: what would actually blow a 29 ms ceiling is Windows rounding each sub-millisecond sleep up to the ~15.6 ms timer tick, so two iterations alone reach 31 ms. That is in the comment now.

The recorder as first written patched `claude_locks.time.sleep`, which **is** the shared `time` module. Measured: a foreign thread's three `sleep(0.5)` calls all landed in the recorded list — enough to fail correct code with *"the rmdir-failure path slept 0.500s"* — and every foreign sleep was flattened to `real_sleep(0)`, turning a leaked paced loop into a hot spin. `conftest.py` documents that this suite leaves such threads running. Scoped by `threading.get_ident()` now; re-measured, only our own 0.001 is recorded and the foreign thread's pacing stays at [0.05, 0.05, 0.05].

One thing I left alone: `max(slept) <= timeout` is timeout-shaped rather than remaining-budget-shaped, so a mutation to `min(0.05, timeout)` survives it. Making it per-sleep-against-remaining would reintroduce exactly the anchoring gap from item 2 at a 1 ms scale. The comment no longer claims the branch "respects the deadline".

## 4. Agreed, follow-up

Both are pre-existing and this diff cannot have made them worse — the entire `src/` change in this PR is one word inside one comment (`3.13` → `3.14`). Happy to open an issue for the stolen-and-recreated-lock rmdir and the silent persistent-touch retry, with your compromised-lease note attached.

Whole suite at `2a3cfa9`: **2097 passed, 3 skipped**. The two lock files: 35/35 on 3.14 and 35/35 on 3.12, which is what all three CI jobs pin.

Issue #263, opened 2026-08-19
Option to auto-start the 5 hour session when it refreshes


jcalifornia:
### What problem would this solve?

The Claude 5 hour session countdowns don't start until you send/receive the first token under the fresh reset. It would be advantageous to automatically send something in each time the 5 hour session is detected to have reset.

PR #273, opened 2026-08-22
Feature/start five hour timer

jcalifornia:
Implementing feature #263

It sends this user message:

  Reply OK.

  With this system prompt:

  Reply with exactly OK.


AminDhouib:
I independently implemented a standalone alternative in #308 and opened it as a draft so maintainers can compare approaches. The main distinction is lifecycle: #308 provides an opt-in warmup command independent of autoswitch, with one-shot/scheduler/foreground modes. It also fails closed on hollow usage data, keeps inactive accounts in isolated profiles, revalidates exact account identity under locks, and preflights first-party OAuth immediately before sending. Happy to consolidate or adapt based on maintainer preference.

AminDhouib:
Follow-up: #308 now also supports the one-process workflow via a persisted `autoswitch.warmupFiveHour` setting and a TUI Auto-screen toggle, while retaining the standalone command for users who do not want autoswitch.

It also changes the unavailable-usage policy after observing a fleet-wide reset: missing/stale/errored/hollow usage permits one identity-bound request under a full five-hour persistent guard, rather than either skipping forever or sending every poll. Confirmed live windows and known future weekly exhaustion still skip. This makes the overlap with this PR more direct; maintainers can choose or combine the preferred implementation.

PR #276, opened 2026-08-23
Add fish shell completions for cswap

onhazrat:
## Summary

Adds fish completions for the CLI: `completions/cswap.fish`, plus a two-line
`completions/claude-swap.fish` so the long-form console script gets them too.

Covered:

- every subcommand — `list`/`ls`, `status`, `switch`, `add`, `add-token`, `remove`/`rm`,
  `disable`, `enable`, `run`, `map`, `unmap`, `alias`, `swap`, `move`, `auto`, `config`,
  `unclaimed`, `export`, `import`, `tui`, `watch`, `menubar`, `upgrade`/`update`, `purge`
- each subcommand's flags, with descriptions and value candidates taken from the CLI
  help (`--strategy best|next-available`, `--model`, `--include-api-key-accounts`, …)
- account targets complete by **number, email, and alias**, labelled with the active
  marker — `switch`, `remove`, `disable`, `enable`, `run`, `map`, `alias`, `swap`, `move`
- `config get/set/unset` complete the keys from `SETTING_SPECS`, and `config set`
  completes values for the closed-set keys (`ui.theme`, `autoswitch.strategy`,
  `autoswitch.includeApiKeyAccounts`)
- positions are respected: `alias <acct> <NAME>` offers nothing for the new name,
  `move <acct> <SLOT>` offers only slot numbers, `map`/`unmap` complete directories,
  `export`/`import` complete file paths, and everything else suppresses file noise

Also adds a short "Shell completions" section to the README under Installation.

## One design note worth your call

Account candidates are read from the local `sequence.json`, **not** from
`cswap list --json`.

`list` is an on-demand usage caller: entries older than `SERVE_TTL_S` (180s) get
refetched over HTTPS with a 5s per-account timeout. Driving completions from it meant a
Tab press could block the prompt for seconds — and a stale cache is the common case,
since you don't run `cswap` every three minutes. Reading `sequence.json` is ~40x faster
on the warm path and never touches the network:

```
20 iterations, local read:        0.073s
20 iterations, cswap list --json: 3.041s   (warm cache; cold adds network round-trips)
```

The tradeoff is that the completion now depends on the on-disk shape of `sequence.json`
(number / email / alias) rather than the versioned `--json` contract. If you'd rather it
went through a supported interface, I'm happy to switch it to a hidden plumbing
subcommand (e.g. `cswap __complete accounts`) that reads local state only — just say the
word and I'll add it with tests.

Path resolution follows `paths.get_backup_root()`: `$XDG_DATA_HOME/claude-swap` (ignored
when relative, per the XDG spec), then `~/.local/share/claude-swap`, then
`~/.claude-swap-backup`.

## Testing

fish 4.8.1 on macOS, driving real completions with `complete -C` against a live install
across every subcommand and flag context. Checked specifically:

```
> complete -C "cswap switch "
1                        arefe.rohani@gmail.com [aref] (active)
2                        hazrati.dev@gmail.com [me]
aref                     #1 arefe.rohani@gmail.com (active)
...

> complete -C "cswap auto --strategy "
best           Account with the most quota left (default)
consume-first  Account whose weekly window resets soonest

> complete -C "cswap config set ui.theme "
auto
dark
light
```

Degradation cases, all of which yield no candidates and no error rather than breaking
the prompt: no `sequence.json` (fresh install), empty file, truncated JSON, binary
garbage, and a relative `XDG_DATA_HOME`. Verified autoload from a clean
`XDG_CONFIG_HOME` (not just an explicit `source`), and that `claude-swap` inherits the
set via `--wraps`.

`completions/` sits outside `[tool.hatch.build.targets.wheel]` (`src/claude_swap`), so
packaging is unaffected, and CI is Python-only so nothing new runs there.

The awk parser sticks to POSIX (`[ \t]` rather than `[[:space:]]`, which older mawk
handles inconsistently); it was exercised against BSD awk locally, so a sanity check on
gawk/mawk would be welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,18 @@ uv tool upgrade claude-swap
 pipx upgrade claude-swap
 ```
 
+### Shell completions
+
+Fish completions live in [`completions/cswap.fish`](completions/cswap.fish) — every
+subcommand and flag, plus account numbers/emails completed live from `cswap list --json`.
+
+```bash
+curl -o ~/.config/fish/completions/cswap.fish \
+  https://raw.githubusercontent.com/realiti4/claude-swap/main/completions/cswap.fish
+```
+
+(fish picks up new files under `~/.config/fish/completions/` automatically, no restart needed)
+
 ## Usage
 
 ### Add your first account
--- a/README.md
+++ b/README.md
@@ -36,15 +36,19 @@ pipx upgrade claude-swap
 
 ### Shell completions
 
-Fish completions live in [`completions/cswap.fish`](completions/cswap.fish) — every
-subcommand and flag, plus account numbers/emails completed live from `cswap list --json`.
+Fish completions live in [`completions/`](completions) — every subcommand and flag,
+plus account numbers, emails and aliases completed from your local account list, and
+setting keys and values for `cswap config`.
 
 ```bash
-curl -o ~/.config/fish/completions/cswap.fish \
-  https://raw.githubusercontent.com/realiti4/claude-swap/main/completions/cswap.fish
+base=https://raw.githubusercontent.com/realiti4/claude-swap/main/completions
+curl -o ~/.config/fish/completions/cswap.fish $base/cswap.fish
+curl -o ~/.config/fish/completions/claude-swap.fish $base/claude-swap.fish
 ```
 
-(fish picks up new files under `~/.config/fish/completions/` automatically, no restart needed)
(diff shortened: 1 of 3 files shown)


onhazrat:
Superseded by #277, which rebases this onto a conventionally-named branch (`feat/fish-completions`), squashes to one `feat(completions):` commit, and adds `tests/test_completions_contract.py` — a two-layer guard for the `sequence.json` fields the completion parses. Closing in favour of that one.

PR #278, opened 2026-08-25
fix(menubar): stop leaking a menu of items on every refresh

fantasywind:
## Problem

The menu bar leaks memory steadily while idle. A real instance reached **1.28 GB RSS over about seven days** at a 30s refresh — roughly **186 MB/day**.

`rumps` records every `MenuItem` carrying a callback in a process-global dict and never removes it:

```python
# rumps/rumps.py:579
NSApp._ns_to_py_and_callback[self._menuitem] = self, callback
```

`Menu.clear()` only empties the `NSMenu`:

```python
# rumps/rumps.py:276
def clear(self):
    self._menu.removeAllItems()
    super(Menu, self).clear()
```

`rebuild_menu()` ran on every refresh tick and recreated all ~38 items each time, so every tick stranded a menu's worth of items — and the closures holding the app — for the life of the process.

## Fix

The menu's variable content is rendered as plain data first (`MenuModel`), so a tick can decide what to do without touching rumps:

- **nothing** when the model is unchanged,
- **in-place relabel** when only text or check-marks moved (the common case: percentages ticking),
- **full rebuild** only when items must be created or destroyed.

`MenuModel.shape` carries the **account slot tuple**, not a row count. Callbacks bind to a slot at creation, so relabelling across a changed slot set would leave a row switching to whichever account previously occupied that position.

`rebuild_menu()` unregisters what it discards, so the now-rare structural rebuild does not leak either. `_model` is cleared before either update path and restored only on success: the old code rebuilt unconditionally and therefore self-healed after a transient failure, and that property had to be preserved rather than traded away.

Falling out of this:

- `_make_toggle_disabled` captured the row's `disabled` flag at build time. A row now outlives a state change, so it reads the flag when clicked.
- `_history_menu` read the entire log on every rebuild to show ten lines. It now reads a tail sized to the rotation ceiling — a smaller window silently truncates the history, because switch lines are sparse among usage lines (a 64KB window found 2 of 10 entries on a real 514KB log).
- Repeated history entries collapse before the limit is applied. Two identical switches inside one minute render as the same string; rumps identifies items by title and drops the duplicate, so a model listing one would not match the menu it produces.

## Incidental fix

`_history_menu` used `read_text(encoding="utf-8")` under `except OSError`. `UnicodeDecodeError` is a `ValueError`, not an `OSError`, so a single bad byte in the log would have taken down `rebuild_menu()` entirely. The tail read decodes with `errors="replace"`.

## Verification

Driven against real `rumps`/`NSMenuItem` objects the way production does it (new snapshot → `sync_menu()`), 1000 refreshes:

| | registry growth | RSS growth | projected/day @30s |
|---|---:|---:|---:|
| before | 38,000 entries | 60.5 MB | ~174 MB |
| after | **0** | 1.9 MB | ~5.5 MB |

The pre-fix projection matches the ~186 MB/day seen on the real instance, which is what confirms this was the dominant leak. 200 forced structural rebuilds grew the registry by 3 entries — the extra slot's rows — not 200 × 38.

Functional checks read titles and check-marks back off the live `NSMenuItem` objects rather than off the model, since silent staleness is the risk of updating instead of rebuilding.

36 new tests. The suite still never imports rumps or AppKit — a stand-in is injected through `sys.modules`, including a separator that is not a `MenuItem` and has no `values()`, which is what makes the type check in the registry walk load-bearing.

The tests were mutation-checked: 14 of 16 mutations to the fix are caught. The two that survive are `strict=True` on the `_apply_model` zips and the `_items` arm of the registry walk — both are defence in depth whose value is that they never fire, and pinning them would mean first breaking the invariant that makes them unnecessary. Recorded here rather than papered over.

Full suite: 2122 passed. Three failures in `test_move_accounts.py`, `test_real_store_guard.py` and `test_swap_accounts.py` pre-date this branch — they reproduce on a clean `main`, and none of those files import `menubar`.

## Upstream note

`Menu.clear()` leaving `_ns_to_py_and_callback` populated is arguably a rumps bug: any app that periodically rebuilds its menu hits it. This works around it locally rather than waiting on a rumps release.

### DIFF
--- a/src/claude_swap/logging_config.py
+++ b/src/claude_swap/logging_config.py
@@ -4,6 +4,10 @@ import logging
 from logging.handlers import RotatingFileHandler
 from pathlib import Path
 
+# Rotation ceiling for claude-swap.log. Named because the menu bar reads a
+# tail of this file and has to size its window against the same number.
+LOG_MAX_BYTES = 1024 * 1024
+
 
 class _LazyDirRotatingFileHandler(RotatingFileHandler):
     """RotatingFileHandler that creates its parent dir on first emit.
@@ -44,7 +48,7 @@ def setup_logging(log_dir: Path, debug: bool = False) -> logging.Logger:
     log_file = log_dir / "claude-swap.log"
     file_handler = _LazyDirRotatingFileHandler(
         log_file,
-        maxBytes=1024 * 1024,  # 1MB
+        maxBytes=LOG_MAX_BYTES,
         backupCount=3,
         delay=True,
     )
--- a/src/claude_swap/menubar.py
+++ b/src/claude_swap/menubar.py
@@ -28,6 +28,7 @@ from datetime import datetime, timezone
 from pathlib import Path
 
 from claude_swap import pace
+from claude_swap.logging_config import LOG_MAX_BYTES
 from claude_swap.exceptions import ClaudeSwitchError, CredentialReadError
 from claude_swap.switcher import SENTINEL_NOTES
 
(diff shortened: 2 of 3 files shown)


fantasywind:
Superseded upstream: `2df435d` ("fix(menubar): purge rumps callback registry on rebuild to stop unbounded memory growth", merged via #271) fixes the same `NSApp._ns_to_py_and_callback` leak this PR addressed (this PR took the in-place-refresh approach instead). Closing in favor of the merged fix. Thanks!

PR #280, opened 2026-08-25
fix: pin Keychain trusted-app ACL on every credential write

### LINKED ISSUE #279: security add-generic-password -U resets Keychain ACL to writer-only on every credential rotation, locking Claude Code out
On macOS, `set_password()` in `macos_keychain.py` calls `security add-generic-password -U` without `-T`. Per `security`'s own semantics, omitting `-T` on a write does not mean "leave the existing ACL alone", it means "trust only the process performing this write" (`/usr/bin/security` itself). That's fine for this module's own reads (creator == reader == `security`, by design), but Claude Code reads the same `"Claude Code-credentials"` Keychain item **in-process** via its own Security.framework call, a different, untrusted app.

The result: every `cswap` rotation silently re-pins the item's ACL to "security only." Claude Code's next in-process read then requires a GUI "wants to use your keychain" prompt. In a headless or `launchd` context there's no GUI to answer that prompt, the read fails, and Claude Code reports "OAuth session expired and could not be refreshed", even though the underlying credential (and its `expiresAt`) is completely valid.

**Fix:** pass explicit `-T` for `/usr/bin/security` (needed because specifying `-T` at all drops the implicit "trust the calling process" default, so it has to be listed to keep this module's own future reads working) and for the resolved `claude` binary (trust anchored to code signature via macOS's normal trusted-app resolution, so it survives `claude` self-updates without needing a version-pinned path). Applies to both the `security -i` stdin path and the argv fallback for oversized payloads.

Companion change: a best-effort warning when an OAuth payload about to be written is missing `claudeAiOauth.expiresAt`. A missing `expiresAt` makes Claude Code treat the token as already-expired, which produces the exact same "session expired" symptom and is easy to misdiagnose as a recurrence of the ACL bug above. Diagnostic only, never blocks the write.

PR incoming from a fork branch, will link it here once opened.

blaineyeager:
Fixes #279

`set_password()` calls `security add-generic-password -U` unconditionally, with no `-T`. Per `security`'s semantics that resets the item's ACL to trust only `/usr/bin/security` on every write, not "leave the ACL alone." Claude Code reads the same Keychain item in-process via Security.framework, a different app, so every `cswap` rotation silently locked it out. The next read needed a GUI keychain prompt that never arrives in a headless/launchd context, surfacing as "OAuth session expired and could not be refreshed" despite a valid credential.

This adds `_trusted_app_paths()` to `macos_keychain.py`, granting explicit `-T` to `/usr/bin/security` (has to be listed once any `-T` is passed, since that drops the implicit trust-caller default) and to the resolved `claude` binary (`PATH` first, falling back to `~/.local/bin/claude`), on both the stdin and argv-fallback write paths in `set_password()`.

Also adds `_warn_if_oauth_missing_expires_at()` in `credentials.py`: a best-effort log (never raises) when an OAuth payload about to reach the Keychain is missing `claudeAiOauth.expiresAt`, since that produces the identical "session expired" symptom and is easy to misdiagnose as a recurrence of the ACL bug.

Tested: `uv run pytest`, 2086 passed, 3 skipped, 3 failed. The 3 failures are pre-existing on clean `main` (reproduced with these changes stashed), a `chmod 0o000`-based permission-guard test that doesn't hold under the sandbox's effective uid, unrelated to this change.

### DIFF
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -951,6 +951,36 @@ class CredentialStore:
         except OSError as e:
             self._host._logger.warning(f"Failed to remove credentials file: {e}")
 
+    def _warn_if_oauth_missing_expires_at(self, credentials: str) -> None:
+        """Log (never raise) if the OAuth payload about to reach the Keychain is
+        missing ``claudeAiOauth.expiresAt``.
+
+        ``macos_keychain.set_password`` hex-round-trips ``credentials`` byte for
+        byte — it cannot itself drop a field. This is a pass-through sanity check
+        one layer up: an ``expiresAt``-less credential has been observed reaching
+        the active Keychain item while the ``.credentials.json`` file copy was
+        valid, meaning whatever upstream call built ``working``/``rollback_creds``
+        for this write occasionally supplies a stale or partial lineage. Surfacing
+        that here, at the single choke point every OAuth write passes through,
+        beats tracing every producer — and a missing ``expiresAt`` makes Claude
+        Code treat a token as already-expired (``is_oauth_token_expired`` prefers
+        expired-on-error), which reads exactly like the "session expired" reports
+        this was written to explain. Best-effort: a malformed (non-JSON) payload
+        is `CredentialWriteError`'s problem, not this check's — swallow and move
+        on rather than blocking the write over a diagnostic.
+        """
+        try:
+            data = json.loads(credentials)
+            oauth = data.get("claudeAiOauth") or {}
+            if not oauth.get("expiresAt"):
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for the PR. I checked Claude Code’s macOS 2.1.263 binary: its credential reads and writes use `security`, so granting the Claude executable access isn’t needed. #324 also confirms that plain `-U` updates preserve the existing ACL, while adding `-T` can cause authorization failures or timeouts. I’m closing this PR; we can continue investigating the reported session-expired symptom in #279.

PR #282, opened 2026-08-25
fix: don't manufacture a dead-token verdict from an unreadable backup

### LINKED ISSUE #281: _entry_token_dead reports "re-login needed" on a healthy active account when its backup Keychain read merely fails
`_entry_token_dead`'s active-slot second-source check (`switcher.py:4801-4812`) returns `True` (dead) whenever `_read_account_credentials_ex` fails to read the slot's backup Keychain item, lock contention, timeout, any transient `security` failure, with no check on whether `entry.auth_dead_strikes` has ever actually reached `AUTH_DEAD_STRIKES`. The branch's own comment justifies holding a strike because the row "already took AUTH_DEAD_STRIKES invalid_grants", but the code never verifies that premise: a healthy account with zero recorded strikes shows "re-login needed" purely because one Keychain read hiccuped.

Reproduced live: account 2 (active) sat at `authDeadStrikes: 0` in the persisted usage store the entire time, yet `cswap list --token-status` reported "re-login needed — refresh token dead" for it, while the same command's "active profile" line reported the live credential as fresh with a valid refresh token.

Existing coverage in `TestUnreadableBackupIsNotAbsent` only exercises this branch with strikes pre-seeded to threshold before the unreadable read, the zero-strikes case was never tested.

Fix: gate the unreadable branch on `entry.token_dead()` (no `stored_fp`, since the unreadable second source can't be fingerprint-verified) instead of returning `True` unconditionally.

PR incoming from a fork branch.

blaineyeager:
## Problem
`_entry_token_dead`'s active-slot check treats a merely-*unreadable* backup Keychain item as proof of a dead refresh token, regardless of whether any `invalid_grant` strike was ever recorded. A healthy active account can show "re-login needed" from nothing more than Keychain read contention.

## Fix
One-line change: the `unreadable` branch now returns `entry.token_dead()` instead of `True`, asking whether the row has genuinely reached `AUTH_DEAD_STRIKES` rather than assuming it has.

## Testing
TDD: added `test_active_slot_with_zero_strikes_and_unreadable_backup_is_not_dead`, watched it fail against the unpatched code, applied the fix, confirmed green. `test_unreadable_backup_never_erases_a_live_dead_token_strike` (the existing test covering the "hold a real strike" case) still passes unchanged. Full suite: 2087 passed, 3 skipped, 3 pre-existing failures unrelated to this change (unchanged before/after).

Fixes #281

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -4809,7 +4809,15 @@ class ClaudeAccountSwitcher:
             # strike costs one pass of "re-login needed" on a row that
             # already took AUTH_DEAD_STRIKES invalid_grants; erasing it costs
             # the quarantine itself.
-            return True
+            #
+            # Gated on an actual strike existing (``entry.token_dead()``, no
+            # ``stored_fp`` — we can't verify which generation, so this only
+            # asks whether the count itself has reached threshold): the
+            # first check above already proved the LIVE credential doesn't
+            # carry the struck generation, so a row with zero strikes has
+            # nothing to hold — an unreadable backup on an otherwise-healthy
+            # account must not manufacture "re-login needed" out of nothing.
+            return entry.token_dead()
         return bool(backup) and entry.token_dead(
             stored_fp=oauth.credential_fingerprint(backup)
         )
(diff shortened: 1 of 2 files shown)


realiti4:
Thank you for the PR and the fix! Looks good merged it.

PR #284, opened 2026-08-25
fix: rescue a successful retry read instead of discarding it in the usage sentinel

### LINKED ISSUE #283: _static_usage_sentinel discards a successful retry and reports "no credentials" for a slot with a valid backup
### Summary
`_static_usage_sentinel` (switcher.py) distinguishes "keychain unavailable" from "no credentials" for an idle slot by re-reading the backup via `_read_account_credentials_ex`, but only checks whether that retry *failed* (`unreadable`), never whether it *succeeded* and found real data. When the earlier read in `_build_accounts_info` hit a transient hiccup (silently swallowed to `""`, since that call passes no `failed` list), and the sentinel's own retry moments later succeeds and finds a real, valid credential, the retry's value is discarded and the sentinel still reports `USAGE_NO_CREDENTIALS`.

### Impact
`cswap list --token-status` (and the TUI) can show the "no credentials" sentinel for an idle slot whose backup is present and healthy, nudging the user to unnecessarily re-add/re-login an account that's fine.

### Reproduction (field case)
Observed live: an idle slot showed "no credentials" in `cswap list --token-status` while a plain `security find-generic-password -a account-<n>-<email> -w -s claude-swap` read of the exact same Keychain item, moments later, returned valid JSON (`claudeAiOauth.accessToken`/`refreshToken` present, `expiresAt` valid ~24h out).

### Root cause
```python
if not is_active and self._read_account_credentials_ex(
    str(num), email
)[1]:
    return USAGE_KEYCHAIN_UNAVAILABLE
return USAGE_NO_CREDENTIALS
```
Only `[1]` (the `unreadable` boolean) is inspected; `[0]` (the retried value) is thrown away even when it's non-empty and contains a usable OAuth token.

### Fix
PR incoming from a fork branch.

blaineyeager:
### What
`_static_usage_sentinel` re-reads an idle slot's backup (`_read_account_credentials_ex`) to distinguish "keychain unavailable" from "no credentials", but only checked whether that retry failed, never whether it succeeded and found real data. Fixes #283.

### Why
A transient hiccup on the EARLIER read in `_build_accounts_info` (silently swallowed to `""`, since that call passes no `failed` list) followed by a clean retry moments later still reported "no credentials" and discarded the retry's valid credential, sending the user to re-add a slot that already has one. Confirmed as a field case: `cswap list --token-status` showed "no credentials" for an inactive slot while a plain `security -w` read of the same item, moments later, returned valid JSON.

### Change
- `_static_usage_sentinel`: capture the retry's `(value, unreadable)`; if not unreadable and the value is real (API-key shaped, or has an extractable OAuth access token), use it, returning `None` so the collect pass proceeds to fetch, instead of falling through to `USAGE_NO_CREDENTIALS`.
- `_build_accounts_info`: inactive-slot read now goes through `_read_account_credentials_ex` (like every other backup read site in this module) instead of the bare `_read_account_credentials`, so a transient failure is at least distinguishable at the read site.

### Tests
- New: `TestBackupUnreadableDisplay::test_idle_slot_stale_empty_creds_but_retry_finds_valid_backup`, reproduces the bug (red without the fix, green with it).
- New: `TestBuildAccountsInfoReadRouting::test_inactive_slot_read_uses_the_ex_variant`, pins the secondary read-routing change.
- Full suite: `uv run pytest -q`, 2089 passed, 3 skipped, 3 pre-existing unrelated failures (macOS sandbox chmod tests, fail identically on main).

### DIFF
--- a/src/claude_swap/switcher.py
+++ b/src/claude_swap/switcher.py
@@ -3750,7 +3750,16 @@ class ClaudeAccountSwitcher:
                 creds = active.value or ""
                 self._record_active_verdict(active)
             else:
-                creds = self._read_account_credentials(str(num), email)
+                # `_ex` (not the bare reader): a transient Keychain failure
+                # on this read must not be silently indistinguishable from
+                # a genuinely absent backup — the bare variant takes no
+                # `failed` list and always swallows to "". This call site's
+                # `unreadable` verdict isn't threaded further today (the
+                # sentinel does its own independent re-read), but reading
+                # through `_ex` here matches every other backup read site
+                # in this module and keeps the door open without widening
+                # this tuple's shape.
+                creds, _unreadable = self._read_account_credentials_ex(str(num), email)
 
             accounts_info.append((num, email, org_name, org_uuid, is_active, creds, alias))
         return accounts_info
@@ -4469,9 +4478,7 @@ class ClaudeAccountSwitcher:
                 self._active_keychain_unavailable or self._active_read_unreadable
             ):
                 return USAGE_KEYCHAIN_UNAVAILABLE
-            if not is_active and self._read_account_credentials_ex(
-                str(num), email
-            )[1]:
+            if not is_active:
                 # THIS slot's own read, not the process flag — which one
                 # slot's clean read erased for every other slot. Measured with
                 # every read denied and a real backup on slot 2:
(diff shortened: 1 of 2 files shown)


realiti4 (changes requested):
Thanks for this one.

Could you check the findings below from review agent please:

---

**Handoff gap (blocker)**

`_static_usage_sentinel` returns `None` after the retry finds a valid OAuth blob, but `accounts_info` is never updated. `_collect_usage_entries` then calls `_fetch_account_usage(info)` with the original row, which unpacks `creds = ""` and reaches `try_fetch_usage_for_account(..., "")` → `error="no-access-token"` → `consecutive_failures=1` + backoff. Traced with a probe on this branch: seeded the fake Keychain with a valid backup, passed `(2, email, "", "", False, "", "")`, captured the fetch's `creds` argument — got `''`. So the slot goes from "no credentials" to a bogus failed fetch with backoff, which reads as an auth/network problem instead of a missing backup. `_entry_token_dead` and `_token_status_lines` also still see `""`.

The API-key arm is fine since `USAGE_API_KEY` is terminal.

**Test gap**

`test_idle_slot_stale_empty_creds_but_retry_finds_valid_backup` only checks the sentinel result `!= USAGE_NO_CREDENTIALS`, so it passes with the handoff broken. The test that would catch it drives `_collect_usage_entries` / `_fetch_account_usage` and asserts the fetch receives the retried credential.

**Suggested shape**

Retry once at the source instead: in `_build_accounts_info`, if the inactive-slot read comes back empty, read again and put the real value in the tuple. Every consumer (sentinel, dead-token check, token-status lines, fetch) then sees the right creds with nothing threaded through, and the sentinel rescue block, the `_ex` routing change, and `TestBuildAccountsInfoReadRouting` all become unnecessary. Smaller diff than this one.

**One question**

What did the `Failed to read credentials from Keychain: …` warning actually say in your case? A retry helps a transient flake; if the login keychain is locked (headless/launchd), it won't, and that'd point at a different fix across #279/#281/#283.

nakpapa:
#284

nakpapa (approved):
Onal mau negacat motor  sama rumah ibu rojali normal kaya dulu,manjain onal,nanti nal kasih take and give yang gak di sangka2

nakpapa:
Gas ung,biar onal punya abang yang dulu,kaya genta waktu kecil,nal mau lomba terus

PR #291, opened 2026-08-27
test: three cases assert a Linux store layout, and the macOS job never ran them

codeslake:
## The blind spot first

The macOS job runs two files:

```
uv run pytest tests/test_macos_keychain_contract.py tests/test_macos_keychain.py
```

Everything else in the suite has never executed on macOS in CI. Three cases
have been failing there the whole time, with every check green.

Measured on a real Mac at `main`, whole suite:

```
FAILED tests/test_move_accounts.py::TestMoveUnreadableSourceIsNotAbsent::test_unreadable_enc_aborts_the_move_before_anything_changes
FAILED tests/test_real_store_guard.py::test_c0_a_scratch_home_still_protects_the_os_account_home_store
FAILED tests/test_swap_accounts.py::TestSwapUnreadableSourceIsNotAbsent::test_unreadable_enc_aborts_the_swap_before_anything_changes
```

## What each one assumed

**`test_c0`** spelled the expected store root out as
`~/.local/share/claude-swap`. That is the XDG layout, and `get_backup_root()`
returns it on Linux/WSL only — on macOS it IS the legacy
`~/.claude-swap-backup`, so the case asserted a directory that is never a root
there:

```
assert PosixPath('/Users/…/.local/share/claude-swap') in [
    PosixPath('…/scratch-home/.claude-swap-backup'),
    PosixPath('/Users/…/.claude-swap-backup'),
    PosixPath('/Users/…/.claude'),
    PosixPath('/Users/…')]
```

Both expectations are now DERIVED from `paths` under the HOME each snapshot
resolves against, which is the rule the freeze itself follows. Hardcoding one
platform's answer is what made a per-platform fact look like a defect.

**The two `unreadable .enc` cases** make a backup unreadable with `chmod`. On
macOS a usable Keychain takes the write and reconciles the `.enc` away, so
there is no file to chmod:

```
FileNotFoundError: [Errno 2] No such file or directory: '…/credentials/.creds-2-account2@example.com.enc'
```

The conflation under test belongs to the FILE reader (`""` meaning both "no
backup" and "unreadable right now"), so each class pins the store to file mode
on every platform. A `skipif` would leave the one platform with two backends
untested against that defect, which is the opposite of what this PR is for.

## The job

Widened to the whole suite. `timeout-minutes` goes 5 → 15 for the wider
selection; `faulthandler_timeout=60` stays, for the same reason its comment
already gives.

The job id stays `macos-keychain` although it now runs everything: it is the
name branch protection matches on, and renaming it silently drops the required
check. Rename it if that is not a concern on your side.

## Verification

Whole suite on a real Mac (Apple silicon, Python 3.14, APFS) at this branch:

```
2116 passed, 3 skipped, 3 warnings
```

and on linux, unchanged at `2116 passed, 3 skipped`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,8 +65,9 @@ jobs:
 
       - name: Run the whole suite (not only the Keychain files)
         # THE NARROW SELECTION WAS A BLIND SPOT, not a saving. Three cases
-        # assert a Linux store layout and failed on macOS while every job
-        # stayed green, because this job never ran them. A merge that
+        # failed on macOS -- one asserting the Linux XDG layout, two because a
+        # usable Keychain takes the write and leaves no `.enc` to chmod --
+        # while every job stayed green, because this job never ran them. A merge that
         # restores a two-file selection here re-opens that hole invisibly:
         # those cases pass now, they would simply stop running.
         #
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,16 +71,19 @@ jobs:
         #
         # faulthandler dumps every thread's stack past the budget, which is
         # the only clue when a `security` call blocks on an invisible
-        # SecurityAgent dialog. Merged, a sibling sets
-        # `faulthandler_exit_on_timeout` repo-wide and this step overrides
-        # only the budget, so the dump also ABORTS the worker. Which backstop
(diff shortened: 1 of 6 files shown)


realiti4:
Thanks for the PR and tests! Merged it.

PR #294, opened 2026-08-28
feat(session): share the plugin store across profiles with --share-plugins

lbagic:
## Summary

Adds an opt-in `--share-plugins` flag to `cswap run`: the session profile's `plugins/` directory (the install store — installed plugins, marketplaces, per-plugin data) is symlinked to `~/.claude/plugins`, so a plugin installed or updated in any account is available in all of them. This closes the gap where `enabledPlugins` already travels with `settings.json` under plain `--share`, but the profile has no copy of the store those entries point at.

## Design

Follows the `--share-history` precedent throughout:

- POSIX-only; `run()` rejects the flag on Windows (copies would fork the store).
- A profile's pre-existing store is merged into `~/.claude` first, never discarded: additive union with source-wins conflicts (plugin auto-update converges versions anyway). Per-scope entries (user vs project@path) under the same name are unioned by scope identity, so a project-scoped install in the profile survives a user-scoped one in the source.
- The merge is registry-driven (`installed_plugins.json`, `known_marketplaces.json`) with adopted entries' content copied under the shared store and their `installPath`/`installLocation` re-pointed. Copies are staged and renamed into place, so an interrupted merge resumes safely. External directory marketplaces keep their paths verbatim.
- A stale share manifest can never delete a real store — only symlinks are ever unlinked, same as history.

One hazard is new relative to history sharing: Claude running in a shared profile records registry paths spelled through its own `CLAUDE_CONFIG_DIR` (`<session>/plugins/...`), which resolve only while the link exists. Every launch re-spells those entries at the source, and toggle-off heals before unlinking, so removing the link cannot orphan registry entries for every other account. Registry writes are serialized by a dedicated file lock (the backup-dir lock is non-reentrant and `_sync_sharing` runs both under and outside it).

## Testing

- 23 new tests covering the merge (fresh source, populated source, per-scope union, external marketplaces, data/ merge), the heals (steady-state, toggle-off, heal-failure keeps the link), fail-open paths (lock timeout, corrupt source registry, live profile, foreign symlink, non-directory), path-escape hardening, resumed/interrupted merges, and Windows behavior.
- Registry shape assumptions validated against a real `~/.claude/plugins` store (v2 `installed_plugins.json` with per-name scope-entry lists, `known_marketplaces.json` with `installLocation`).
- Full suite passes except three failures that also fail on `main` in this environment (unrelated files).

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,17 +125,19 @@ cswap run 2                     # launch Claude Code as account 2, here only
 cswap run user@example.com      # by email
 cswap run 2 -- --resume         # everything after '--' is forwarded to claude
 cswap run 2 --share-history     # share your chat history with this account too
+cswap run 2 --share-plugins     # share your installed plugins too
 cswap run 2 --require-session   # refuse rather than run plain claude if 2 is the default login
 ```
 
-Sessions use your normal `~/.claude` setup (settings, CLAUDE.md, skills, MCP servers, etc.), but each account keeps its own chat history — pass `--share-history` if you want your accounts to continue the same conversations.
+Sessions use your normal `~/.claude` setup (settings, CLAUDE.md, skills, MCP servers, etc.), but each account keeps its own chat history and plugin installs — pass `--share-history` if you want your accounts to continue the same conversations, and `--share-plugins` if a plugin installed once should be available everywhere.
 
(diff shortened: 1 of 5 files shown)


lbagic:
Tracking issue: #304 — filed so the problem can be discussed separately from this patch.

PR #297, opened 2026-08-29
Add --require-session to cswap run

GrahamCampbell:
When the account named in `cswap run` is already the active default login, the fast path launches plain claude on that login instead of creating a session profile, because a second copy of the active credential would drift. That is the right default, but a wrapper that hands each terminal to one account and relies on session isolation has no way to notice it happened: the launch silently lands on the default login, and a later `cswap switch` pulls that login out from under the running session.

This adds a `--require-session` flag that turns the fast path into a refusal with a clear message, so such wrappers can guarantee they only ever run in a session profile. Nothing changes without the flag.


realiti4:
Thanks for this! It looks good merged it.

PR #298, opened 2026-08-29
Adopt an exited session's credential into its backup before it is activated or refreshed

GrahamCampbell:
Claude rotates the OAuth token family inside a session profile, and nothing ever synced it back, so once a session has refreshed even once the slot's stored backup holds a consumed generation. The usage collector already works around that by fetching with the profile's credential, which is exactly why the account keeps looking healthy right up until `cswap switch` activates the backup: its first refresh gets `invalid_grant`, Claude Code wipes the login, and the slot is struck as needing a re-login. The collector had the same landmine on its expired-profile path, where it fell back to POSTing the dead backup grant. In practice a switch onto an account with a session running took the default login down within two minutes.

This adds adoption of a quiescent session profile's credential into the slot backup whenever the profile is a newer generation of the same family, decided and written under cswap's own lock and without the invalidation a normal backup write triggers, since the profile is the source rather than a stale seed. Generations are ordered by access-token issue time so a fresh re-login in the backup is never mistaken for drift, and profiles that were re-pointed at another account or flagged stale are left alone. The switch adopts before activating and the collector adopts before taking the idle refresh path, so the common case simply works. While a session is still live its family is claude's to rotate, so the switch now refuses to activate a backup the profile has already rotated past rather than warning and installing a credential that can only fail; the existing warning stays for the case where both copies are still on the same generation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,8 @@ cswap run 2 --share-history     # share your chat history with this account too
 
 Sessions use your normal `~/.claude` setup (settings, CLAUDE.md, skills, MCP servers, etc.), but each account keeps its own chat history — pass `--share-history` if you want your accounts to continue the same conversations.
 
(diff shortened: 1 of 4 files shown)


realiti4:
Thanks for addressing this and for the PR. Merged it.

PR #299, opened 2026-08-29
Ignore session records whose pid has been recycled

GrahamCampbell:
A session record is written by the claude that started it and outlives a crash, and `scan_sessions` only checked that the recorded pid was still alive. Once the OS hands that number to an unrelated process, the record reads as a live session indefinitely: the account's profile can never be re-bootstrapped or removed, and every guard that asks whether a claude is running against it answers yes.

Claude Code stamps each record with `procStart`, its own reading of `ps -o lstart=` for its process under `LC_ALL=C TZ=UTC`. The scan now takes the same reading of the live process and drops the record when that process started after the recorded one, which is the one thing a recycled pid cannot avoid. On Linux, `ps` builds every start time from the boot time in `/proc/stat`, and the kernel moves that with each wall-clock step, so after a WSL2 resume re-syncs the clock a live session can read as younger than its own record. A record is therefore only dropped when the process at the pid is also not a claude; a pid genuinely recycled by another claude lingers for that process's lifetime, which is today's behaviour. Records without `procStart`, Windows, and any failure to read the process keep today's behaviour too, since not knowing must never turn a live session into nobody there.

### DIFF
--- a/src/claude_swap/process_detection.py
+++ b/src/claude_swap/process_detection.py
@@ -7,9 +7,11 @@ currently running. Uses the same mechanism Claude Code itself uses internally.
 
 from __future__ import annotations
 
+import calendar
 import json
 import logging
 import os
+import subprocess
 import sys
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -17,6 +19,18 @@ from pathlib import Path
 from claude_swap.paths import get_claude_config_home
 
 logger = logging.getLogger(__name__)
+
+# A session record names a pid, and the OS recycles pids: once the claude
+# that wrote the record is gone, the number can belong to anything. The
+# record also carries claude's reading of its own start (``procStart``): on
+# Linux the ``/proc/<pid>/stat`` start time in clock ticks since boot, fixed
+# for the process's lifetime, and elsewhere ``ps -o lstart``, a wall-clock
+# time. Only the latter needs slack, for the small clock steps that move a
+# ``ps`` start time on some platforms.
+PID_REUSE_SLACK_S = 120
+
+_MONTHS = ("Jan", "Feb", "Mar", "Apr", "May", "Jun",
+           "Jul", "Aug", "Sep", "Oct", "Nov", "Dec")
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


realiti4:
Hi, thanks a lot for the PR! It is a really nice approach to fix a valid issue.

Based on my checks there might be a problem with reconstructing the process start time as `time.time() - elapsed` though. On Linux, `ps` derives `etime` from boot-relative process timing, while `time.time()` and the record’s `startedAt` are wall-clock values. If the wall clock is corrected independently—for example, after a WSL2 host sleeps and resumes—the inferred process start can move forward even though the process has not restarted. A live session could then be classified as a recycled PID, which would make this guard fail in the dangerous direction.

Would it be possible to use the existing `procStart` field instead? Every session record I have from Claude Code 2.1.170 onward includes it. On Linux, it represents the process start token from `/proc/<pid>/stat`, so comparing the current token with the recorded `procStart` would detect PID reuse without mixing clock domains. Records without `procStart`, or cases where it cannot be read, could retain the current fail-open behavior.

What do you think?

GrahamCampbell:
Thanks, good catch on the clock domains. I've reworked the PR around it.

One correction on procStart though: it isn't a /proc/<pid>/stat token. I checked the binary (2.1.258, and 2.1.170 where the field first appears) and on every non-Windows platform it's just the trimmed output of `LC_ALL=C TZ=UTC ps -o lstart= -p <pid>`. Windows gets a separate procStartFt instead.

That's a problem for a straight string compare, because on Linux procps computes lstart as btime + start_ticks / Hertz, and btime moves with every wall-clock step. So the WSL2 resume you describe would shift lstart for every live process and an exact match would flag all of them as recycled, after a step of even one second in either direction. macOS is fine, since there both lstart and etime come from p_starttime, which is a wall-clock snapshot taken at fork.

What the PR does now: read the live process with the same ps -o lstart= command, parse both to seconds, and pass anything that isn't newer than its record plus a small slack. If it still reads as newer, check ps -o comm=,args= and only drop the record when the process at that pid isn't a claude. That makes the WSL2 case fail open, since a live session is always a claude, and a pid genuinely recycled by some other claude just lingers for that process's lifetime, same as today. No procStart, Windows, or any ps failure keeps the current behaviour.

realiti4:
Thanks for the rework. The `lstart` + “is it a Claude process?” fallback handles the clock-step case nicely.

There is one thing, though. From my checks with Claude 2.1.263 on all three platforms, it looks like `procStart` is platform-specific rather than `lstart` everywhere.

* **macOS:** `"Mon Sep  7 06:21:00 2026"` — this matches `LC_ALL=C TZ=UTC ps -o lstart=` exactly, so your reading is correct there.
* **Linux:** `"11485"` — this is `/proc/<pid>/stat` field 22 (start time in clock ticks since boot). I see the same format in every Linux record I have going back to 2.1.170, from both the CLI install and the VS Code extension's bundled binary. The Linux binaries I checked (2.1.251/252/258/263) read `/proc` first and only fall back to `ps -o lstart=`. That branch is platform-gated and compiled out of the macOS builds, which is probably why you didn't see it.
* **Windows:** `"134332352612628209"` — a FILETIME stored in the same `procStart` key; there doesn't appear to be a separate `procStartFt` on disk.

Because `_lstart_seconds` raises on the Linux form and the caller treats that as unknowable, the recycled-PID check currently never fires on Linux. I confirmed this with a fake recycled PID: a Linux-format record is kept, while an `lstart`-format record is dropped.

Could you add a Linux branch that reads `/proc/<pid>/stat` field 22 and compares it to `procStart` by string equality? Since both values are the same boot-relative tick count, no clock conversion or slack should be needed. The existing `lstart` path can stay for macOS.

GrahamCampbell:
Thanks, you're right about Linux, and I've corrected my earlier note on Windows too: the writer puts the FILETIME under procStartFt, but the reader accepts either key, so a digits-only procStart can turn up there as well.

The check now dispatches on the stamp's shape rather than the platform. A digits-only procStart is compared by string equality against field 22 of /proc/<pid>/stat, counted from the last closing parenthesis so a command name with spaces or parentheses cannot shift it, with no slack and no claude-name fallback since a process keeps that value for life and no two processes at one pid share it. Anything else still goes through the lstart path as before. When /proc cannot be read, which includes a FILETIME on Windows and a hidden /proc, the record passes as it did.

Also rebased onto main.

realiti4:
Thanks a lot, it looks good merged it.

Issue #220, opened 2026-08-05
usage: /api/oauth/usage 429 (Retry-After: 0) keys on the account, not token×UA — busy accounts never refresh; proposal: cswap refresh + opt-in 1-token header fallback


cagrimertbakirci:
## Symptom

On a Linux machine with several long-running Claude Code sessions, `cswap list` serves days-old numbers for exactly the accounts being used, while idle accounts refresh fine. When I looked, the active account's row was 54 h old with `lastError: http-429`, and `claude-swap.log` shows weeks of `http-429, retry-after 0s` — always striking whichever account was active at the time.

## What I measured (2026-08-04, cswap 0.24.1, Linux)

`poll_policy` models the limiter as a per-access-token budget per UA class. What I observed doesn't fit that model:

1. For an account under active inference load, `GET /api/oauth/usage` returned `429` with `Retry-After: 0` for:
   - the access token cswap holds — although **no cswap poll had run for 22 hours**, so the budget wasn't spent by polling;
   - a **different, freshly-issued access token of the same account** that had never been used against the usage endpoint at all;
   - both `claude-swap/1.0` and the first-party `claude-cli/...` User-Agent strings.
2. At the same moment, from the same IP, an idle account answered `200` instantly.
3. The block isn't permanent — the same account intermittently answers `200` again later; over weeks of logs the 429s correlate tightly with which account is busy.

So the limiter keys on the **account**, not on (token × UA), and rejection correlates with the account's own activity rather than with polling volume. I don't know the exact trigger (load-, volume- or flag-based, and Anthropic can retune it any day); the robust part is: busy account → `429 Retry-After: 0` for every token and UA, idle account → `200`.

Consequences for cswap: failure backoff can never win for a busy account — the accounts users most want to watch are the ones that never refresh — and the `last429At`/AIMD cadence logic is reacting to account load, not to cswap's own request rate. (Related: #208 is about how far a 429-frozen value should be trusted; this is about why it freezes in the first place.)

## A fallback that works: unified rate-limit response headers

A minimal `/v1/messages` request (`claude-haiku-4-5`, `max_tokens: 1`, one-word prompt — ~10 tokens total) returns current utilization in its response headers even while the usage endpoint refuses the account:

```
anthropic-ratelimit-unified-status: allowed
anthropic-ratelimit-unified-5h-utilization: 0.34     # fraction; ×100 = five_hour utilization
anthropic-ratelimit-unified-5h-reset: 1785904800     # epoch seconds
[...]

PR #308, opened 2026-09-02
feat: add guarded five-hour window warming

### LINKED ISSUE #263: Option to auto-start the 5 hour session when it refreshes

AminDhouib:
## Tracking

- Closes #263.
- Supersedes #273: this implements the same five-hour timer-starting goal while adding opt-in `cswap auto` and TUI integration, unknown-usage handling, identity isolation, duplicate-spend guards, and dry-run support.
- Related to #220, which documents that the usage endpoint can be unavailable under account load. This PR does not replace its proposed live-metrics fallback; it prevents unavailable usage from leaving a cold five-hour window unwarmed indefinitely.

## Summary

- Add an opt-in five-hour window warmer, available either as standalone `cswap warmup` or inside the existing `cswap auto` process through `autoswitch.warmupFiveHour`.
- Add a persisted TUI toggle on the Auto screen (`w`), with a spend warning that reflects whether Auto is dry-run or live.
- Warm only cold or unconfirmed OAuth accounts. A reliable live five-hour reset/usage signal still skips; known active weekly exhaustion, disabled/API-key/non-switchable accounts, and identity mismatches also skip.
- If usage remains missing, stale, errored, or hollow after a probe, make one guarded Haiku request instead of leaving the window cold indefinitely.
- Keep every prompt bound to the expected OAuth identity in an isolated account profile, without switching the global Claude login.
- In the Auto event feed, render already-live and other informational warm-up results in muted grey; render actual and dry-run warm actions with the existing orange accent, with successful sends labeled `(5h warm-up sent)`.

## Why both integration modes

The integrated mode lets the normal `cswap auto` process perform both jobs, including a ten-minute warm-up deadline even when the configured auto poll interval is longer. The standalone command remains useful for users who want warming without automatic switching.

Both modes use the same lock and persistent state. Users should run one mode, not both.

## Unknown-usage policy

The original revision failed closed when fresh successful usage was unavailable. In practice that can coincide with a fleet-wide reset and leave every five-hour window cold. The opt-in now treats unavailable usage as unconfirmed rather than permanently ineligible:

- A fresh positive percentage or a future reset proves the five-hour window is live.
- Stale data must have a future absolute reset; a stale percentage alone cannot suppress warming forever, including `trust_extended` rows.
- A known future account-wide or selected-model weekly reset at 100% still prevents a pointless request.
- Missing/errored/hollow data permits one request, protected by identity-bound persistent state for the full five-hour window.

## Safety details

- Successful, timed-out, nonzero-exit, crashed, and otherwise possibly accepted requests receive a full-window duplicate-spend guard. The pre-launch record conservatively covers the maximum timeout, then successful or ambiguous completions are timestamped at completion.
- State is keyed by the verified `(organization UUID, normalized email)` identity, migrates the previous slot-keyed schema, and continues to protect accounts after slot swaps.
- The account slot, stored identity, isolated profile, and launch identity are revalidated while holding cswap's account lock.
- The exact launch environment is preflighted with `claude --safe-mode auth status --json`; only the expected first-party Claude OAuth identity may proceed.
- API/provider override environment variables are scrubbed, and managed cloud-provider routing fails closed at the auth preflight.
- The warm-up uses fixed `claude-haiku-4-5`, low effort, safe mode, disabled tools, and no session persistence.
- Stopping or disabling the warmer prevents later accounts in an active sweep from launching, and rapid TUI toggles cannot overwrite a replacement scheduler deadline.
- `--dry-run` remains spend-free in standalone and integrated modes.

## Validation

- `uv run pytest -q -p no:cacheprovider` — 2,129 passed, 75 skipped
- Focused warmer/autoswitch/TUI tests after the event-style update — 411 passed
- Ruff on all changed source files and clean changed test lines — passed
- `python -m compileall -q src` — passed
- `git diff --check` — passed
- Graphify extraction, path, impact, and post-change queries confirm warm-up events reuse `event_text` and shared `Palette` roles
- Five independent correctness, compatibility, security, and test-quality review passes — no remaining blockers

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,10 @@ cswap auto --model Fable       # also switch when the Fable weekly limit is hit
 cswap auto --once              # single check-and-switch, for cron/scripts
 cswap auto --dry-run           # log what it would do, never switch
 cswap auto --strategy consume-first   # burn the soonest-resetting account first
+
+# One process can also keep five-hour windows started:
+cswap config set autoswitch.warmupFiveHour true
+cswap auto
 ```
 
 <details>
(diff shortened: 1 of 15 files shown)


AminDhouib:
Update: Been running this for roughly ~24 hours. I can confirm it is working just fine.

image

PR #316, opened 2026-09-03
Read a symlinked profile's keychain entry under its target path

GrahamCampbell:
Claude derives its keychain service name from the exported CLAUDE_CONFIG_DIR string, so a session profile that is a symlink can hold its credential under two names: the link's own path, when claude was launched through the link, or the target's path, when a tool launches claude at the target directly and keeps the per-account symlink for cswap's bookkeeping. Every read of a session profile hashed only the link's path, so on macOS a profile of the second kind always missed the keychain entry and fell back to the plaintext seed at the target, which claude stops updating the moment it first writes the keychain. The usage collector then fetched with that stale seed until it expired and, with the session still live, deferred every pass as a token-expired sentinel, so the account's usage silently froze at the last pre-rotation measurement while the credential in the keychain was perfectly fresh. The same lookup fed the reuse check and the credential adoption on exit, which saw the seed rather than the head.

This makes the read try the link's own service name first and, when the profile is a symlink, the target's next, before falling back to the plaintext file, and applies the same order to the existence test the validation-timeout fallback uses. Deletion keeps hashing only the profile's own path, since removing an entry under a target that another tool manages is not cswap's to do. Absence under one name still means looking further, and an unreadable keychain keeps its existing strict and best-effort behaviour.

### DIFF
--- a/src/claude_swap/session.py
+++ b/src/claude_swap/session.py
@@ -305,13 +305,14 @@ def _may_have_credential_material(session_dir: Path) -> bool:
         pass  # no readable seed — the keychain may still hold the material
     if Platform.detect() != Platform.MACOS:
         return False
-    try:
-        material = macos_keychain.get_password(
-            keychain_service_name(session_dir), _keychain_account_name()
-        )
-    except macos_keychain.KEYCHAIN_ERRORS:
-        return True  # unreadable is not absent: preserve the profile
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for addressing this and the PR! I've merged it.

PR #317, opened 2026-09-04
auto: a voluntary move landed four points under the wall and bounced back

codeslake:
Stacked on #199 (`feat/tui-autoswitch-live`); this PR's own diff is that branch..HEAD.

Observed on a live fleet: consume-first moved the active account (66% used)
onto a peer at 86% used because that peer's weekly window reset sooner. The
peer walled to 91% four minutes later and the engine bounced straight back —
a round trip of two switches inside about seven minutes, gaining four points
of quota for the cost of two swaps. Consume-first's landing gate admits a
voluntary target whenever it sits under the operator's `threshold` (90% by
default), with no margin; `hysteresis_pct` already guards the `best` arm's
landing the same way, and the voluntary consume-first move had no
equivalent.

`optim` adds that margin without changing anything else: the same key, the
same triggers, as consume-first. A VOLUNTARY move (the active below
threshold, drifting to a strictly-sooner-resetting peer) only admits a
target whose binding window sits under `threshold - hysteresis_pct`, not
just under `threshold`. A hold because every sooner target lacks room keeps
the existing `already-consuming-soonest` reason, so nothing changes for a
watcher keyed on that text.

What `optim` does not change:
- The departure threshold itself — the operator's margin stays the operator's.
- 5h stays a landing/servability gate, never a ranking key.
- The 7-day reset stays the key's reset element; a scoped (per-model)
  window's own reset is now shown, not keyed.
- Forced moves (proactive, at-limit, failover, disabled-active) keep the
  plain `threshold` landing gate — a near-wall target still beats staying on
  one.

Panel: "Next best" drew a chip for the account-wide 5h/7d windows only, while
the ranking key it mirrors already folds in each per-model scoped window
named by `autoswitch.model`. A candidate bound by a scoped weekly limit
therefore ranked on a number the panel never displayed. It now shows one
chip per scoped window in `models`, and under `optim` sorts by the same
landing-margin key the engine's voluntary trigger uses, so the panel can
never show a target the live engine would refuse.

TUI: the auto screen gets an `s` binding to cycle strategy (best ->
consume-first -> optim), persisted the same session-only way `t` persists an
adjusted threshold — an in-memory settings swap plus `engine.apply_strategy`
(mirrors the existing `apply_threshold`), so a LIVE engine keeps its lock
and needs no restart. Nothing is written to settings.json.

https://claude.ai/code/session_01QFPGPPSYmx8NNGqbEsSpNc


codeslake:
Withdrawn by the author: the strategy discussion moved on. The landing-margin rule and the per-model window chip may return in a different shape.

Issue #255, opened 2026-08-17
feat(autoswitch): per-account threshold override — reserve headroom on shared accounts


aphexcx:
## Use case

I run `cswap auto` across a pool of accounts, and one of them is **shared with human users outside the rotation**. I want automation to stop drawing from that account before it's drained — e.g. "never take this account's Fable window above 85%" — so the remaining headroom stays reserved for the humans. The other accounts in the pool should keep the global threshold (they're fully dedicated, so stopping them early just wastes quota).

Today `autoswitch.threshold` is global, so there's no way to express a per-account ceiling. Setting the global threshold to the strictest account's cap taxes every other account in the pool.

## Proposed shape

A per-account threshold override that both sides of the engine respect:

- **switch-away**: the active account switches out when *its own* cap is reached (falling back to the global threshold when unset);
- **target selection**: an account at/over its own cap is not a landing candidate (same rule as today's global check, per-account).

Config surface could be whatever fits the codebase best, e.g.:

```
cswap config set accounts.3.threshold 85
# or
cswap threshold <num|email> 85 [--model Fable]
```

A per-model variant would be ideal since per-model weekly windows are already first-class (`--model`, and `usage.scoped[]` in `list --json`), but even an account-wide override covers the core need.

## Current workaround

An external cron/launchd script polls `cswap list --json`, reads `usage.scoped[].pct` for the shared account, and runs `cswap disable <acct>` at the cap (plus a one-shot `cswap auto --once --threshold <cap>` if that account is active), then `cswap enable` after the window resets. It works, but it lives outside cswap's own decision engine — happy to retire it the day this ships.

Thanks for the tool — the JSON surfaces made even the workaround clean.

PR #318, opened 2026-09-04
auto: per-account thresholds and standby accounts

matbanik:
Adds two per-account auto-switch policy fields — a per-account **threshold** and a
**standby** ("last man standing") flag — as additive keys on the `sequence.json` account
record. Both are opt-in: an account with neither key behaves exactly as it does today.

Refs #255.

## Why

The auto-switch engine currently applies one global policy to every account: a single
`autoswitch.threshold`, one ranking strategy, and no way to say that one account should be
spent differently from another. Two common asks fall out of that:

- **Per-account threshold** — a heavily-used primary and a rarely-touched spare want
  different switch-away lines.
- **A reserve account** — one account that is *only* used once no other account can be
  switched onto.

Issue #255 sketches a per-account priority field. This PR delivers the two stages of that
idea that are orthogonal to the ranking sort (a **gate** and a **filter**); a chaining
`order` field, which lands on the sort key itself, is deliberately left to a follow-up so
it can be reviewed against #260 rather than tangled with this change.

**#343 (`autoswitch.fallbackAccount`)** addresses the same "last man standing" need from a
different shape: one globally-named account forced onto only when the engine would otherwise
block. This PR puts `standby` on the account record instead, allows more than one, and adds a
`failback` trigger that hands back to a primary once one recovers. Flagging the overlap so
the two can be weighed together.

**Naming.** The flag was called `backup` in the first revision. It is now `standby`,
because "backup" already means the stored credential copy throughout claude-swap
(`~/.claude-swap-backup`, `backup.cswap`, the *Backup and migration* section).

## What changed

| Field | Type | Meaning |
|---|---|---|
| `threshold` | `float`, 50–99.9 | Per-account switch-away line. Overrides `autoswitch.threshold` for this account only. Absent ⇒ the global default. |
| `standby` | `bool` | Reserve account. Excluded from normal candidate selection; promoted only when no other account can be switched onto. |

Both live on the **account record** in `sequence.json`, not on a slot-keyed settings map,
because `swap` / `move` relocate whole records — a slot-keyed policy would silently transfer
to a different account.

### Engine

- The departure gate and the landing gate are now both **per-candidate**: an account is
  compared against its own threshold, not the global one, on both sides of a switch. Like the
  global threshold, it is a switch-away line, not a hard cap: when no candidate is under its
  own line, or the active account hits its limit, the engine still lands on the best candidate
  rather than stall.
- Standby accounts are filtered out of the candidate pool and re-added only as a last resort,
  after which the existing two-phase commit re-ranks against fresh usage — so a primary that
  turned out to have recovered is preferred over spending the reserve.
- `failback` returns from a standby to a primary as soon as one is usable again.
- The engine stays UI-agnostic: no new printing, only new typed events / reasons.

### CLI

```
cswap threshold                 # list the global default and every override
cswap threshold 2 85            # set account 2's line to 85%
cswap threshold 2 --unset       # return account 2 to the global default
cswap standby 3                 # mark account 3 as the reserve
cswap unstandby 3               # clear it
```

Both verbs accept a number, an email, or an alias, matching the other account verbs. The
long-flag spellings `--standby-account` / `--unstandby-account` work too, mirroring
`--disable-account`.

Values are validated **at the write boundary**: an out-of-range or non-numeric threshold is
rejected before `sequence.json` is opened, so a rejected write leaves the file byte-identical.

### TUI / JSON

The account list shows a `(standby)` badge and a `th NN%` badge when an account carries a
policy, alongside the existing `(disabled)` badge. `--json` rows gain `threshold` / `standby`
keys — but only when set, following the same convention `disabled` already uses, so a fleet
that never touches the feature emits rows identical to before.

## Compatibility

- Purely additive. Existing `sequence.json` files parse unchanged and behave identically.
- No change to the default policy: with no per-account fields set, ranking, cooldown,
  hysteresis, and every existing trigger string behave as before.
- Unknown/absent keys are tolerated on read; nothing is written unless a verb sets it.

## Tests

`+286` tests across six new modules (`tests/test_account_policy_values.py`,
`test_account_policy_store.py`, `test_autoswitch_account_threshold.py`,
`test_autoswitch_standby_accounts.py`, `test_account_policy_cli.py`,
`test_tui_policy_display.py`), all synthetic fixtures — no real credential store is read or
written.

```
2471 passed, 81 skipped, 3 warnings
```

Skip and warning counts are unchanged from the pre-change baseline (2185 passed, 81 skipped,
3 warnings) on the upstream `main` this branch merges (`7187ce8`), i.e. no existing test was
skipped, xfailed, or otherwise neutralised.

The branch also carries a `README.md` commit documenting exactly these two fields and nothing
else, so it does not describe features that live in the follow-up PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,35 @@ cswap auto --strategy consume-first   # burn the soonest-resetting account first
 - It fails safe: if a usage check errors it keeps trusting the last-known numbers while retries back off, and an expired token on an idle machine makes it hold rather than fail over (Claude Code refreshes the token on your next message).
 - An account whose refresh token has died is quarantined and reported until you either log in with it and re-run `cswap add --slot N`, or replace its stored credentials from a known-good export — a plain `cswap import backup.cswap` replaces dead-token slots on its own (`--force` is still required to replace other existing accounts; note a stale export can carry an already-superseded token). API-key accounts are never rotated onto unless you pass `--include-api-key-accounts`.
 - To hold an account out of rotation yourself — a work account you don't want touched, one you're resting — run `cswap disable <num|email>`; `cswap enable <num|email>` puts it back. Disabled accounts are skipped by auto-switch, bare `cswap switch`, and the `best` / `next-available` strategies, but stay fully managed and remain a valid explicit `cswap switch <num|email>` target. They show a `(disabled)` marker in `cswap list`, in the [TUI](#interactive-dashboard-tui), and in the [menu bar](#menu-bar-macos) — both of which also let you toggle the state in place (TUI: menu → *Disable / enable account…*; menu bar: *Disable / enable account*).
+- **Per-account policy.** Two knobs let one account be spent differently from the rest. `cswap threshold <num|email> <pct>` gives that account its own switch-away line, used in place of the global one on both sides of a switch — the engine leaves an account when *its own* line is reached, and refuses to land on a candidate already past *its own*. `cswap backup <num|email>` holds an account back as a reserve: it is skipped while any other account can still be used, and taken only when none can — after which auto-switch returns to a primary as soon as one recovers. Both are per-account overrides, not global settings, so an account you never touch keeps behaving exactly as it does today. `cswap threshold` with no arguments lists the global default and every override; the [TUI](#interactive-dashboard-tui) shows `th 85%` and `(backup)` badges beside the account. Clear either with `cswap threshold <num|email> --unset` or `cswap unbackup <num|email>`.
+
+  A three-account fleet where account 1 is spent early, account 2 runs a little longer, and
+  account 3 is the reserve:
+
+  ```console
+  $ cswap threshold 1 75
+  Set Account-1 (a@example.com) threshold to 75%.
+
+  $ cswap threshold 2 85
+  Set Account-2 (b@example.com) threshold to 85%.
+
+  $ cswap backup 3
+  Account-3 (c@example.com) marked as backup.
+    It will be skipped by auto-switch while any other account can still be used, and taken only when none can.
+
+  $ cswap threshold
+  Switch-away thresholds:
+    global default  90%
+    Account-1  75%
+    Account-2  85%
+  ```
+
+  `cswap auto` now leaves account 1 at 75% and account 2 at 85%, and only reaches for account 3
+  once neither of the others can be landed on. Only the keys you set are written — account 2's
+  record gains `"threshold": 85.0` and account 3's gains `"backup": true`, and clearing them
+  removes the keys again, leaving `sequence.json` byte-identical to a store that never used the
+  feature. A rejected value (out of the 50–99.9 range, or not a number) is refused before the
+  file is opened, so a bad command changes nothing.
 - By default only the account-wide 5h/7d windows drive switching. If you work on one model and hit its **weekly per-model limit** first (e.g. Fable), add `--model Fable` (or `cswap config set autoswitch.model Fable`) to fold that model's window into the decision, so it switches off an account whose model quota is spent even while its 5h/7d windows still have room.
   - **Model names** are Anthropic's own per-model `display_name`s, matched case-insensitively. The exact strings for your accounts are the per-model rows in `cswap list` (e.g. a line reading `Fable: 100%`).
 
@@ -192,6 +221,11 @@ cswap add --alias dev           # Add account and give it a short alias
 cswap remove 2                  # Remove an account
 cswap disable 2                 # Hold an account out of auto-rotation (keeps its login)
 cswap enable 2                  # Return a disabled account to rotation
+cswap threshold 2 85            # Switch away from account 2 at 85%, not the global default
+cswap threshold 2 --unset       # Return account 2 to the global default
+cswap threshold                 # Show the global default and every per-account override
+cswap backup 3                  # Hold account 3 back as a last resort ("last man standing")
+cswap unbackup 3                # Return it to normal rotation
 cswap alias 2 dev               # Give an account a short alias (usable anywhere NUM|EMAIL is)
 cswap alias 2 --unset           # Remove an account's alias
 cswap alias                     # List all aliases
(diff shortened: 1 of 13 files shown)


ivan-andreyev:
Our workflow needs backup → primary failback on the next fresh usage poll after
a primary recovers, even during cooldown. Would an opt-in cooldown exception
for `failback` fit here? Existing defaults and freshness/identity checks would
stay unchanged.

matbanik:
> Our workflow needs backup → primary failback on the next fresh usage poll after a primary recovers, even during cooldown. Would an opt-in cooldown exception for `failback` fit here? Existing defaults and freshness/identity checks would stay unchanged.

Thanks, worth raising. For context on what this PR does today: failback is gated by cooldown, and cooldown is measured from the last switch (the move onto the reserve). So failback only waits when a primary recovers within cooldownSeconds of that switch; after that it fires on the next poll. With defaults (300s cooldown, 60s interval) the reserve carries at most ~5 extra minutes.

I'd rather not add the exception to this PR. It's a separate policy setting, and it has a real trade-off: at-limit already bypasses cooldown, so exempting failback too would leave the primary⇄reserve pair with no rate floor, with only hysteresis and the recovery check against a flapping usage read. That's a call for @realiti4, and this PR is big enough already.

Two questions so it's easy to pick up as a small follow-up once this lands or gets direction:

What cooldownSeconds do you run? The gap matters much more at 3600 than at 300.
Would lowering cooldownSeconds work for you in the meantime, or is keeping a long floor on proactive switches the point?

stephanmundtcode:
+1 for merging this. My setup is the one from #255: a team account shared with people outside the rotation, which should drop out at around 60%, next to my own account that can run to 90%. A single global threshold can't express that, and this can.

I checked out the branch and ran it against my own two-account setup. Day to day it does what I want: it leaves the team account at 60% and won't switch onto it while it's above its line. The suite passes locally.

One thing I'd flag, not as an objection but so nobody is caught out by it: the per-account value is a switch-away line, not a hard ceiling. With my own account at 92% and the team account at 65%, the engine still switches onto the team account, and the same happens once my account is at its limit. That's how the global threshold behaves today, so it's a sensible default. But the README says the engine "refuses to land on a candidate already past its own", which promises more than the code does, and that sentence is what a shared-account user will read. For a case like #255 an opt-in strict cap would be a good follow-up.

On the overlaps: #319 and #320 sit on top of this one. #328 and #329 add per-window lines, a separate axis, but they touch the same code and will need a rebase whichever lands first. #343's `fallbackAccount` covers similar ground to `backup` here; having it on the account record and allowing more than one looks like the better base to me.

ivan-andreyev:
Our local wrapper uses 95/99/80/80% account thresholds with one reserve. We run
300s cooldown / 60s polling and want to preserve cooldown for ordinary rotation,
while opting into failback on the first fresh eligible poll. Lowering cooldown
globally wouldn't fit that policy. Agreed: a separate follow-up, pending
@realiti4's direction.

matbanik:
Thanks both. I've pushed two changes to this PR (and carried them up into #319 and #320). Neither changes behaviour.

**1. `backup` is now `standby`.** In claude-swap, "backup" already means the stored copy of an account's credentials: `~/.claude-swap-backup`, `cswap export … backup.cswap`, and the *Backup and migration* section of the README. With a second, unrelated meaning on the same word, `cswap backup 3` read like "back up account 3". `standby` has no existing use in the codebase, and it describes the role: the account waits outside rotation and is used only when nothing else can be. So the commands are now `cswap standby <num|email>` / `cswap unstandby <num|email>`, the record key is `"standby": true`, and the badge is `(standby)`. `failback` keeps its name.

@stephanmundtcode, if you set `cswap backup N` while testing, re-run it as `cswap standby N`. The old key isn't read, because this field has never been in a release.

**2. The README now describes what the code does.** @stephanmundtcode, you're right, and thanks for testing it against a real setup. The per-account value is a switch-away line, not a hard ceiling. The engine leaves an account at its own line and prefers candidates still under theirs. But when no candidate is under its line (your 92% / 65% case), or the active account is at its limit, it still lands on the best one rather than stall, the same as the global threshold does today. The sentence that said it "refuses to land" is gone. The bullet now says this, and it also spells out the standby lifecycle:
- It is skipped while auto-switch can still move to any other account, and used only when it can't.
- It hands back to a primary as soon as one is under its own threshold again, once the cooldown since the last switch has passed.

I agree an opt-in strict cap for the shared-account case in #255 is a good follow-up. It changes what happens when every account is over its line (hold instead of move), so it's a separate PR.

**On #343.** I don't think the two compete. They answer different questions:
- **`standby`** (here) is a tier: a property on the account record, allowed on several accounts. Those accounts are kept out of rotation until nothing else can take the switch, and `failback` returns to a primary once one recovers.
- **`fallbackAccount`** (#343) is a last-resort landing: one named account that auto-switch lands on when it would otherwise block, whatever its own headroom.

A fleet could reasonably use both. @realiti4, if you'd rather have one mechanism, I'm happy to fold whichever parts of #343 you want into this, or to rework this on top of it if it lands first.

**@ivan-andreyev**, thanks for the numbers. 95/99/80/80 with one reserve is exactly what the per-account lines and `standby` are for. Keeping the cooldown for ordinary rotation while letting failback go on the first fresh eligible poll is a coherent policy. I'd model it as its own failback delay (the way keepalived separates `preempt_delay` from the rest of its timers) rather than as an exception to cooldown. As agreed, that's a small stacked follow-up once @realiti4 has had a look.

ivan-andreyev:
Are you planning to implement the failback-delay follow-up, or would you prefer a PR from us?

matbanik:
> Are you planning to implement the failback-delay follow-up, or would you prefer a PR from us?

@ivan-andreyev I took a stab at it — it's up as #378, so you can just review it rather than
write it. If it isn't what you want, no hard feelings: say so, or send yours and I'll close
mine.

It's one opt-in setting, `autoswitch.failbackDelaySeconds` (also `cswap auto
--failback-delay`). Unset, nothing changes — the handback keeps following `cooldownSeconds`.
For your fleet:

```bash
cswap config set autoswitch.failbackDelaySeconds 0
# cooldownSeconds stays at 300 and still governs ordinary rotation
```

That's the first fresh eligible poll, in the same tick — it doesn't cost you an extra 60 s
cycle.

**One design call worth your eye.** The delay is measured from the moment the handback
becomes *actionable* — the first tick where a primary is genuinely landable — not from the
last switch. A floor measured from the switch onto the reserve would be nearly inert for
every value but 0: the reserve gets promoted *because* the primaries are exhausted, so by the
time one rolls over, a floor measured from that moment has long since passed. Anchored on the
recovery, every value means "the primary has to hold for N seconds before we hand back",
which is what `preempt_delay` means in keepalived and what the preemption delay means in the
VRRP implementations generally. At your `0` the two are identical anyway; the difference only
shows up if someone later wants a real stability window.

The anchor is one additive key in `autoswitch_state.json`, persisted rather than in-memory
because `cswap auto --once` is a cron mode where an in-memory timer would never accumulate.

**And the trade-off, stated rather than buried:** at `0` there is no rate floor left on the
primary⇄reserve pair beyond the engine's refusal to undo its own last move. That's the point
of the setting, and the README now says so instead of implying the stability a preemption
delay usually carries. If you ever see the pair flapping on a primary that hovers around its
line, a small non-zero value is the knob rather than raising `cooldownSeconds` again.

Two things you can check that I can't: that it behaves on your real 95/99/80/80 fleet, and
that `0` is actually the semantics your wrapper needs — "first fresh eligible poll" is my
reading of your message, and if you meant something stricter (say, don't hand back until the
primary has been under its line for two consecutive polls) that's a different value, not a
different design.

@realiti4 — this is the follow-up we said would be stacked. It's additive and default-off;
happy to hold it until you've looked at #318, or to drop it entirely if you'd rather the
handback stayed on the one timer.

PR #322, opened 2026-09-04
Read setup-token usage from a one-turn Claude Code session

rjt-rockx:
## Problem

A `claude setup-token` credential carries only the `user:inference` scope. The usage endpoint (`/api/oauth/usage`) needs profile scope, so every setup-token account answers 403 there on every pass, renders a permanent `usage unavailable (http-403)`, and is invisible to `cswap auto` and the rankings. Worse, a handful of those 403s trip a host-wide 429 on the endpoint for an hour, which then blanks the full-scope accounts on the same host too. Observed on a host with three setup-token slots: the log cycles `http-403` x4 then `http-429, retry-after 3600s` all day.

## Approach

Read usage for setup-token accounts the way Claude Code itself sees it. Every inference response carries the plan windows, and Claude Code surfaces them in stream-json as a `rate_limit_event`. So for these accounts cswap runs one isolated, non-interactive turn under the account's `cswap run` session profile:

```
CLAUDE_CONFIG_DIR=<profile> claude -p ok --output-format stream-json --verbose --model claude-haiku-4-5-20251001 --max-turns 1 --tools ""
```

and parses the last `rate_limit_event` line into the raw usage-API shape, so `build_usage_result` and everything downstream (list, status, TUI, menubar, autoswitch) are unchanged.

All network activity is the `claude` binary's own ordinary client request. cswap never builds an API request with the token itself, deliberately: that keeps consumer accounts inside normal Claude Code traffic.

## Cost

- An exhausted account still emits the event (`status: rejected`, utilization 1.0) and the process exits non-zero; stdout is parsed regardless of exit code. That read costs nothing.
- A healthy account spends one tiny Haiku turn (empty prompt, no tools) per poll, on the existing `usage_store`/`poll_policy` cadence. No new flags or settings.
- The probe runs with cwd `<profile>/usage-probe/` (0700), so its transcript never lands in a real project's history.

## Details

- Detection is narrow: scopes exactly `["user:inference"]` and no refresh token, the shape `add-token` writes. Full-scope logins never touch this path.
- `try_fetch_usage_for_account` gains an optional `probe_profile_for(account_num, email)` callable (same pattern as `refresh_via`). The switcher passes the session profile for inactive slots and Claude Code's own config dir for the active login (`CLAUDE_CONFIG_DIR` left as inherited there, since on macOS the keychain service name hashes the raw env value).
- A slot without a bootstrapped session profile reports `no-session-profile` with an `ERROR_NOTES` remedy (`cswap run <num>` once) rather than bootstrapping inside a collect pass. Other kinds: `no-claude-binary`, `probe-timeout` (90 s, child killed), `no-rate-limit-event`. None is a permanent auth verdict, so none strikes or quarantines.
- Tests: 28 new in `tests/test_setup_token_usage.py`; an autouse fixture in `conftest.py` stubs the spawn seam so no test can spend a real turn. Full suite: 2204 passed.

## Verified live

Ubuntu host, Claude Code 2.1.257, two setup-token slots (one at 100% of its 5-hour window). `cswap status` and `cswap list` show both accounts' 5h/7d windows; the log carries one `Usage via Claude Code turn for account N` line per slot; `ls` confirms nothing but `usage-probe/` was created in the profiles.

One migration note: a store that already holds an `http-429` backoff from the endpoint path keeps that backoff until it expires (up to an hour) before the first probe runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,6 +354,21 @@ cswap add-token --email user@example.com     # optional label override
 `--email` is optional; omitted values use `setup-token-{slot}@token.local`
 (or `api-key-{slot}@token.local` for API keys). No Anthropic API calls are made.
 
+**Usage for setup-token accounts.** A setup-token carries only the `user:inference`
+scope, so the usage endpoint every other account is read from answers 403 for it,
+and a few such probes get the endpoint rate-limited for an hour for every account
+on the host. cswap therefore never queries the endpoint for these accounts. It reads
+their usage the way Claude Code itself sees it: one isolated, non-interactive Claude
+Code turn (`claude -p`, Haiku, one turn, no tools) under the account's `cswap run`
+session profile, parsing the rate-limit event Claude Code emits from the response
+headers. An exhausted account still emits that event and the read costs nothing; a
+healthy account spends one tiny Haiku turn per poll. The probe runs in a private
+`usage-probe/` directory inside the profile, so nothing lands in a real project's
+history. `claude` must be on PATH, and a slot's session profile must exist (run
+`cswap run <num>` once); until then the account shows `usage unavailable
+(no-session-profile)`. The active/default login is probed in Claude Code's own
+config directory instead.
+
 **API-key accounts.** An `sk-ant-api...` value registers a managed API-key account
 (the kind Claude Code uses after `/login` with a key) rather than an OAuth
 setup-token. It switches like any other account; since API keys have no subscription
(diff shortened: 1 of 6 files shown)


non7top:
Tried this branch against a real setup-token account (`add-token` → `watch`, no `cswap run` in between, since that's not part of my normal flow for these accounts) and initially still saw the old `usage unavailable (http-403)` on the detail view, not the `no-session-profile` remedy text. Running `cswap run <num>` once fixed it and usage now shows correctly. Since `add-token` accounts don't otherwise need a session profile for anything today, this extra step is easy to miss — might be worth calling out more prominently (e.g. in `add-token`'s own output, or bootstrapping the profile automatically on first fetch) rather than only in the PR description.

PR #325, opened 2026-09-05
An account whose weekly reset was never reported can never be chosen, so it stays unknown forever

codeslake:
## The failure

An account whose weekly window has never reported a `resets_at` can never be chosen by
`consume-first` (or by `dynamic`, which shares its ranking), so it is never used, so its
window is never established, so its reset stays unknown. The exclusion is permanent and
self-sustaining.

It is not a ranking problem. `consume_first_rank_key` maps a missing reset to `+inf`, which
would merely sort the account last — but the candidate never reaches the key. In
`_rank_candidates_pass` the strictly-sooner gate drops it one step earlier:

```python
if trigger in CONSUME_FIRST_STRATEGIES and (
    reset_ts is None
    or active_reset_ts is None
    or reset_ts >= active_reset_ts
):
    continue
```

`reset_ts is None -> continue` removes the account from `qualifying` outright. On a fleet where
several accounts sit at 0% with no reported reset, the strategy holds
`already-consuming-soonest` forever while the accounts it refuses to consider stay unmeasured.
The panel makes this invisible rather than visible: `data.chip_label` returned a bare `5h:`
for a window with no reset, so the column the reader would use to notice simply went blank.

## The fix

- `_rank_candidates_pass` no longer drops an unknown-reset candidate. One otherwise-eligible
  candidate — servable, not disabled, below threshold, exactly the same tests as before — is
  admitted as a **probe** and ranked ahead of every known reset (`consume_first_rank_key` takes
  a `probe` flag and puts `-inf` in the reset field). Using it once is what establishes its
  window; from the next successful poll the ordinary ordering applies with no special case left.
- Bounded against ping-pong in three ways: at most **one** probe candidate is admitted per
  decision; a probed account records a cooldown (`PROBE_COOLDOWN_S`, one hour — longer than
  several poll ticks, so a durably unknown reset is not re-probed every cycle, and short enough
  that a small fleet is not locked out of its only unmeasured peer); and landing on an account
  for any reason clears its cooldown, because being used is the thing the probe was for.
- `_seven_day_reset_unmeasured` separates a reset that was **never reported** from one that has
  merely **elapsed**. Only the first is a probe candidate.
- The cooldown lives in the engine's existing `autoswitch_state.json`, beside `lastSwitchFrom`
  and `leftHeadroom`. No new file, no new scheduler, no new thread.
- The switch is named in the decision log: the trigger is `probe` and the line says the weekly
  reset was unknown, so a reader can tell a probe from an ordinary consume-first move.
- `data.chip_label` renders `5h(⟳?):` instead of a blank, in the same token shape as a known
  countdown so the panel's columns still line up; `widgets._reset_parts` says `reset unknown`
  rather than nothing. All four chip call sites — the dashboard's two, the auto view's rows and
  its chip-width table — read that one function, so the marker appears everywhere at once.

## What is deliberately unchanged

`best` never enters this gate, and known-reset ordering under `consume-first` and `dynamic` is
unchanged. Both are pinned by tests, and the ordering test carries a mutant control that moves
its result — an unchanged assertion with nothing that can move it proves nothing.

## Corners this deliberately does not cover

Named here so the next reader does not mistake them for oversights.

- **The active account's own reset.** `if active_reset_ts is None: continue` runs before the
  probe branch, so nothing is probed while the ACTIVE account's weekly reset is unreported.
  That is not the same self-sustaining loop: the active account is by definition already being
  used, so switching away cannot teach its reset. The existing `reset-unknown` decision reason
  already names this state.
- **`dynamic` with every account over the threshold.** A probe is admitted only under
  `not by_recovery_axis`; in the recovery-axis regime the ranking key has a different shape and
  a probe cannot be compared against it. Everyone is blocked there, so the value of probing is
  low.
- **The panel's candidate set is looser than the engine's.** It applies neither the no-return
  bar, nor the quarantine exclusion, nor the usage-freshness requirement. Pre-existing
  coarseness which the probe's rank makes more visible, since a wrong row now sorts to the top
  rather than into the middle. Mirroring all three changes what the panel considers a
  candidate, which is a different subject from this one.
- **The spend row.** `_reset_parts` is shared with the pay-as-you-go budget row, so a spend
  window with no reported reset now also reads `reset unknown`. It is honest, but no probe can
  ever close it: `relevant_windows` excludes spend from the ranking.
- **`cswap list` still prints a bare `7d: 10%`.** `_format_usage_lines` is untouched, so the
  CLI list and the TUI card read differently for the same account until it is brought along.


codeslake:
Narrowing one claim from the commit message: the JSON contract change (`usageStatus` going non-`ok` for a rolled window) is narrower in production than the fixture rewrite in an earlier pass suggested. It needs BOTH the 5-hour and 7-day windows to have rolled (or be pct-less) at once -- if either window is still live, `--list`/`--status` keep reporting `ok` from that surviving window's real headroom, per `_drop_rolled_windows`. A single rolled window with the other window still live is the common case and stays `ok`.

PR #326, opened 2026-09-06
fix: pin Keychain trusted-app ACL via delete-and-recreate, not in-place widen

### LINKED ISSUE #279: security add-generic-password -U resets Keychain ACL to writer-only on every credential rotation, locking Claude Code out

oficiallyAkshay:
Thank you for open-sourcing this and keeping up with it — that's real, unpaid extra effort, and it shows. It saved me a genuine chunk of time I'd have otherwise spent building this myself. Appreciated.

---

Alternative to #280, fixing #279.

## Symptom

Right after `cswap` rotates accounts, Claude Code reports "OAuth session expired and could not be refreshed" — against a credential that was just written and is valid. Reproducible every time on a headless machine (no GUI session, e.g. under `launchd`).

## Root cause

`set_password()` calls `security add-generic-password -U` with no `-T`, resetting the item's ACL to trust only `/usr/bin/security`. Claude Code reads the same item via Security.framework — a different, now-untrusted caller — so the next read fails. Same diagnosis #280 reached.

## Divergence from #280

#280 widens the *existing* item's ACL in place (`-U` + `-T` on an update). Verified against a real keychain: macOS routes that through `SecKeychainItemSetAccess`, which needs interactive authorization and can raise a SecurityAgent prompt — even unlocked, even headless. Nothing answers it there, so it hangs rather than fails.

This PR deletes the item and recreates it with the trust list set at *creation* — which never triggers that prompt. Proven live against a real, unlocked keychain by `test_set_password_trusted_apps_on_existing_item_deletes_then_recreates`.

**The problem with #280 as written:** it replaces a write that silently locks out a legitimate reader with a write that can silently hang forever. For an interactive terminal, one click. For anything unattended, a stuck process with no failure signal at all.

## Scope

Two new keyword-only params. `trusted_apps` on `set_password` is the fix — default (`None`/`[]`) is byte-for-byte unchanged, tested. `keychain` on all four functions lets tests target a throwaway keychain file instead of the caller's real one — default unchanged.

This only touches the write-time ACL bug. #281 and #283/#284 are a related but separate read-time bug in the usage-sentinel retry logic — same "Keychain unreliable off the happy path" cluster, different code path, not addressed here.

## Why no GUI prompt by default

Installing `cswap` is already opting into seamless credential rotation — that's the tool's premise. A user who wants a confirmation step on every switch hasn't opted into what `cswap` does. A `--confirm` flag is a small, additive follow-up if that default's ever wrong — it doesn't need to gate this fix.

## Testing

- New real-keychain test (`TestKeychainAclPreservedOnWrite`): `-T` on a new item works directly; `-U -T` on an existing item doesn't safely widen it; delete-then-recreate does.
- New cases covering `keychain=`/`trusted_apps=`, including the no-op-by-default guarantee.
- Full suite: 2,195 passed, locally on real macOS.
- Project's real-Keychain CI job, run locally: 63 passed.

Fixes #279. Happy to adjust, or defer to #280 if that's already trending toward merge.

### DIFF
--- a/src/claude_swap/credentials.py
+++ b/src/claude_swap/credentials.py
@@ -336,7 +336,7 @@ class CredentialStore:
         self._residual_verdict: bool | None = None
         self._last_active_credentials_backend: str | None = None
 
-    def _kc_call(self, fn, *args):
+    def _kc_call(self, fn, *args, **kwargs):
         """Run a ``macos_keychain`` wrapper call, learning Keychain usability.
 
         A success (including ``get_password`` returning ``None`` for a missing
@@ -351,7 +351,7 @@ class CredentialStore:
         "absent" and "failed", so a timeout would be misread as a usable Keychain.
         """
         try:
-            result = fn(*args)
+            result = fn(*args, **kwargs)
         except macos_keychain.KEYCHAIN_ERRORS:
             # A Keychain op FAILED. Recorded separately from the capability
             # cache because that cache is a routing decision others overwrite —
@@ -837,6 +837,7 @@ class CredentialStore:
                     CLAUDE_CODE_MANAGED_KEYCHAIN_SERVICE,
                     macos_keychain.keychain_account_name(),
                     api_key,
+                    trusted_apps=macos_keychain.resolve_trusted_claude_apps(),
                 )
             except macos_keychain.KEYCHAIN_ERRORS as e:
                 # _kc_call flipped routing to file mode; fall back to config below.
@@ -978,6 +979,7 @@ class CredentialStore:
                     CLAUDE_CODE_KEYCHAIN_SERVICE,
                     macos_keychain.keychain_account_name(),
                     credentials,
+                    trusted_apps=macos_keychain.resolve_trusted_claude_apps(),
                 )
             except macos_keychain.KEYCHAIN_ERRORS as e:
                 # _kc_call flipped routing to file mode; fall through to the file.
--- a/src/claude_swap/macos_keychain.py
+++ b/src/claude_swap/macos_keychain.py
@@ -36,6 +36,7 @@ its functions are only meaningful on macOS.
 from __future__ import annotations
 
 import os
+import shutil
 import subprocess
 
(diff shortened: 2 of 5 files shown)


realiti4:
Thanks for investigating the authorization problem with #280. I checked Claude Code’s macOS 2.1.263 binary: it reads credentials through `security`, so the additional ACL grants aren’t needed. Deleting and recreating the item would also introduce an unnecessary failure window on each write. I’m closing this PR and keeping the existing `-U` behavior; the original symptom can be investigated further in #279.

PR #329, opened 2026-09-06
feat(autoswitch): a switch line per window, not one for both

### LINKED ISSUE #303: feat(autoswitch): per-window thresholds — 5h and 7d want opposite values under consume-first
### What problem would this solve?

Running `cswap auto` with `strategy: consume-first` across four accounts on staggered weekly resets. The 5h and 7d windows want opposite threshold values, and `autoswitch.threshold` is a single scalar compared against `max(pct)` of both, so there is no way to express that.

The asymmetry is in what an overshoot costs:

- **7d** — quota not spent before the weekly reset is gone permanently. This window wants an aggressive threshold (97-99) so the perishable quota consume-first exists to drain actually gets drained.
- **5h** — an overshoot costs one interrupted turn, and the window recycles in hours. Under consume-first the engine returns to that same account after the rollover anyway (it still resets soonest), so the last few points are recovered on the next cycle rather than lost. This window wants a moderate threshold (~90) to keep swaps clean.

With one knob I have to pick a compromise that is wrong for one of them:

- `threshold: 90` — clean 5h swaps, but every account is abandoned with ~10 points of weekly quota still on it. Across four accounts that is roughly 40 points of weekly quota expiring unused per cycle. This is arithmetic from the threshold, not a measurement.
- `threshold: 95` — better weekly drain, but the 5h overshoot budget halves to 5 points against a blind window of roughly 90s (`URGENT_INTERVAL_S = 60.0` plus the "macOS ~30s Keychain pickup tail" noted in `AutoSwitchSettings`' docstring). Heavy parallel/subagent turns burn through that.

Reading `account_headroom` (`oauth.py:543` in 0.25.0), the windows collapse to `100.0 - max(pcts)` before the threshold is ever applied, so which window bound the account is already discarded at the comparison. `SETTING_SPECS` has a single `autoswitch.threshold` entry, so this is not reachable through configuration today.

### What would you like to happen?

An optional per-window threshold map, with `autoswitch.threshold` staying as the default for any window not named:

```jsonc
{
  "autoswitch": {
    "threshold": 90,              // default for every window, unchanged
    "thresholds": {               // optional overrides
      "seven_day": 98
    }
  }
}
```

My case is `five_hour: 90, seven_day: 98`. It would want to compose with the per-account override in #255 rather than conflict with it.

#### Design note

The reason I am opening an issue rather than a PR: the engine collapses everything to one scalar, and `_binding_recovery_ts` states the invariant explicitly — it measures "the same set `account_headroom` measures, so ranking and headroom can never disagree about which window matters." Per-window thresholds can break that, and the two obvious implementations break it differently. I would rather have your read on the shape before writing it.

**Option A — normalize.** Rank on `max(pct / threshold_for_window)`; the threshold test becomes `>= 1.0`. One axis, invariant preserved. But it changes the units of every measured constant that is currently in raw percentage points — `hysteresis_pct`, `SPENT_HEADROOM_PCT = 3.0`, `HORIZON_HEADROOM_RATIO = 2.0` — so each would need re-deriving. That looks like a lot of re-measurement for a configuration knob.

**Option B — filter only.** Leave headroom and all ranking untouched at `100 - max(pct)`. Change only the trigger test and the candidate-legality gate from `max(pcts) >= threshold` to `any(pct_w >= threshold_w)`. Small diff, no constant re-measurement. Known weakness: legality and ordering can then disagree about which window binds — but legality is only a filter, and ranking still orders the survivors on the existing axis, so the disagreement is bounded rather than the scattered-gate problem `_recovery_is_useful` describes.

I would propose B, and I am happy to write it with tests if that is the direction you want. I have read the engine but have not measured tick behavior for either option, so treat the above as a reading of the code rather than evidence.

Version 0.25.0, macOS.

mrvnklm:
# feat(autoswitch): a switch line per window, not one for both

Closes #303. `+335 / −31`, two source files.

`autoswitch.threshold` is one number compared against `max(5h%, 7d%)`.
@lbagic's report states the asymmetry better than I can:

> **7d** — quota not spent before the weekly reset is gone permanently. This
> window wants an aggressive threshold (97-99) […] **5h** — an overshoot costs
> one interrupted turn, and the window recycles in hours.

Two optional keys, both defaulting to `null` = fall back to
`autoswitch.threshold`, so a config that never sets them behaves exactly as
before:

```bash
cswap config set autoswitch.thresholdWeekly 98
cswap config set autoswitch.thresholdFiveHour 90
```

## The part that is easy to get wrong

`max(pct)` throws away *which* window carried the maximum, and there is no
single line to compare it against that survives both directions. Both rows are
tests on this branch:

| lines (5h / weekly) | 5h | 7d | should fire? | vs the strictest line | vs the carrying window's line |
|---|---|---|---|---|---|
| 90 / 95 | 89.5 | 94 | no | 94 ≥ 90 → **false positive** | 94 < 95 → holds |
| 90 / 99 | 92 | 95 | yes, 5h by 2 | 95 ≥ 90 → fires | 95 < 99 → **false negative** |

So each window is divided by its own line and the closest decides:
`pressure = max(pct_w / line_w)`, trigger at `>= 1.0`. With both keys unset it
collapses to the old comparison exactly.

## Choices worth objecting to

- **`weekly`, not `sevenDay`.** Per-model scoped windows (`autoswitch.model`)
  reset on the weekly cadence and answer to the same line. Naming the key after
  the 7d window would leave a third, undeclared line for anyone using `--model`.
- **Two flat keys, not the nested `thresholds` map the issue sketches.**
  `SETTING_SPECS` is flat and `cswap config --json` reads it; a nested map needs
  a second validation path. `window_threshold(label, settings)` takes the
  settings object, so a per-account override (#255) composes by passing a
  different one.
- **No guard for a zero or negative override.** `_clamped` holds every float to
  its spec range and this floor is 50.0 — measured, a written `-3` and a written
  `0` both load as `50.0`. I had a branch for it until I checked.

## Out of scope, named rather than left to be found

`_no_return_account`'s `left_headroom > 100 - threshold` leg and
`_collect_scheduled_usage(threshold=…)` still read the base threshold. The first
only *disarms* a bar (on a 90/95 config it holds where the config would release
— costs a tick, never a wrong switch); the second drives poll cadence, so a
weekly line above the base polls slightly early. Both widen signatures with
several direct-call test sites, which felt like a separate change.

## Verification

`2197 passed, 3 skipped`. The `test_config_cli` counting assertions move 9 → 11,
and `test_json_list` pins that an unset override reports `null` rather than the
base threshold — the fallback is the engine's, and a JSON consumer has to be
able to see that the user never set one.

Replacing `window_threshold`'s body with `return settings.threshold` fails 7
tests; neutering the landing gate fails 11.


ardabayhan (commented):
Read this on the branch. The suite is green here too (2197 passed, 3 skipped), and I checked the compatibility claim rather than taking it: 20k random (5h, 7d) pairs with both overrides unset, zero divergence from `max(pct) >= threshold`. The `_landable` swap is faithful for the same reason you give — `account_headroom` is `100 - max(pct)` over the same `relevant_windows`, so `headroom is None` and `pressure is None` coincide exactly. The ratio design is right, and your table showing why `max(pct)` cannot work is the part I would have gotten wrong.

One finding, in the half you named as out of scope, but for the direction the note does not cover.

## A per-window line below `threshold - ESCALATION_MARGIN_PCT` decides outside the escalation band

`_collect_scheduled_usage` escalates on `100 - active_headroom >= threshold - ESCALATION_MARGIN_PCT`, against the base threshold, margin 15. Before this PR the ordinary trigger implied that condition: it fires at `pct >= threshold`, always 15 points inside the band. With a per-window line it no longer does.

Taking the shape your own motivation suggests — aggressive weekly line, relaxed 5h one — I ran it through `EngineHarness` on this branch:

```python
# base 98 -> escalation from 83;  5h line 80 -> trigger from 80
h = harness(threshold=98.0, threshold_five_hour=80.0)   # 3 accounts, active = 1
tick(usage={"1": _usage(81), "2": _usage(10), "3": _usage(20)})
```

```
fetch counts: {'1': 1, '2': 1}          <- account 3 never fetched
log:          Switched from account 1 to 2
```

Control, same 81%, no override:

```python
h = harness(threshold=90.0)              # 81 is inside 90-15
```

```
fetch counts: {'1': 1, '2': 1, '3': 1}  <- everyone refreshed
outcome:      NO_ACTION
```

So the switch is real and it is chosen from a snapshot up to `CANDIDATE_MAX_INTERVAL_S` old, with one candidate not consulted this tick at all. The codebase already treats this exact hazard as needing mitigation — the consume-first two-phase commit exists because "consume-first decides below the threshold, where the collector only escalates inside the ESCALATION_MARGIN_PCT band". This adds a second way to decide outside the band without one.

Your note covers a weekly line *above* the base ("polls slightly early") — true and harmless. It is the below-base direction that breaks the implication, and 50.0 is a legal value for either key against a default base of 90.

Passing `min` of the three lines to the collector would restore it without widening any signature: the escalation band only ever needs to start at the earliest line anything can fire on.

## Smaller

`_no_return_account`'s `left_headroom > 100 - settings.threshold` leg — reading it, your conservativeness claim looks right to me: on 90/98 it holds a bar the config would release, which costs a tick rather than causing a wrong switch. I did not exercise that path, so treat that as a reading rather than a measurement. But `_landable()` now exists and that leg is exactly its shape, so folding it in is a one-line change rather than the signature widening you were avoiding.

PR #330, opened 2026-09-07
Treat a live session's rejected credential as expired

GrahamCampbell:
While a session-mode claude is live, cswap reads its account's usage with a credential it must never refresh: the profile's own copy, or the stored backup when the profile has nothing readable. Two shapes of that read were still being recorded as transient failures. A backup copy that had already expired was sent anyway, and a credential the server refused with a 401 was recorded as an error, which starts the failure backoff and, on the fourth probe, draws a usage-endpoint 429 that parks the slot for over an hour. Neither record can lead anywhere, because the only party allowed to refresh that family is the live claude, which does so on its own next call.

This change returns the existing token-expired sentinel in both cases, the same answer already given for an expired profile credential under a live session. An expired backup copy is not requested at all, and a 401 on either read-only credential becomes the sentinel rather than an error. Sentinels are re-derived every pass and never persisted, so nothing accumulates and the slot resumes normally once the session renews or exits. Every other outcome keeps its identity, a 429 above all, since that budget belongs to the account rather than the credential. The idle path is untouched: with nobody live, a 401 still goes through the consume gate and the store's retry and strike accounting, and the active login's path keeps surfacing a 401 as an error because it has a recovery of its own.

The three-way split in the fetch is now written out, since after the exited-session adoption the profile-credential branch only ever runs under a live session.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,7 +132,7 @@ Sessions use your normal `~/.claude` setup (settings, CLAUDE.md, skills, MCP ser
 
 Running the account that is already your default login launches plain `claude` on that login instead of a session (a second copy of the active credential would go stale). Scripts that need the isolation guaranteed can pass `--require-session`, which refuses in that case instead.
   
-A session refreshes its own copy of the account's token, so once it exits, the credential it rotated is captured back into the account's stored backup before a switch or usage check uses that backup. While a session is still running, `cswap switch` refuses to move the default login onto its account if the stored backup has already fallen behind (activating it could only fail); exit the session first, or pick another account. While a session runs, its account's usage is read with the session's own credential and never refreshed by cswap; a read the server refuses shows as token expired until the session renews the credential on its next call.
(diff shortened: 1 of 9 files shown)


realiti4:
Hi, thanks a lot for addressing this. Skipping requests for an already-expired backup and preserving the handling of other errors both look good to me.

My concern is converting the 401 into a non-persisted sentinel. This removes the backoff, but the next collection can still request the same rejected token. I reproduced five requests across five consecutive collections, whereas the previous behavior made one request and backed off.

That seems likely to hit the 429 we're trying to avoid even faster. Could we keep showing "token expired" while suppressing requests for that rejected credential until it changes, or retain some bounded retry delay?

GrahamCampbell:
Agreed, and thanks for measuring it. A sentinel is never persisted, so the row stayed past the serve TTL and was handed out again on every collection, which under a one-minute watch is worse than the backoff it replaced.

The fix takes your first suggestion. A refused read-only fetch now stamps the hash of the access token it was refused with onto the row, riding on the sentinel as the one thing a sentinel record persists, and later passes skip the request while the credential still carries that token. A rotated token is worth one request; the next success clears the stamp. Other errors are unchanged, and the idle path never consults it. The new collect-pass test is your five-collections case: one request across consecutive passes, both reporting token expired.

realiti4:
Thanks for the adjustment and the PR again. Merged it!

PR #331, opened 2026-09-07
feat(json): report when a stored login expires (loginExpiresAt)

RizhongLin:
I keep a few accounts in cswap and run `cswap list --json` from a small script that pings me when something needs attention. The one thing it could not warn me about was the login itself: Claude Code stores `refreshTokenExpiresAt` next to each login's access token, the access token renews on its own, the refresh token does not, and the first sign was `relogin_required` after the fact.

This adds an additive `loginExpiresAt` (ISO-8601 UTC, same formatting as `usageFetchedAt`) to each `--list --json` row, read from that slot's own stored credential, so a script can say "log in again this week" a few days early. Absent when the login carries no such date (older logins), so schema 1 is unchanged.

- `oauth.login_expires_at_iso(credentials)` is the one place that knows the field; anything but a positive epoch is "unknown", never "now".
- `account_row(..., login_expires_at=...)` emits the field only when known.
- `_build_list_payload` already had the credential string in hand, so there is no extra keychain read.
- README schema notes updated; tests for the helper, the row and the list payload.

Left out on purpose: the human `cswap list` view and `status --json`. Happy to add a `(login expires in Nd)` marker to `list` in a follow-up if you would take it.

Full suite locally: 2245 passed, 4 skipped.


realiti4:
Thanks for the PR, it looks good merged it.

PR #332, opened 2026-09-07
fix(oauth): keep the persist warning off stdout

RizhongLin:
When a refreshed token cannot be saved back, `_persist` prints its warning to stdout. Inside `cswap list --json` that line lands in front of the JSON object and breaks whatever parses it, while `_build_list_payload`'s own comment states the contract: stdout stays a single machine-readable object. I hit this from a script that reads `cswap list --json` on a schedule.

Routing the warning to stderr keeps it visible in every mode and the payload parseable. The existing test now asserts the message on stderr and an empty stdout.

### DIFF
--- a/src/claude_swap/oauth.py
+++ b/src/claude_swap/oauth.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import hashlib
 import json
 import logging
+import sys
 import urllib.error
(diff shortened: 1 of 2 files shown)


realiti4:
Thanks for the PR! It looks good, merged it.

PR #335, opened 2026-09-07
Report the last fetch failure and retry time on unavailable rows

GrahamCampbell:
A row whose usage is null and whose status is unavailable tells a consumer nothing about why. The store knows: it holds the kind of the last fetch failure and, after a 429 or a run of failures, the time until which it will not try again. A dashboard or wrapper reading the JSON can only show a bare unavailable for what may be an hour-long park, which is what made a recent incident hard to read from the outside.

This adds two additive fields to such rows, on list rows and on the managed active row from status --json. usageError carries the last failure by kind, such as http-429 or timeout, and usageRetryAt carries the end of the live backoff as a UTC timestamp, omitted once the backoff has lapsed. Both are emitted only when the status is unavailable: a served measurement or a sentinel already explains the row, and a failure left over from an earlier pass would only contradict it. Rows without a recorded failure are unchanged, so consumers keying on the existing schema see nothing new.


realiti4:
Good idea, thanks for the PR! Merged it.

PR #343, opened 2026-09-10
feat(autoswitch): add a fallback account for the all-exhausted case

buck-openart:
Two opt-in additions to `cswap auto`, both about telling the engine which account you'd rather be on when headroom alone doesn't decide it. `fallbackAccount` covers "everything is exhausted, burn this one instead of idling"; `strategy=priority` covers "come back to this one as soon as it recovers."

---

# 1. `autoswitch.fallbackAccount`

When every OAuth account is exhausted, `cswap auto` sits `BLOCKED` until the earliest reset. That's the right default, but if you have a separately-paid seat you'd rather burn than go idle — an enterprise account, a spare — there's currently no way to say so. This adds an opt-in escape hatch for that case.

## What it does

`autoswitch.fallbackAccount` designates one account to force onto **only** when the engine would otherwise block:

- The active account must genuinely be unable to carry on — `at-limit` (out of quota) or `failover` (dead credential). Under a merely-`proactive` trigger the active still holds headroom the fallback doesn't, so nothing happens.
- Every OAuth candidate must be at or over the limit on whichever window binds first. An ordinary switch always wins; the fallback never preempts one.
- The target must be a real rotation candidate: switchable, not the current account, not quarantined.

It reuses the existing freshen/quarantine/switch loop as a single-candidate ordering with `trigger="fallback"` rather than adding a bespoke switch path, so cooldown bookkeeping, credential freshening and quarantine all behave as they do everywhere else. Once any account recovers headroom, normal rotation resumes.

If the fallback can't be landed on — unfreshenable, quarantined, or held by a live `cswap run` session — the tick falls back to the normal exhausted block, keeping the `AllExhaustedEvent`, `earliestResetAt` and the `BLOCKED` `--once` exit code a caller gets today, along with its reset-aware sleep. That does mean the hatch can land up to `MAX_SLEEP_S` late when the cause was self-clearing. I tried the obvious refinement — retry those at the normal interval — and backed it out: `skip-live-session` never leaves `candidates`, so it never terminates, and a user running `cswap run` on their designated seat would wake the loop at 10x for the life of that session. Ten minutes late is a rounding error against the hour of idling this exists to avoid.

An API-key account can't serve as the fallback, and is refused in either setting of `includeApiKeyAccounts`. With the opt-in off, switching onto a metered slot here is a one-way door — the next tick's `active-api-key` early return holds, so rotation never resumes even after the OAuth accounts reset. With it on, `api_key_candidates` are already reachable through the ordinary ranking, in rotation order — so naming one here buys no control over which metered key gets billed, and pretending otherwise would imply a control that doesn't exist.

An unresolvable or ineligible identifier emits a one-shot `ConfigWarningEvent`, mirroring the existing `autoswitch.model` typo guard. This matters more than it looks: `_resolve_account_identifier` returns a bare digit unexamined, so `--fallback-account 9` with three accounts resolves to `"9"` and would otherwise look like a safety net while being inert.

## Interface

- `cswap auto --fallback-account <num|email|alias>`
- `cswap config set autoswitch.fallbackAccount enterprise`

Off by default — with no fallback configured, behaviour is unchanged.

---

# 2. `autoswitch.strategy=priority` + `autoswitch.priorityAccounts`

The reported bug this fixes: a higher-priority enterprise account recovered to 0% used while a lower-priority `max` fallback account stayed active at 63% used. Both existing strategies chase headroom, and 63% is nowhere near the threshold, so nothing ever moved back — the expensive seat sat idle for the rest of the week.

`priority` follows a rank-ordered list you set instead of chasing headroom:

```
cswap auto --strategy priority --priority-accounts enterprise,max
cswap config set autoswitch.priorityAccounts enterprise,max
```

## What it does

Each tick, **before** the usual "active account is below the threshold, do nothing" early return, the engine asks whether an account ranked strictly above the active one now reads below `threshold`. If so it recalls to it — deliberately leaving an active account that is itself perfectly healthy. That is the opposite of the no-return bar's usual "don't flap back" rule, which is the point: a fixed rank order is an instruction, not a heuristic.

Four properties bound it:

- **Recall fires only while the active account is itself below `threshold`.** At or above it the ordinary headroom ranking decides, so `at-limit`/`failover` keep the full candidate list and `fallbackAccount` stays reachable. A rank order can never strand you on an exhausted account.
- **Every entry is capped by `threshold`, the last one included.** Recall departs a healthy account, so landing on a spent one would only bounce straight back.
- **An account not in the list ranks below every account that is**, so an unlisted active yields to any listed entry that qualifies.
- **Recall respects `cooldownSeconds`**, checked both at the decision site and again under the state lock in `_perform`, so a competing engine can't be superseded in the gap.

A quarantined higher-ranked entry is skipped rather than treated as a rank boundary, so a dead #1 can't hide a healthy #2. A recall that can't be landed returns `NO_ACTION`, not `BLOCKED` — the active is healthy by construction, and `best` in the same fleet state returns `NO_ACTION`, so cron wrappers keying on exit 3 mustn't be paged because a strategy flag is set. Like `fallbackAccount`, the pick runs through the ordinary freshen/quarantine/switch loop with `trigger="priority"` rather than a bespoke path.

Config problems get their own one-shot `ConfigWarningEvent`, one sentence per cause, because the remedies differ: an ambiguous email surfaces the resolver's own error, an identifier that can't be recalled to (unknown, disabled, no usable backup, API-key) is named and dropped, a duplicate of something ranked higher is named and dropped, and `--strategy priority` with an empty list says outright that it behaves exactly like `best`.

The rank order is acted on **only** when the strategy is `priority`; setting `priorityAccounts` alone can't perturb `best` or `consume-first` (the typo guard still validates the list under any strategy). Both strategies are byte-identical before and after this branch.

## What changed during review

Two of the four properties above came out of `pr-harden` rather than the original design, and both were real defects:

- The last entry was originally **uncapped**, on the stated theory that the freshen step judges it live. It doesn't — it judges the credential, never the quota. So recall left a healthy account for a spent one and `at-limit` bounced straight back, walking `[3, 1, 3, 1, …]` at cooldown cadence for the life of the window. Three lenses found it independently. The rule was removed entirely rather than special-cased to 100%, because a terminal entry at 95% with threshold 90 flaps identically.
- Recall originally **pre-empted `at-limit`/`failover`**, replacing the whole ranked candidate list with one rank-chosen account. When that account couldn't be freshened — `skip-live-session` never self-heals — an exhausted active was stranded for good, with both the ordinary ranking and the `fallbackAccount` escape unreachable. Measured 10/10 ticks `BLOCKED` on a 100% account with a healthy one idle at 10%.

- A **mutation sweep** over the new code (30 mutants) found the feature's headline promise untested: deleting the recall bypass — the line that stops rank-chosen targets being re-ranked by headroom — left all 2305 tests green, because every existing case happened to make the rank-preferred account the headroom winner too. Pinned now by `test_rank_decides_the_recall_target_not_headroom`. Six further survivors got a test each; three were proven equivalent mutants and left alone.
- Later rounds were almost entirely **prose**: the feature is described at five sites, and three consecutive rounds each found a fix that corrected one site and left an identical, now-false sentence standing at another. The structural answer was to delete the drifting `settings.py` copy and replace it with a pointer to the two canonical sites.

## Tests

`TestPriorityAccounts` covers the reported recall, the below-threshold cap on every entry including the last, a ten-tick no-flap walk, unreadable candidate usage, an unlisted active being outranked, quarantine skipping, the cooldown at both check sites, an exhausted active still getting the ordinary ranking, the unreachable-target exit code on both the drained-loop and the transient-freshen tails, inertness under `best`, and each of the four config-warning branches. Plus `merged_with_cli` coverage for `priorityAccounts` and a pin that `priority` is in the strategy spec's `choices` — `_clamped` reverts anything the spec doesn't list, so a missing entry would make the strategy unsettable through both surfaces at once.

Full suite green (2313 passed, 4 skipped).

<details>
<summary>Reviewed and deliberately skipped</summary>

- **Stale pre-escalation snapshot.** Priority recall fires outside the adaptive poller's escalation band, so it decides on a stored usage snapshot that can be up to `STALE_OK_S` old. Confirmed reachable, but bounded: `at-limit` bounces off a target that turns out to be spent on the very next tick, no wrong state persists, and no flap follows. The proposed fix — fetching the target's usage before committing — would have tripled baseline HTTP traffic or broken the flat-traffic invariant the poller exists to hold. Took the one-line docstring correction naming the bound instead.
- **No-return-bar poisoning.** A claim that recall arms the anti-flap bar in a way that later strands the fleet. Refuted on verification: the control arm was rigged (`best` never switched at tick 1, so it never armed the bar either), hand-seeding `best` with identical state gives an identical `BLOCKED`, and the proposed fix disarms an anti-flap gate with no demonstrated benefit.
- **Making `at-limit`/`failover` honor rank order.** Proposed as the alternative fix to the pre-emption defect above. Rejected: it re-opens the same stranding class the fix closed — an exhausted active with an unfreshenable rank-0 target has nowhere to go.
- **The 100-word cap on the README strategies bullet.** Compacted 185 → 136 words rather than cut to the cap. The sentences flagged for deletion encode exactly the properties an earlier round required the README to state, so deleting them trades one finding for another; the untouched sibling `fallbackAccount` bullet already sits at 145 words.

</details>


buck-openart:
Pushed a fix after hitting this running the branch: a 2-account fleet with account 1 at 100% used and account 2 as the configured `fallbackAccount`, whose refresh token was dying — so its usage read as unreadable, but it wasn't quarantined yet. The escape hatch never fired, because the gates ahead of it required every OAuth candidate *including the fallback destination itself* to have known `<=0` headroom, so the engine logged `no candidate has readable usage` on every tick while a usable fallback sat right there. `f92929f` resolves fallback eligibility up front and excuses only the fallback's own headroom from that exhaustion test — the freshen step immediately below tries the account live anyway, so insisting on a cached reading first was never buying anything, and every *other* candidate must still be confirmed exhausted before the jump. `b19791a` is a self-review follow-up that corrects a comment which stopped being true once the fallback no longer had to prove its own exhaustion, and pins the excused-fallback-cannot-land shape in tests.

One consequence worth naming rather than burying: those freshen-loop tails can now report the block on a headroom nobody read. A 2,240-case A/B sweep against the parent commit bounds it — 36 shapes nap where the parent polled, **zero** spin where the parent napped (so 770d2a6's unattended-laptop guarantee holds), and in all 36 the parent had no escape hatch at all. Happy to change that trade if you'd rather. Full suite green: 2284 passed, 4 skipped.

🤖 Generated with Claude Code

PR #352, opened 2026-09-14
Add `import-usage`: take usage readings from another machine instead of polling

BarganConstantin:
## Summary

When the same accounts are logged in on more than one machine, every machine asks `/api/oauth/usage` for the same numbers, and together they run into 429s. This adds `cswap import-usage`: one machine keeps polling, and the others take its readings instead of polling those accounts themselves.

```bash
# On the machine that polls, once per machine that should stop polling:
cswap list --json | ssh laptop cswap import-usage - --hold 600
```

## Why

On the current regime the usage endpoint's budget is per account (#152), so every machine that polls an account spends the same budget. Nothing lets one machine tell the others it already has the numbers, so today they only settle down the way #152 describes: after hitting 429 and backing off.

On one of my machines, `claude-swap.log` recorded 170 usage 429s between Aug 6 and Sep 14 (159 with `Retry-After: 0`, 10 with 3600s), up to 60 in a week, with the same accounts logged in on other machines too.

## How it works

- **The input is `cswap list --json` from another machine.** Each row whose `usageStatus` is `"ok"` is matched to a local account by email and organization, the key `import` already uses. Slot numbers are never used, because they differ between machines.
- **Only newer readings are taken.** A reading replaces the stored one only if it was measured later. Its age comes from `usageAgeSeconds`, so the stored fetch time is "now minus that age" on this machine's clock, and the two clocks never have to agree. Failure, backoff, poll plan and claim state are not touched: they describe this machine's own requests.
- **`--hold SECONDS` stops this machine from polling those accounts for that long.** It writes a new `heldUntil` field on the row. `_row_eligible` and `due_candidate` skip a held row, so none of the collectors (`list`, `status`, `switch`, `auto`, the dashboard, `watch`, the menu bar) fetches it. While the hold lasts, the reading stays trusted for switch decisions, the same way a scheduled poll interval keeps a reading trusted today.
- **A hold always ends.** It can't run past `TRUST_MAX_AGE_S` (one hour) after the stored reading was measured, so an account never sits unpolled on numbers too old to act on. Whoever sends the readings renews the hold each time; if they stop, this machine goes back to polling when the hold runs out.
- **One change in the collector.** A held active account whose token has expired now shows its reading instead of `token_expired`. That state covers the wait for the next fetch, which is what refreshes the token. A held account isn't waiting for a fetch, because it was just handed a current reading. The fetch refreshes the token once the hold ends.

## What stays the same

- claude-swap still knows nothing about other machines. Whatever coordinates them, a cron job over ssh for example, decides which machine polls and pipes the output across.
- No new settings, no new dependencies, no platform-specific code.
- `list --json` output is unchanged; `import-usage` reads it as it is.
- A machine that never runs `import-usage` behaves exactly as before.

Left out on purpose: rows without usage (`unavailable`, `token_expired` and so on) are skipped rather than adopted, and 429 state isn't transferred, since a machine that is backing off can simply renew the hold. Happy to add either in a follow-up if you'd want them.

## Files

- `src/claude_swap/usage_store.py` +74/-3: the `heldUntil` field, checked by `_row_eligible`, `due_candidate` and the trust check, and the new `UsageStore.adopt`.
- `src/claude_swap/json_output.py` +75: `usage_from_json`, which reads `list --json` usage back into the stored format.
- `src/claude_swap/transfer.py` +105: `import_usage`, which checks the whole document before writing anything, as `import` does.
- `src/claude_swap/switcher.py` +8: the `token_expired` change above.
- `src/claude_swap/cli.py` +29: the `import-usage` subcommand and `--hold`.
- `README.md` +10: a short "Share usage readings between machines" section.
- `tests/` +343.

## Testing

- Full suite, no failures: macOS 2297 passed, 4 skipped (main: 2262); Linux, Python 3.12, 2271 passed, 30 skipped; Windows 2220 passed, 81 skipped.
- 35 new tests. One goes through the real collector: after an import with a hold, `list --json` serves the imported reading and sends no request for that account, while an account nobody sent a reading for is fetched as usual.
- Each new check was mutation-tested: removing the hold check in `_row_eligible` or `due_candidate`, the trust extension, the one-hour cap, the newer-only rule or the `token_expired` change makes at least one test fail.


cristikl5:
Nice idea, I’ve had the same problem with multiple machines. Would definitely be useful for my setup too.

dorintriboi:
This is a really nice solution. I’ve been running into the exact same problem with the same accounts being used across multiple machines and getting unnecessary 429s from the usage endpoint.

Having one machine poll and letting the others reuse those readings makes a lot more sense than having every machine hit the same endpoint independently. The fallback when the hold expires is especially nice, no extra coordination service needed.

Would definitely use this.

PR #355, opened 2026-09-16
Per-window switch limits (autoswitch.windowThresholds) and a window selector (autoswitch.windows)

kmadan:
`autoswitch.threshold` is compared against the binding window, the higher of an account's 5h and 7d utilization. One number therefore governs both windows, and the two do not behave alike: the 5-hour window empties in minutes under heavy use and then recycles, while the weekly window is the budget and is worth spending close to its ceiling. A 5-hour limit of 90 forces 90 on the weekly window, so a weekly figure at 91% moves you off an account whose 5-hour window is nearly empty, and rules that account out as a landing target.

Two settings, both defaulting to today's behaviour.

**`autoswitch.windowThresholds`** (default empty) gives a window its own limit: `5h:85,7d:97`. Anything unnamed stays on `autoswitch.threshold`. Per-model names work too (`Fable:95`).

**`autoswitch.windows`** (default `5h,7d`) selects which account-wide windows the decision reads at all. `5h` drops the weekly window below exhaustion. This is the narrower of the two, and the second commit largely supersedes it: judging the weekly window at 97 rather than 90 removes the reason I originally wanted it dropped. It is a separate commit and easy to leave out if you would rather not carry it.

### Design

Roughly a dozen call sites compare the binding window against `autoswitch.threshold`: the engine tick, candidate ranking, the poll scheduler, `_every_account_above_threshold`, manual `switch --best`. Teaching each of them a second number invites them to disagree, and the module deliberately relies on them agreeing ("the same set `account_headroom` measures, so ranking and headroom can never disagree about which window matters").

So `effective_pct` restates a window carrying its own limit onto the threshold scale, inside `relevant_windows`, where every one of those sites already reads:

- at or over its limit, report at least `threshold`, so it blocks;
- below its limit, report just under `threshold`, so it cannot;
- at `WINDOW_EXHAUSTED_PCT`, report unchanged. 100% means no request can be served whatever was configured, so the no-viable-target handling stays correct.

The restatement is exact outside the interval between the two numbers. With `7d:97` against a threshold of 90, a weekly window at 70 reports 70 and at 98 reports 98; only `[90, 97)` compresses, which is the interval where the raw figure and the verdict disagree anyway.

One consequence worth calling out for review: the every-account-above-threshold escape ranks by `_binding_recovery_ts`, which reads the same restated windows. With every account in the 90s on 5h and the 93s on 7d, a single threshold makes the weekly window bind, and its reset is days out, past `RECOVERY_HORIZON_S`. Ranking then falls back to headroom, where nothing clears `HORIZON_HEADROOM_RATIO` because every account is equal, and the engine parks. Judging the weekly window at 97 keeps the 5-hour window binding, so the escape ranks by a reset minutes away and keeps working. That case is what prompted this.

Both resolvers memoize on the `settings.json` mtime, matching `decision_windows`, so `cswap config set` reaches a running `cswap auto` without a restart. Both fall back to documented defaults on any failure to read settings, so a corrupt file cannot silently narrow a decision.

### Tests

50 new tests, suite at 2313 passed / 3 skipped locally on 3.14.

- Both parsers, including each way a hand edit can be wrong. Every bad input falls back rather than raising, so a typo costs one window its custom limit and nothing more.
- `effective_pct` at the points that decide a switch, plus exactness outside the band asserted per value and monotonicity across the limit.
- `relevant_windows` honouring both settings, including that `resets_at` survives the restatement, since the scheduler reads its reset from the same tuple.
- Settings resolution: configured values, a later edit picked up through the mtime key, and both unreadable-settings fallbacks.
- `TestConfigList` updated from 9 keys to 11, which is what caught the new settings needing to be declared.

### Compatibility

Defaults reproduce current behaviour exactly: `windows` defaults to `5h,7d`, and an empty threshold map leaves every window on `autoswitch.threshold`. The existing `TestRelevantWindows` cases pass unchanged.


yunjeongiya:
Field report from a two-account fleet that hit exactly the case in the description today, plus a run on this branch.

**Setup**: `cswap auto --strategy consume-first --threshold 85` as a resident loop, v0.26.0. Account 2: 5h 95% / 7d 47% (7d resets in ~2d). Account 3: 5h 7% / 7d 93% (7d resets in ~6h).

**Symptom**: a human `cswap switch 3` was reverted by the loop **within the same tick, three times in a row** — account 3's binding window (7d 93%) is over the single threshold, so it is evacuated, and the only landing target is account 2 at 95% on its 5h window:

```
{"event":"poll","ts":"2026-09-18T08:55:57Z","active":{"number":3},"headroomPct":{"2":7.0,"3":7.0},"threshold":85.0,"windowsPct":{"2":{"5h":93.0,"7d":47.0},"3":{"5h":5.0,"7d":93.0}}}
{"event":"switch","ts":"2026-09-18T08:55:57Z","trigger":"proactive","from":{"number":3},"to":{"number":2}}
{"event":"poll","ts":"2026-09-18T09:01:56Z","active":{"number":3}, ...}
{"event":"switch","ts":"2026-09-18T09:01:57Z","trigger":"proactive","from":{"number":3},"to":{"number":2}}
{"event":"poll","ts":"2026-09-18T09:10:47Z","active":{"number":3}, ...}
{"event":"switch","ts":"2026-09-18T09:10:47Z","trigger":"proactive","from":{"number":3},"to":{"number":2}}
{"event":"switch","ts":"2026-09-18T09:22:48Z","trigger":"at-limit","from":{"number":2},"to":{"number":3}}
```

The reverse leg never qualifies (account 3's 7d reset is outside the recovery horizon, and 7 vs 5 points of headroom fails the ratio), so the fleet converges on the account that is about to hit its 5h limit and only leaves it at 100%. A weekly window at 93% that resets in six hours is far more quota than a 5h window at 95% that resets in forty minutes — one number for both windows cannot express that.

**On this branch** (`f96153d` + `6cb78cf`, installed from the PR head on Windows 11 / Python 3.12): `cswap config set autoswitch.windowThresholds 5h:85,7d:97`, restarted the loop — the next ticks stay on account 3 (7d 94% < 97) with no proactive evacuation. `uv run pytest -q` on the branch: 448 passed, 6 skipped.

One gap this does not cover (and probably should not): a human's explicit `switch` still has no way to be respected while the engine disagrees. I'll open that as a separate opt-in (`autoswitch.manualHoldSeconds`) so it does not widen this PR.

yunjeongiya:
Follow-up after running the branch for an hour: one wiring gap, easy to miss because it only shows when `--threshold` is passed on the command line.

`oauth.window_thresholds()` returns `loaded.threshold` from **settings.json** (90 when unset), and `effective_pct` anchors the restatement to that number — while the engine compares the binding window against the **CLI-merged** threshold (`merged_with_cli`, here `--threshold 85`). With `windowThresholds=7d:97` and settings-file threshold 90, a 7d value of 89.99 is restated to `min(89.99, 90 - 0.01) = 89.99`, which the engine then reads as ≥ 85 → `proactive`. Observed:

```
{"event":"poll","ts":"2026-09-18T09:51:00Z","active":{"number":3},"headroomPct":{"2":52.0,"3":10.01},"threshold":85.0,"windowsPct":{"2":{"5h":0.0,"7d":48.0},"3":{"5h":25.0,"7d":89.99}}}
{"event":"switch","ts":"2026-09-18T09:51:00Z","trigger":"proactive","from":{"number":3},"to":{"number":2}}
```

Setting `autoswitch.threshold 85` in the file (matching the CLI) makes the restatement land at 84.99 and the eviction stops, so the logic is right — it is only the anchor. Two ways out: thread the engine's effective threshold into `relevant_windows`/`effective_pct` (probably via the same settings object the engine holds), or document that `windowThresholds` assumes the file's `threshold` is the one in force and `--threshold` on the CLI is not.

kmadan:
Thank you for both of these, and for running the branch.

**The wiring gap is real and is fixed in `329b0cf`.** Reproduced exactly as you describe: with `windowThresholds=7d:97` and a settings-file threshold of 90, a 7d window at 95 restates to 89.99, and an engine running `--threshold 85` reads that as over the line and evacuates. A TUI session threshold reaches the same state, since `apply_threshold` also diverges from the file.

I took the first of your two options. `relevant_windows`, `account_headroom` and `poll_policy.binding_pct` now take the effective threshold, defaulting to the file's when the caller has no opinion, and every in-engine call passes `settings.threshold`, which is the CLI-merged value and carries a session override too. Documenting the limitation instead would have left the restatement anchored to a number the engine does not compare against, which is exactly the kind of quiet disagreement the single-funnel design is meant to prevent.

The manual `switch --best` and `next-available` paths keep reading the file deliberately: no flag overrides the threshold there, so the file is the number in force.

Two regression layers, both new:

- the restatement never blocks a window under its own limit and always blocks one at it, parametrised across effective thresholds from 50 to 99;
- end to end, an engine at 85 against a file saying 90 holds an account whose weekly window reads 95 under a 97 limit, which is the case in your log.

Suite is 2406 passing locally.

**On the field report**: that is the shape this PR is for, and your log makes the mechanism concrete in a way the description did not. Account 3 is evacuated on its 7d at 93 against a single 85 threshold while its 5h sits at 5, and the only landing target is account 2 at 93 on the 5h, so the fleet oscillates and neither leg is wrong on its own axis. With `windowThresholds=5h:85,7d:97` account 3 stops being evicted at 93 and the reverse leg never arises. If you try that on the branch I would be glad to know whether it holds on your fleet, since two accounts is the case where there is no slack to hide a bad decision.

kmadan:
Correction to my previous comment: the SHA I quoted, `329b0cf`, is on a stacked branch in my fork, not on this PR. The fix is now on this branch as `42a0188`, so what you tested and what the PR proposes are the same thing again. Apologies for the noise.

The content is as described: `relevant_windows`, `account_headroom` and `poll_policy.binding_pct` take the effective threshold, defaulting to the file's when the caller has no opinion, and every in-engine call passes `settings.threshold`. Suite 2326 passing on this branch.

yunjeongiya:
Thanks for reproducing and fixing the threshold anchor in `42a0188`.

Follow-up from the same two-account Windows fleet: after our September 18 restart at 12:05:29 UTC, through the September 21 23:55:39 UTC poll, the resident log contains 10 automatic switches (5 proactive, 2 at-limit, 3 consume-first) and 66 `manual-hold` decisions. Five switches occurred in the first night; the remaining five were spread over the following days. This is substantially quieter than the repeated immediate reversals in the original report.

A configuration caveat matters here: this was our combined local build with #375's 30-minute manual hold as well as this PR, initially configured with `5h:85,7d:97`. On checking the machine today, the current window limits are `5h:85,7d:99`, the resident CLI passes `--threshold 95`, and the settings-file threshold is still 85. So these multi-day observations are not a controlled validation of an unchanged `5h:85,7d:97` setup, nor can all of the improvement be attributed to this PR alone.

I have now cherry-picked `42a0188` onto that combined build, run the autoswitch/OAuth/poll-policy/switcher/CLI tests (1,068 passed, 19 skipped), reinstalled it, and restarted the resident loop. The installed functions now accept the effective threshold, and the first live poll completed normally. This removes our reliance on keeping the file and CLI thresholds aligned. The multi-day numbers above are from before this latest fix; longer-running field validation of `42a0188` starts now.

The per-window policy has been useful on this two-account fleet. Thank you for following through on the CLI/session override case.

PR #364, opened 2026-09-16
fix(update-check): notify users running a pre-release build

BarganConstantin:
## Problem

`_parse_version` in `src/claude_swap/update_check.py` is `tuple(int(x) for x in v.split("."))`, so it only handles purely numeric dotted versions. This project ships every cycle as a pre-release first — `pyproject.toml` on `main` is `version = "0.27.0b1"`, and 0.26.0b1, 0.25.0b2, 0.24.0b1 and friends came before it — and `claude_swap.__version__` is read straight from package metadata. `"0.27.0b1".split(".")` is `["0", "27", "0b1"]`, and `int("0b1")` raises `ValueError`.

That `ValueError` is raised at the one comparison in `check_for_update`, inside its blanket `except Exception: return None`, so the function reports "no update available" rather than failing. `cli.main` calls `check_for_update(__version__)` on every invocation that is not `--json`, `--purge` or `--upgrade`, which means anybody on a pre-release build silently stops receiving the passive update notice — not for the next beta, and not when the stable release that supersedes their build lands. Nothing is printed and nothing is logged, so there is no way to notice from the outside; the maintainers, running the repo's own version, cannot see the feature work at all.

## Fix

`_parse_version` now reads a release segment plus an optional PEP 440 pre-release suffix (`a`/`alpha`, `b`/`beta`, `c`/`rc`/`pre`/`preview`, with the usual `-`/`_`/`.` separators and an optional number) and returns a `NamedTuple` sort key of `(release, pre_rank, pre_number)`. Comparing two of those orders 0.27.0a1 < 0.27.0b1 < 0.27.0rc1 < 0.27.0 < 0.28.0, which is what the check needs. Trailing zeros are dropped from the release segment so that 0.27 and 0.27.0 compare equal, as PEP 440 requires — without that, a final "0.27" would sort below the "0.27.0b1" it supersedes. It is all `re` and `typing`: `packaging` is not a dependency of this project and one notification string is not worth making it one.

A few things are deliberately left alone. Development and post releases are not modeled, so `0.27.0.dev1` and `0.27.0.post1` collapse onto `0.27.0` instead of sorting before and after it; the project publishes neither, and this is still strictly better than today's silence. A version string the pattern cannot read still raises `ValueError`, which `check_for_update` already treats as "say nothing", so a garbage `info.version` from PyPI keeps failing closed exactly as it does today.

The one judgement call worth flagging: the comparison moved into a small `_is_newer` helper that refuses to advertise a pre-release to someone running a final release. PyPI puts the latest *stable* in `info.version` whenever one exists, so that side of the comparison is very unlikely to be a pre-release in practice — but it would have been a free ride from "silently nothing" to "every stable user is prompted to upgrade onto a beta", and that is a policy change this fix has no business making. Users already on a pre-release still hear about later pre-releases, which is the case that actually matters here.

## Testing

New `TestCheckForUpdatePrereleases` in `tests/test_update_check.py`, reusing the existing `_make_pypi_response` helper and `CACHE_PATH` monkeypatch so nothing touches the real cache. Three of them were watched failing on unmodified `main` — `test_prerelease_is_told_about_later_release`, `test_prerelease_is_told_about_its_own_final_release` and `test_prerelease_is_told_about_later_prerelease` all asserted `result is not None` and got `None`, for exactly the swallowed-`ValueError` reason above (`3 failed, 28 passed`). The other two, `test_final_release_is_not_pushed_onto_a_prerelease` and `test_unparseable_version_stays_silent`, pass before and after on purpose: they pin the behaviour the fix must *not* change.

Full suite after the change: `2268 passed, 3 skipped` (2263 passed, 3 skipped on `main`, plus the five new tests).


realiti4:
Looking thanks for the fix and PR!

PR #375, opened 2026-09-18
auto: respect a human's switch for autoswitch.manualHoldSeconds (opt-in)

yunjeongiya:
## Why

A human's `cswap switch` can be undone by a resident `cswap auto` loop within the same tick. If the account the person chose already sits above `autoswitch.threshold` on its binding window, the next poll classifies it as `proactive` and moves off it — there is no way for a deliberate pick to be respected while the engine disagrees.

Seen today on a two-account fleet (`--strategy consume-first --threshold 85`): account B at 5h 7% / 7d 93% (weekly resets in six hours) was picked three times by hand and reverted three times, each within the same tick, to account A at 5h 95%. Field report with the log excerpts is on #355; that PR fixes the *threshold* half (one number governing both windows). This PR is the *human* half: even with sensible per-window limits, a person sometimes knows something the usage numbers don't.

## What

Opt-in setting `autoswitch.manualHoldSeconds` (int, default `0` = off, so nothing changes unless configured).

- The CLI `switch` forms (`switch <num|email>`, bare rotate, `--strategy best`) record the live account as a manual pick in the engine's existing `autoswitch_state.json`, under the same lock as the cooldown bookkeeping — no second store. A switch that never moved (cancelled prompt, unmanaged-account notice) records nothing.
- In the tick, `proactive` and `consume-first` are suppressed while the pick is active and younger than the hold: `NoSwitchEvent(reason="manual-hold", detail="<n>s left on manual pick of account N")`. `at-limit` and `failover` are untouched — a dead account is never held.
- The hold ends when the engine itself moves off the picked account (`_perform` pops the marker), so a later engine landing back on that slot is not a pick. A re-pick of the already-active account refreshes the hold on purpose.

## Tests

`tests/test_autoswitch.py::TestManualHold` (proactive suppressed, consume-first suppressed, at-limit still switches, expiry, ignored when active != pick, default 0 unchanged, engine switch never records, no live login is a no-op) and `tests/test_cli.py` (each `switch` form records; a cancelled switch does not). Config-surface test updated for the new key.

`uv run pytest -q`: 2196 passed, 81 skipped (Windows 11, Python 3.12).

Only the CLI records a pick; the TUI/menubar switch paths would be a one-line `switcher.note_manual_switch()` each if you want them included — left out to keep this small.


yunjeongiya:
Pushed a follow-up (`0428699`) after finding the hold silently never armed in real use.

`switch()` / `switch_to()` returned `None` whenever `json_output` was off (`... if json_output else None`), so the CLI gate I added in the previous commit ("record a pick only when the switch actually ran") never saw a payload for a plain `cswap switch <email>` — which is the form anything scripted around the human path uses. Result: `note_manual_switch()` was never called and the engine undid every manual pick on the next tick. The existing regression test missed it because it mocks the switcher (a `MagicMock` is not `None`).

`json_output` now only governs printing; the resolved result (switched / already-active / activated) is returned in both modes, and `None` is reserved for "nothing happened" (cancelled prompt, auto-add notice). The CLI already prints only under `if args.json and payload`, so human-mode output is unchanged. Added `test_human_mode_switch_reports_the_landing` and updated the two tests that asserted the old `None`.

Verified on a live install: `cswap switch <email>` now writes the `manualSwitch` marker to `autoswitch_state.json`, and the resident `cswap auto` loop logs `{"event": "no-switch", "reason": "manual-hold", "detail": "1792s left on manual pick of account 2"}` on the next poll instead of reverting.

